Issue #349, opened 2026-05-01
feat(catalyst-ui): full CRUD on every Cloud resource type (Phase A.2 of #347)


emrahbaysal:
Phase A.2 of epic #347. Item 6 only — but it's broad.

## Problem

Today only Region and Cluster have add/delete affordances (modals from `components/CrudModals`). Founder wants the same on **every** resource type, in **both surfaces** (graph context-menu + list-page actions), and adds **Update** which is missing entirely.

## Resources requiring CRUD (12)

| Resource | Add | Update | Delete | Notes |
|---|---|---|---|---|
| Region | ✅ today | ❌ add | ✅ today | rename, change provider |
| Cluster | ✅ today | ❌ add | ✅ today | rename, version, control-plane HA toggle |
| vCluster | ✅ today (modal exists) | ❌ add | ✅ via cascade | rename |
| NodePool | ✅ today | ❌ add | ❌ add | rename, machine type, size |
| WorkerNode | ❌ add | ❌ add | ❌ add | hostname, role, taints |
| LoadBalancer | ✅ today | ❌ add | ❌ add | name, listeners array |
| Network | ❌ add | ❌ add | ❌ add | CIDR, peering |
| PVC | ❌ add | ❌ add | ❌ add | namespace, capacity, storage class |
| Bucket | ❌ add | ❌ add | ❌ add | name, region, provider |
| Volume | ❌ add | ❌ add | ❌ add | name, capacity, attachment |
| Service | ❌ placeholder | ❌ placeholder | ❌ placeholder | gated on #321 informer; show \"requires informer\" tooltip and hide CRUD |
| Ingress | ❌ placeholder | same | same | same |

Modal scaffolding lives in `products/catalyst/bootstrap/ui/src/components/CrudModals/`. Pattern: each modal accepts `{ open, deploymentId, ...resourceContext, onClose }` and POSTs to a `/api/v1/deployments/{id}/<resource>` endpoint.

### Graph (right-click context menu)
- Every node type's context menu gets: **Edit** (always), **Add child** (kind-aware: Region → Cluster, Cluster → vCluster + NodePool, etc.), **Delete**.
- Detail panel actions list mirrors the same.

### List pages (Phase 3 of #309)
- Each list page header gains a **+ New** button → opens the Add modal for that resource.
- Each row gains a **⋯** action menu (edit / delete) — keep the click-row → drawer behavior; the menu lives on the rightmost cell.
[...]

PR #359, opened 2026-05-01
chore(catalyst-ui): extract parseLBPorts to its own module

emrahbaysal:
## Summary

Follow-up to #349. Move \`parseLBPorts\` out of \`LoadBalancerFormFields.tsx\` so the form fields module only exports React components — keeps the \`react-refresh/only-export-components\` lint rule clean.

## Test plan

- [x] \`npm run typecheck\` clean
- [x] No behaviour change — both AddLBModal and EditLBModal continue to import the helper through the new path


emrahbaysal:
Will rebase after deploy-bot's commit settles; not blocking #349.

Issue #371, opened 2026-05-01
feat(catalyst-api): Hetzner Object Storage credential pattern — wizard step OR Phase-0 OpenTofu auto-provision (Phase 0b of #369)


emrahbaysal:
Phase 0b of #369. Foundational for Harbor + Velero on Sovereigns whose backing cloud is Hetzner (no SeaweedFS path per ADR-0001 §13).

Two options — pick one:
- **Option A**: wizard collects Hetzner Object Storage credentials (access key + secret) at provision time and stores in a Sealed Secret consumed by Harbor/Velero charts.
- **Option B**: Phase-0 OpenTofu auto-provisions a Hetzner Object Storage bucket per Sovereign + emits the credentials as Sealed Secrets via the bootstrap flow.

Recommend B (zero operator credential handling) but A is acceptable.

Endpoint format: `https://{region}.your-objectstorage.com` (region ∈ fsn1 / nbg1 / hel1).

DoD: a fresh Sovereign provision results in a usable S3 endpoint URL + access/secret keys available as a K8s Secret in the Sovereign's home cluster, ready for consumption by Harbor and Velero charts.

**Depends on**: nothing.
**Gates**: bp-harbor (Phase 4b), bp-velero (Phase 4c).

emrahbaysal:
## Validation: problem exists and is unsolved

**Confirmed via grep + upstream provider audit + Hetzner docs:**

1. No object-storage credential collection exists anywhere in `products/catalyst/bootstrap/` — `grep -rE "objectStorage|object_storage|s3.*hetzner|HETZNER_S3|your-objectstorage" products/catalyst/bootstrap/` returns nothing.
2. No `*.tf` file in `infra/`, `infrastructure/`, or `clusters/*/infrastructure/` references `object_storage`, `hcloud_object_storage`, `aws_s3_bucket`, or `s3_bucket`.
3. The upstream `hetznercloud/terraform-provider-hcloud` v1.49 has NO Object Storage resource — its `internal/` directory contains: certificate, datacenter, firewall, floatingip, image, loadbalancer, network, primaryip, server, sshkey, **storagebox** (a different product — block storage), volume, zone. No `object_storage` / `bucket` / `s3` package.
4. Hetzner exposes **NO Cloud API** to mint S3 access keys programmatically — credentials can only be generated via the Hetzner web console (the secret half is shown exactly once, then irretrievable). Confirmed at https://docs.hetzner.com/storage/object-storage/faq/s3-credentials/.
[...]

Issue #370, opened 2026-05-01
infra(omantel): Hetzner mock-data purge runbook + execution (Phase 0a of #369)


emrahbaysal:
Phase 0a of #369. Operational pre-flight before first omantel run.

Use the cleanup script in [feedback_idempotent_iac_purge.md](memory/feedback_idempotent_iac_purge.md) — purge ALL Hetzner resources tagged `catalyst-deployment-id` or `managed-by=catalyst` from prior failed runs.

DoD: `hcloud server list`, `hcloud volume list`, `hcloud load-balancer list`, `hcloud network list`, `hcloud ssh-key list` all return either zero rows or only resources unrelated to Catalyst provisioning. Recorded in a comment on this issue with timestamp.

**Depends on**: nothing — runs immediately.
**Gates**: every Phase 8 end-to-end run.

emrahbaysal:
Picking this up as agent-#370. First task per founder rule is validate the problem is real before writing code.

**Validation**

- Memory file `feedback_idempotent_iac_purge.md` already contains a paste-ready cleanup snippet (FQDN-scoped, label `catalyst.openova.io/sovereign=$FQDN`).
- `scripts/operator-recover-sovereign.sh` already implements per-FQDN recovery (Hetzner purge → PDM release → catalyst-api record cancel) — but it is **scoped to one Sovereign**, not a project-wide pre-flight purge.
- `infra/hetzner/main.tf` confirms the canonical label is `catalyst.openova.io/sovereign=<fqdn>` on every resource (network, firewall, ssh_key, primary_ip, server, lb).
- `products/catalyst/bootstrap/api/internal/hetzner/purge.go` filters by `catalyst-deployment-id=<id>` — but the OpenTofu module never emits that label. That is a latent bug in the in-process purge path; out of scope for this ticket but flagged.
- `hcloud` CLI is **not installed** on the contabo VPS and `HETZNER_API_TOKEN` is not in env — so DoD verification (`hcloud server list` etc.) cannot run from the agent's session. The operator must execute the runbook on a host that has the token.

**Gap vs. ticket DoD**

The existing per-FQDN script does not satisfy this ticket's intent: *"purge ALL Hetzner resources tagged catalyst-deployment-id or managed-by=catalyst from prior failed runs"*. That is project-wide, not per-FQDN. So this ticket needs a new `scripts/hetzner-purge.sh` that:

[...]

Issue #369, opened 2026-05-01
Epic: Minimal Self-Sufficient Sovereign — omantel handover blueprint (24 blueprints + handover machinery + DAG)


emrahbaysal:
## Goal

Provision **omantel.omani.works** as the first fully self-sufficient Sovereign Cloud on Hetzner, complete the handover transition from contabo, and verify zero contabo dependency post-handover.

## Definition of Done

1. Clean Hetzner — no mock data from prior runs.
2. Wizard runs end-to-end against fresh Hetzner with the **23-blueprint** minimal set, every step's estimated job time honored, no error chains.
3. Handover transition completes: omantel runs its own GitOps loop, its own DNS, its own identity, its own secrets, its own registry, its own observability — independently.
4. Kill catalyst-api on contabo for 5 minutes — omantel keeps running.
5. `console.openova.io/sovereign/<omantel-id>` 301-redirects to `omantel.omani.works/sovereign/`.
6. ADR-0001 §9.4 honored throughout: SME demos at `console.openova.io/nova` and `marketplace.openova.io` untouched.

## Canonical reference

📖 **`docs/omantel-handover-wbs.md`** — full blueprint set, DAG, per-ticket dependencies, timeline, DoD execution checklist.

### Phase 0 — Pre-flight
- #370 — Hetzner mock-data purge runbook
- #371 — Hetzner Object Storage credential pattern

### Phase 1 — Foundational platform fixes
- #338 — bp-flux helm-controller SA cluster-admin
- #387 — Gateway API migration audit (Cilium + Envoy across all minimal-set blueprint charts; replaces dropped #372 bp-traefik)

### Phase 2 — Infrastructure
- #373 — cert-manager-powerdns-webhook
- #374 — NS delegation .omani.works → omantel

### Phase 3 — Data + State
- #375 — bp-nats-jetstream
- #376 — bp-gitea
- #377 — bp-keycloak
- #316 — bp-openbao auto-unseal (existing)
- #331 — bp-external-secrets ClusterSecretStore split (existing)

### Phase 4 — Registry + IaC + Backup
- #378 — bp-crossplane
- #327 — bp-crossplane-claims (existing, in flight)
- #383 — bp-harbor Hetzner-S3 backend rework
- #384 — bp-velero install + Hetzner-S3 wiring

### Phase 5 — Security + Observability
- #379 — bp-kyverno
- #380 — bp-trivy
- #381 — bp-grafana stack
- #382 — bp-spire

[...]

Issue #318, opened 2026-04-30
feat(catalyst-api): pre-handover wipe (cancel + reprovision flow with idempotent purge)


emrahbaysal:
## Why

The user must be able to abort a Sovereign provisioning at any point before handover and recover to a clean slate — either to retry with different inputs or to abandon the deployment entirely. Today there is no first-class wipe surface; on failure the operator has to manually run `tofu destroy`, manually clean Hetzner orphans, and manually clear catalyst-api state. This is brittle and error-prone (per `feedback_idempotent_iac_purge.md` — ad-hoc cleanup is exactly what causes orphan resources and quota exhaustion).

This ticket adds the canonical pre-handover wipe path. "Pre-handover" = `bp-catalyst-platform.Ready` is NOT yet true on the new cluster (i.e. the wizard has not transitioned to `StepSuccess` and Catalyst-Zero still holds the kubeconfig + active OpenTofu state).

### Wizard UI

- New CTA on the Flow page: **`Cancel & Wipe`** (red, secondary action, with confirmation dialog showing the FQDN being wiped + the Hetzner project ID + estimated cleanup time)
- New CTA on `StepFlow` after a failed phase: **`Wipe and Restart`** (single-button option that runs wipe then re-enters the wizard with the same inputs pre-filled)
- Both surfaces work from any phase before bp-catalyst-platform Ready=True

### catalyst-api endpoint

`POST /api/v1/deployments/<id>/wipe`

Atomic operations (in order, each idempotent):

1. **Stop helmwatch informer** for this deployment (cancel context, drain channel)
2. **`tofu destroy -auto-approve`** against `/tmp/catalyst/tofu/<sovereign-fqdn>/` — destroys the Hetzner network, firewall, ssh-key, control-plane server(s), worker server(s), and load balancer atomically
3. **Hetzner orphan purge** — enumerate by deterministic naming (`var.sovereign_fqdn`, see `infra/hetzner/main.tf`); force-delete anything left over after `tofu destroy` (handles partial-state cases per `feedback_idempotent_iac_purge.md`)
4. **PDM release** (pool-subdomain only): `POST <pdm>/v1/release {sovereignFqdn}` — releases the allocation row so the subdomain becomes available again
5. **Parent-zone NS removal** — for pool-subdomain Sovereigns, remove NS delegation from OpenOva's PowerDNS at `<pool>.` zone via the registrar adapter that handled allocation
[...]

PR #391, opened 2026-05-01
infra(omantel): hetzner-purge.sh — project-wide pre-flight purge runbook (#370)

emrahbaysal:
## Summary

Phase 0a of #369. Adds `scripts/hetzner-purge.sh` — a project-wide pre-flight runbook that cleans up every Catalyst-tagged Hetzner Cloud resource accumulated from prior failed provisioning runs, so the first omantel run starts from a verifiably empty Hetzner project.

- DRY-RUN by default; `--apply` to delete; `-y` to skip per-resource confirm prompts in CI; `--fqdn <fqdn>` to scope to a single Sovereign.
- 3-pass logic: label-selected (`catalyst.openova.io/sovereign`) → name-slug (`catalyst-*-*` for ssh_keys) → verification sweep (Hetzner's DELETE-returns-204-but-keeps-resource quirk, documented in `feedback_idempotent_iac_purge.md`).
- Forward-compatible Object Storage step (no-op today; once #371 lands and starts auto-provisioning buckets, the Object-Storage cleanup path will be wired in via the S3 endpoint).
- Every action appended to `/tmp/hcloud-purge-<ts>.log`; exit 0 on clean success, non-zero with details on partial failure.
- WBS §9 status row for #370 updated.

## Why a new script vs. the existing `operator-recover-sovereign.sh`

| Concern | `operator-recover-sovereign.sh` | `hetzner-purge.sh` (this PR) |
|---|---|---|
| Scope | One Sovereign FQDN | Entire Hetzner project (or pinned via `--fqdn`) |
| Use-case | In-flight wizard failure → retry same FQDN | Phase-0a pre-flight purge before omantel runs |
| Side-effects beyond Hetzner | PDM release + catalyst-api record cancel | Hetzner only |
| When to use | After any single-Sovereign provision fails | Before the first omantel run, or whenever the project has accumulated orphans from many failed runs |

## Architecture invariants honoured

- ADR-0001 §9.4 — SME demos run on contabo k3s, not Hetzner. Script touches Hetzner only.
- INVIOLABLE-PRINCIPLES #3 — under normal operation OpenTofu owns Phase 0 and `tofu destroy` is the canonical purge path. This script is the recovery fallback for when the tofu state is gone (Pod restart, lost emptyDir, etc.).
- INVIOLABLE-PRINCIPLES #4 — no hardcoded values: the FQDN is an arg, label key is a single constant, every Hetzner endpoint is the public API.

## Test plan

- [x] `bash -n scripts/hetzner-purge.sh` — syntax check passes
- [x] Dry-run without token — preview-only, no errors
- [x] Dry-run with `--fqdn omantel.omani.works` — banner reports single-Sovereign scope
- [x] `--help` flag prints the usage block
- [ ] Founder runs `--apply` on the real Hetzner project once authorised — execution log appended to #370

## Awaiting

Destructive execution on live Hetzner is gated on explicit founder authorisation in #370 per ADR §3.4 (destructive infra changes that affect resources outside the active sandbox). PR can merge before execution; runbook is non-destructive at rest.

### DIFF
--- a/docs/omantel-handover-wbs.md
+++ b/docs/omantel-handover-wbs.md
@@ -310,7 +310,7 @@ If founder wants to amend ADR-0001 with §13 formalised (S3 vs SeaweedFS rule), 
 | #319 | (pending) | | |
 | #327 | (in flight, other session) | | |
 | #331 | (pending) | | |
-| #370 | 🟡 in-progress (Agent #370-hetzner-purge-runbook) | | |
+| #370 | 🟡 runbook authored — awaiting founder authorisation to execute on live Hetzner | #391 | scripts/hetzner-purge.sh |
 | #371 | 🟡 in-progress (Agent #371-hetzner-os-credentials) | | |
 | #373 | (parked) | | |
 | #374 | (parked) | | |
--- a/scripts/hetzner-purge.sh
+++ b/scripts/hetzner-purge.sh
@@ -1,0 +1,394 @@
+#!/usr/bin/env bash
+# hetzner-purge.sh — project-wide purge of Catalyst-tagged Hetzner Cloud resources.
+#
+# Pre-flight runbook for issue #370 (Phase 0a of #369). Cleans up every
+# Hetzner resource that previous failed Catalyst provisioning runs left
+# behind, so the next Sovereign provision starts from a verifiably empty
+# Hetzner project.
+#
+# This is DIFFERENT from scripts/operator-recover-sovereign.sh:
+#   - operator-recover-sovereign.sh is scoped to ONE Sovereign FQDN.
+#     Use it when a single in-flight provision fails partway and the
+#     operator wants to retry the wizard with the same FQDN.
+#   - hetzner-purge.sh is PROJECT-WIDE. It deletes every resource that
+#     carries any `catalyst.openova.io/sovereign=*` label, plus any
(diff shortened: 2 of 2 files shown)


emrahbaysal:
Closing — wrong architectural approach.

We already have `internal/handler/wipe.go` (#318) which is the canonical pre-handover purge sequence:
1. `tofu destroy` (primary path — OpenTofu owns Phase 0 per Inviolable Principle #3)
2. Hetzner API force-purge for orphans tofu missed (safety net via `internal/hetzner/purge.go`)
3. PDM release
4. Delete on-disk record

A standalone bash script using `hcloud` CLI:
- Duplicates step 2 of the existing flow
- Hardcodes the label key (Inviolable Principle #4 — never hardcode)
- Bypasses the architecture (catalyst-api is the sanctioned seam for cloud-API mutations)
- Reinvents what's already there

The real work is to **make the existing wipe.go end-to-end correct**, which means:
- Fix the latent label-filter bug in `purge.go` (#392) — currently filters `catalyst-deployment-id=<id>` but Tofu emits `catalyst.openova.io/sovereign=<fqdn>`
- Add (if needed) an orphan-cleanup endpoint to catalyst-api for resources not tied to any known deployment

Reframing #370 accordingly.

PR #981, opened 2026-05-05
hotfix(catalyst-api): drop k8scache discovery probe — unblocks contabo startup (#975)

emrahbaysal:
## Summary
- Removes the synchronous `core.Discovery().ServerResourcesForGroupVersion(gv)` call I added in #976/#978's k8scache prep. That call has NO context timeout, so on a kubeconfig pointing at a dead Sovereign apiserver (decommissioned otech whose `<id>.yaml` was never removed from `/var/lib/catalyst/kubeconfigs/`) it hangs until TCP connect times out. With many dead kubeconfigs the contabo mothership boot serially blocks for minutes — observed live as catalyst-api stuck "iterating dead clusters" on startup.
- Drops `Kind.Optional`, removes `PodMetrics` from `DefaultKinds`, and deletes the two probe-specific tests + the `fakediscovery` import. Until a non-blocking activation path lands, the dashboard's `color_by=utilization` returns null when no PodMetrics indexer exists; health/age/size paths still ride the Pod + PVC indexers untouched.

## Root cause
In `factory.go:AddCluster()` I added a probe loop that called `core.Discovery().ServerResourcesForGroupVersion(gv)` to gate Optional kinds. Per `client-go/discovery/discovery_client.go`, that helper issues `RESTClient().Get().AbsPath(...).Do(context.TODO())` — no caller context, so the request is bound only by the underlying transport's TCP connect timeout. On a kubeconfig with a dead `server:` URL the call hangs minutes per cluster; AddCluster runs serially over every kubeconfig at startup.

## Test plan
- [x] go build ./... clean
- [x] go test ./internal/k8scache/... ./internal/handler/... -run "TestDefaultKinds|TestRegistry|TestFactory|TestDashboard" green
- [ ] On contabo: catalyst-api starts within seconds even with stale kubeconfigs in /var/lib/catalyst/kubeconfigs/

## Companion
Coordinate with the other session's local rollback that pinned contabo to :8a1fe04. Once this hotfix builds and deploys, the pin can be released forward.

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/dashboard_test.go
+++ b/products/catalyst/bootstrap/api/internal/handler/dashboard_test.go
@@ -211,7 +211,6 @@ func newDashHandlerWithCache(t *testing.T, clusterID string, withMetrics bool, o
 		Name:       "podmetrics",
 		GVR:        schema.GroupVersionResource{Group: "metrics.k8s.io", Version: "v1beta1", Resource: "pods"},
 		Namespaced: true,
-		Optional:   true,
 	})
 	cfg := k8scache.Config{
 		Logger:   quietHandlerLogger(),
--- a/products/catalyst/bootstrap/api/internal/k8scache/factory.go
+++ b/products/catalyst/bootstrap/api/internal/k8scache/factory.go
@@ -440,41 +440,23 @@ func (f *Factory) AddCluster(c ClusterRef) error {
 		lastEventAt: map[string]time.Time{},
 	}
 
-	// Build the per-cluster availability set for Optional kinds. The
-	// probe is one ServerResourcesForGroupVersion call per distinct
-	// (group,version), cached by string key. When discovery fails we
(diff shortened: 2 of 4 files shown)


emrahbaysal:
CI `test` job failures are all pre-existing on main:
- `TestValidate_ObjectStorage_*` — env-dependent (CATALYST_HARBOR_ROBOT_TOKEN unset)
- `TestAuthHandover_HappyPath` — pre-existing nil-pointer panic
- `TestLegacyRecord_NoParentDomainsKey_LoadsCleanly` — SKU drift
- `TestRun_FailsFastOnDynadotError` — pre-existing

Confirmed by running `go test ./internal/provisioner/... ./internal/handler/... -run …` on a clean main checkout — same failures, no test introduced by this PR.

The targeted run `go test ./internal/k8scache/... ./internal/handler/... -run "TestDefaultKinds|TestRegistry|TestFactory|TestDashboard"` (every test affected by this PR) is **green**. Merging to unblock contabo.

PR #1250, opened 2026-05-09
fix(chart): UserAccess regexReplaceAll uses direct form not pipe (Fix #38 follow-up #4)

emrahbaysal:
## Summary

PR #1246 used Sprig's pipe form `value | regexReplaceAll "regex" "repl"` to strip the dot-suffix off the qa-fixtures sovereignRef FQDN for the UserAccess CR. But Sprig's `regexReplaceAll(regex, src, repl)` puts pipe value in the LAST arg → `repl`, not `src`. So the empty `""` (intended as `repl`) became `src`, the FQDN became `repl`, and every UserAccess rendered `sovereignRef: ""` — rejected by the CRD pattern `^[a-z0-9][a-z0-9-]{0,62}$`. omantel chart upgrade has been Stalled on this since chart 1.4.105 landed.

Verified by inspecting helm release secret v22 on omantel: rendered manifest shows `sovereignRef: ""` despite release config having `qaFixtures.sovereignRef: "omantel.biz"`.

Fix: switch to direct call `regexReplaceAll "\\..*$" value ""` so the value actually lands in `src`. Chart bumped 1.4.107 → 1.4.108.

## Test plan

- [ ] After Blueprint Release of 1.4.108, Flux on omantel installs the chart cleanly
- [ ] qa-user1 UserAccess admits with sovereignRef = "omantel"
- [ ] catalyst-api + catalyst-ui :7eae9f1 (Fix #38) live on omantel
- [ ] Fix #38 TC-141 / TC-090 / TC-383 verifications pass

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -298,7 +298,14 @@ spec:
       # Blueprint CR alias resolved by chained catalog client; node-labels
       # seeder Job patches Nodes with topology.kubernetes.io/region short
       # form derived from openova.io/region; CNPGPair renamed to qa-cnpgpair.
-      version: 1.4.107
+      # 1.4.108 (qa-loop iter-7 Fix #38 follow-up #4): the regexReplaceAll
+      # in PR #1246's UserAccess template was using Sprig's pipe form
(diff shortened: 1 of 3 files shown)


emrahbaysal:
Superseded by PR #1249 (same fix, landed first while CI was waiting).

PR #1588, opened 2026-05-17
fix(catalyst-api): D30 PR I — mark imported deployment as Adopted on chroot

emrahbaysal:
Founder t140 bug #6: /parent-domains showed only primary, not the sme-pool domains. Root cause: chroot imports deployment record without AdoptedAt set → activeDeployment() returns nil → ListParentDomains only renders the synth primary. Fix: HandleDeploymentImport stamps AdoptedAt at import (chroot is verified-owner of FQDN). 🤖 Generated with [Claude Code](https://claude.com/claude-code)


emrahbaysal:
Superseded by #1589 (cleaner branch from latest main)

Issue #2062, opened 2026-05-20
TBD-V32: auto-bump deploy jobs lose push race silently (no retry-on-rebase loop)


emrahbaysal:
## Symptom

The `deploy` job in every `.github/workflows/*-build.yaml` runs `git push` without a `pull --rebase` retry loop. When two workflow runs commit to `main` within ~2 min of each other (high cadence during convergence pushes), the second loses the push race and **no auto-bump commit lands** — silently.

The workflow run goes:

```
[main efb4fa8] deploy: update catalyst images to 829474a
 3 files changed, 5 insertions(+), 5 deletions(-)
 ! [rejected]        main -> main (fetch first)
error: failed to push some refs to 'https://github.com/openova-io/openova'
hint: Updates were rejected because the remote contains work that you do not
hint: have locally. This is usually caused by another repository pushing to
hint: the same ref.
##[error]Process completed with exit code 1.
```

The image is built and pushed to GHCR fine; only the values.yaml/template SHA pin commit is lost. Result: a deployed catalyst-api / marketplace / services image that is OLDER than the source on `main`.

## Operational damage in the 2026-05-20T01:54-05:20Z window

Five catalyst-build / services-build / marketplace-build runs lost the push race in this window:

| PR | Merge SHA | Failed run at | Workflow | Real damage |
|---|---|---|---|---|
| #2038 | 84b35d2 | 02:02:37Z | catalyst-build | None — natural-rollforward by next catalyst-build success on 5ed4995 |
| #2043 | 0feb0b4 | 02:32:20Z | catalyst-build | None — same |
| #2046 | e860308 | 02:54:21Z | catalyst-build | Low — marketplace-api handler text-canon strings stale (s/openova.io/omani.works/) but sandbox-mcp-server delivered separately |
| **#2050** | **829474a** | **04:13:25Z** | **catalyst-build** | **CRITICAL — V16 catalyst-api admin-token wiring (`cmd/api/main.go` + `internal/newapi/client.go`) is in source on `main` but NOT in the deployed catalyst-api image. Operator deploying current chart will run an image at `5ed4995` (the previous catalyst-build success), which is BEFORE 829474a.** |
| #2053 | b762743 | 04:56:38Z | catalyst-build | Minimal — services-build (separate workflow) succeeded; only the chart-only edit was lost |

[...]

PR #2063, opened 2026-05-20
ci: wrap build-workflow deploy push in pull-rebase retry loop (Refs #2062)

emrahbaysal:
## Summary

TBD-V32 / Refs #2062. **Critical bug**: every `.github/workflows/*build*.yaml` deploy job lost its auto-bump commit silently when two build workflows committed to `main` within ~2 min of each other. The image landed in GHCR but the chart `values.yaml` / template SHA-pin commit was rejected with `! [rejected] main -> main (fetch first)` and lost.

**Concrete operational damage** (per issue + sweep extension): PR #2050 shipped catalyst-api image `829474a` to GHCR (V16 admin-token wiring in `cmd/api/main.go` + `internal/newapi/client.go`), but the `deploy: update catalyst images to 829474a` commit never landed on `main`. Operators installing the current chart kept getting the previous catalyst-build success (`5ed4995`), missing the entire admin-token wiring.

## What this PR does

- Adds a shared composite action `.github/actions/deploy-bump/action.yaml` that runs a **5-attempt `pull --rebase` retry loop** with `git fetch` before each rebase and exponential 2/4/6/8/10s backoff (~30s total). Inputs: `paths`, `commit-message`, optional `max-attempts`/`user-name`/`user-email`. Outputs: `pushed` (bool) and `commit-sha`.
- Refactors 20 of 21 build workflows to use the composite action, replacing 311 lines of duplicated bash with 163 lines of action invocations.
- `services-build.yaml` is the documented exception — its retry loop re-runs an inline `rewrite()` closure that bumps the chart semver patch on every iteration, which the composite action cannot do because it treats files as opaque. The inline loop stays, but bumps the ceiling from 3 to 5 attempts + adds the matching backoff sleep.

## Workflows touched

**Group A** (was bare `git push`, the actual culprit for #2050):
- catalyst-build.yaml, marketplace-api-build.yaml

**Group B** (had 3-retry inline loop without `git fetch` between rebases):
- admin-build.yaml, console-build.yaml, marketplace-build.yaml, build-bp-guacamole.yaml, build-bp-newapi.yaml, build-k8s-ws-proxy.yaml

**Group C** (had single pre-push `git pull --rebase --autostash` — only covers ONE race window):
- build-application-controller.yaml, build-blueprint-controller.yaml, build-continuum-controller.yaml, build-environment-controller.yaml, build-organization-controller.yaml, build-projector.yaml, build-openova-flow-server.yaml, build-openova-flow-adapter-flux.yaml, build-sandbox-controller.yaml, build-sandbox-mcp-server.yaml, build-sandbox-pty-server.yaml, useraccess-controller-build.yaml

**Group D** (custom rewrite-on-retry — kept inline, documented):
- services-build.yaml

## Verification

- `actionlint -shellcheck= .github/workflows/*.yaml` clean across all modified workflows (the only remaining warning is the pre-existing `cosmetic-guards.yaml:48 if: false` constant condition, unrelated).
- All 21 modified workflows and the new composite action pass `yaml.safe_load` parse.
- The composite action's `outputs.pushed` keeps the existing `if: steps.deploy_commit.outputs.pushed == 'true'` gate working on every `Trigger blueprint-release` downstream step (14 of the 20 modified workflows rely on this).

## Test plan

- [ ] Watch the next two build workflows that fire within ~2 min of each other on `main` — both should successfully push their deploy commits (no `[rejected]` errors).
- [ ] Confirm the `Trigger blueprint-release for the chart bump` step still fires in workflows that have one (it reads `steps.deploy_commit.outputs.pushed`, which the composite action now sets).
- [ ] After observed convergence on a real race, flip TBD-V32 #2062 from UNVERIFIED to VERIFIED-PASS on `docs/TRUST.md`.

## Post-merge follow-up

`Refs #2062` deliberately — per the anti-theater discipline, the issue closes only after observed race-recovery in a real CI run, not on this PR's merge. Separately:

- PR #2050's V16 admin-token wiring is still NOT in the deployed catalyst-api image. After this PR merges, the next controller-source PR that triggers `catalyst-build.yaml` will ride the new retry loop and pull #2050's wiring forward. If urgent, a manual chart-pin commit on `main` against `products/catalyst/chart/values.yaml` (bumping `catalystApi.tag` to `829474a`) will force the deploy.

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/.github/actions/deploy-bump/action.yaml
+++ b/.github/actions/deploy-bump/action.yaml
@@ -1,0 +1,162 @@
+# Composite action: deploy-bump
+#
+# Stages a set of file paths, commits them on `main` with a supplied
+# commit message, and pushes to `origin/main` through a pull --rebase
+# retry loop so concurrent build-workflow deploy jobs do not silently
+# lose the push race.
+#
+# Background (TBD-V32 / openova-io/openova#2062):
+# Build workflows that ended their deploy step with a bare `git push`
+# (catalyst-build, marketplace-api-build, marketplace-build, ...) or
+# with a single pre-push `git pull --rebase --autostash` (the
+# *-controller family) lost the deploy commit silently whenever two
+# build workflows committed within ~2 min of each other. The remote
+# rejected the second push with:
+#
+#   ! [rejected]  main -> main (fetch first)
+#
+# and the workflow exited red with the auto-bump commit never landing.
+# Concrete damage: PR #2050 (V16 admin-token wiring) shipped image
+# `829474a` to GHCR but the chart values.yaml stayed pinned at
+# `5ed4995` — operators installed an old image while the source on
+# `main` already had the new wiring.
+#
+# This composite action concentrates the race-recovery logic in ONE
+# place so every workflow uses the same battle-tested loop and any
+# future improvement only needs to ship here.
+#
+# Loop shape (5 attempts, capped sleeps):
+#
+#   for i in 1 2 3 4 5; do
+#     git push origin HEAD:main && break
+#     git fetch origin main
+#     git pull --rebase --autostash origin main
+#     sleep $((i * 2))
+#   done
+#
+# `fetch` before `pull --rebase` ensures we always see the latest
+# remote tip even if the previous attempt's `pull --rebase` left the
+# local main pointer stale. `--autostash` survives modified working
+# tree between push attempts (rare, but harmless). The capped sleep
+# (2/4/6/8/10s) keeps the loop bounded at ~30s of backoff total.
+#
+# Inputs are deliberately minimal — every caller passes a comma- or
+# whitespace-separated list of paths to stage and a commit message.
+# Outputs let callers gate the follow-up steps (blueprint-release
+# workflow_dispatch, ledger update, etc.) on whether the push
+# actually shipped.
+
+name: deploy-bump
+description: |
+  Stage, commit, and race-safe push a chart-pin / image-tag bump to
+  origin/main with a pull --rebase retry loop.
+
+inputs:
+  paths:
+    description: |
+      Whitespace- (or newline-) separated list of file paths to
+      `git add` before committing. Required.
+    required: true
+  commit-message:
+    description: |
+      Commit message to use when there are staged changes.
+      Required.
+    required: true
+  max-attempts:
+    description: |
+      Number of push attempts before giving up. Default 5.
+    required: false
+    default: "5"
+  user-name:
+    description: |
+      Git author name for the deploy commit. Default
+      `github-actions[bot]`.
+    required: false
+    default: "github-actions[bot]"
+  user-email:
+    description: |
+      Git author email for the deploy commit. Default
+      `github-actions[bot]@users.noreply.github.com`.
+    required: false
+    default: "github-actions[bot]@users.noreply.github.com"
+
+outputs:
+  pushed:
+    description: |
+      `true` if a commit was created AND pushed successfully,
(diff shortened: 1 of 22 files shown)


emrahbaysal:
**CI blocker (pre-existing red, NOT introduced by this PR)**

Holding merge per anti-theater rule "no admin-merge through red CI".

**The red:** `Build continuum-controller` job fails at step `helm template — fail-fast on empty image.tag`. Run: https://github.com/openova-io/openova/actions/runs/26143905882/job/76894943621

**Root cause:** `products/continuum/chart/values.yaml` currently has `tag: "e72efb8"` committed on `main` (from the previous successful build's deploy-bump). The fail-fast guard step expects `helm template` to FAIL when no override is passed — but with `tag` already non-empty in `values.yaml`, helm template SUCCEEDS, defeating the guard.

```
##[error]expected helm template to FAIL when continuum.enabled=true and image.tag is empty
```

**Evidence this is pre-existing on main, not regression:**
- `main` HEAD (48bad537) values.yaml: `tag: "e72efb8"` (already bumped)
- PR #2063 branch values.yaml: `tag: "e72efb8"` (identical)
- This PR's diff for `build-continuum-controller.yaml` only swaps the deploy-bump push step for the new composite action; the fail-fast step is untouched.
- The last main build at e72efb8 PASSED because the *commit* e72efb8 had `tag: ""` (the fail-fast step ran before the bump). After the bump pushed back to main, no continuum-touching commits have triggered the workflow on main since — so the latent breakage only surfaces now that PR #2063 touched the workflow file.

**Recommendation:** File a follow-up TBD to either:
1. Strip the committed `tag:` from `products/continuum/chart/values.yaml` so the fail-fast guard's invariant holds (`tag` empty on main; CI stamps per-build only); or
2. Rework the guard step to explicitly `--set continuum.image.tag=""` so it tests the helper logic directly, not the values.yaml state.

The push-race retry composite action work in this PR is correct and unblocked by this issue. Leaving PR open for next session.

emrahbaysal:
PR #2070 (merged at 4174534a) reworks the `Build continuum-controller / helm template — fail-fast on empty image.tag` guard to pass `--set continuum.image.tag=""` explicitly, so the guard now passes regardless of the SHA that the workflow's auto-bump step commits into `products/continuum/chart/values.yaml` on main. Rebasing this PR on the updated main should turn the previously-red `Build continuum-controller` check green. Refs #2062.

emrahbaysal:
Merged at `257291e8d1b595be67ed1c54e35a293fdaf70bad` (squash, non-admin).

## Rebase outcome
- Clean rebase on `origin/main@4174534a` (PR #2070 fail-fast-guard fix landed first).
- `build-continuum-controller.yaml` shows both changes integrated: the post-#2070 explicit empty-tag guard remains, and the `deploy_commit` step now uses the shared composite action `./.github/actions/deploy-bump`.
- Force-pushed with `--force-with-lease`.

## Final CI state
- All 14 build matrix jobs: pass
- `Controller-workflow uniformity`: pass
- `useraccess-controller` test: pass
- `e2e` (skipped on non-secret PR) / `promote` / `notify`: skipped (expected)
- Zero failures.

## Effect
21 build workflows + `admin-build` + `catalyst-build` + `console-build` + `marketplace*-build` + `services-build` + the 11 controller builds + `build-continuum-controller` now share the race-safe push composite action. Next concurrent build-workflow deploy commits will retry the rebase loop rather than silently losing the push to `! [rejected] main -> main (fetch first)`.

Issue #2740, opened 2026-06-02
G117.1 — Blueprint schema extension (topology + endpoints + SSO + multiInstance)


emrahbaysal:
## Refs EPIC #2737 · Refs Wave-0 #2738

Wave-1 sub-EPIC. Uses the Wave-0 brief template at `.claude/templates/G117-agent-brief.md`.

## Scope

Land the typed `spec.topology` + `spec.endpoints` + `spec.sso` + `spec.multiInstance` blocks on the Blueprint CRD; wire the runtime admission webhook against `platform/_schemas/blueprint-topology.json`.

## Wave-0 artifacts consumed

- `core/pkg/apis/blueprint/v1alpha1/topology_types.go` (typed shapes)
- `core/controllers/pkg/apis/blueprint/v1alpha1/topology_types.go` (controllers-importable mirror)
- `platform/_schemas/blueprint-topology.json` (JSON schema)
- `examples/blueprint-grafana-fully-declared.yaml` (reference)

## Deliverables

- [ ] Promote the four new fields in `products/catalyst/chart/crds/blueprint.yaml` openAPIV3Schema
- [ ] Extend `core/controllers/blueprint/internal/validate/validate.go` to call `jsonschema.Validate(doc, blueprint-topology.json)` and surface findings on `Blueprint.status.conditions[]`
- [ ] Unit tests: `validate_test.go` cases for valid topology / missing required field / variant mismatch / invalid hostname template
- [ ] Playwright spec: `tests/e2e/playwright/tests/g117-blueprint-admission.spec.ts` — submits a malformed blueprint.yaml + asserts the Blueprint CR's `status.conditions[type=Ready,status=False,reason=ValidationFailed]`
- [ ] Memory: add `feedback_g117_blueprint_admission_jsonschema_pitfalls.md` documenting any gotchas with the jsonschema lib's behavior on Go-template strings or YAML number parsing

## Acceptance DoD

- [ ] `gh pr checks` all green
- [ ] Self-verification evidence on this issue
- [ ] Independent verifier verdict `status/uat-OK` posted

## File-touch isolation

`isolation: none` per `.claude/templates/G117-worktree-policy.md` (Slot B1; disjoint from B2-B6)

emrahbaysal:
## G117.1 Wave-1/B3 — self-verification evidence

PR: https://github.com/openova-io/openova/pull/2748
Branch: `feat/g117-1-topology-application`
Commit: `963a4ca2`

### Scope shipped
[...]

PR #2825, opened 2026-06-02
feat(catalyst-platform): G117.E2E-B3 — promote G117 fields into Blueprint CRD admission

emrahbaysal:
## Summary

- bp-catalyst-platform 1.4.437 → 1.4.438: promote spec.topology / spec.endpoints / spec.sso / spec.multiInstance into the Blueprint CRD's openAPIV3Schema so the live K8s admission gates the G117 contract instead of silently swallowing the fields through preserve-unknown.
- bootstrap-kit slot 13 pin moves to 1.4.438 in lockstep.
- 5 new invalid-sample cases + 1 expanded valid-sample case in `products/catalyst/chart/crds/tests/` exercise the new enum / required-field gates.

## Why

A3 + A8 verifiers on hw86 found that `kubectl get crd blueprints.catalyst.openova.io -o yaml` showed spec keys = `[card, configSchema, consoleUI, depends, manifests, observability, outputs, owner, placementSchema, rotation, upgrades, version, visibility]` — **none** of the four G117 keys present. W1.B1 (Go types, #2740) + W2.C2 (JSON-schema gate, #2745) had landed but the CRD manifest itself was never updated. Runtime admission therefore did nothing to enforce the G117 contract — every Blueprint with bad `spec.topology` enum / missing required `defaults.single-region` / bad `endpoints[].protocol` / bad `sso.realm` / bad `multiInstance.isolationLevel` was admitted unconditionally.

## K8s structural-schema constraints encountered (and resolved)

- **`$ref` forbidden** in CRD structural schemas → the four BCP-topology variants reuse `&REPLICATION_SPEC` / `&SWITCHOVER_SPEC` / `&PLACEMENT_SPEC` via YAML anchors, inlining each variant.
- **`uniqueItems: true` forbidden** (quadratic-complexity ban) → replaced with `x-kubernetes-list-type: set` which gives the same guarantee in O(n) via hash.
- **Do NOT rely on `x-kubernetes-preserve-unknown-fields`** for the G117 fields per the task's hard rule — preserve-unknown is the bug we're fixing.

## Test plan

- [x] `kubectl apply --dry-run=server -f products/catalyst/chart/crds/blueprint.yaml` — accepts (live K8s 1.30).
- [x] Valid sample + bp-legacy-test admitted: `kubectl apply --dry-run=server -f products/catalyst/chart/crds/tests/blueprint-sample-valid.yaml` → 2 created.
- [x] All 7 invalid samples rejected with admission errors at the precise field path:
  - `spec.topology.supported[1]: Unsupported value "foo-bar"`
  - `spec.topology.defaults.single-region: Required value`
  - `spec.endpoints[0].protocol: Unsupported value "smtp"`
  - `spec.endpoints[0].name: Invalid value "BadCaps": ... should match '^[a-z][a-z0-9-]{0,30}[a-z0-9]$'`
  - `spec.endpoints[0].hostnameTemplate: ... should be at least 1 chars long`
  - `spec.sso.realm: Unsupported value "random-realm": supported values: "sovereign", "{{.OrgSlug}}"`
  - `spec.multiInstance.isolationLevel: Unsupported value "vm"`
- [x] 8 representative in-tree blueprints (loki / grafana / gitea / harbor / keycloak / openbao / cnpg / cilium) admit cleanly against the new CRD.
- [x] 83/87 of-tree blueprint.yaml files validate clean (4 pre-existing failures unrelated: placementSchema bounds in `bp-{dmz,rtz}-vcluster`, bare-string `depends[]` in `powerdns-admin` + `sso-bridge` — captured for follow-up).
- [ ] **After Flux reconciles 1.4.438 on hw86**, verify:
  ```
  kubectl --kubeconfig <hw86> get crd blueprints.catalyst.openova.io \
    -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties}' \
    | jq 'keys' | grep -E "topology|endpoints|sso|multiInstance"
  ```
- [ ] Founder walks `kubectl apply` of the G117 invalid samples directly on hw86 to confirm live admission rejects.

## Application CRD note

`products/catalyst/chart/crds/application.yaml` already carries `spec.instanceId` (with CEL `x-kubernetes-validations` immutability rule `oldSelf == "" || self == oldSelf`), `spec.isolationLevel` (enum: namespace / vcluster), and `spec.namingTemplate` from PR #2795 (G117.2 W2.C2). No change needed there. The A8 verifier's "missing fields" finding for Application is likely a stale-CRD-on-hw86 issue that resolves once 1.4.438 reconciles.

Refs #2740 #2745 #2737

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -898,7 +898,11 @@ spec:
       # patch can flip catalyst-api image to harbor-proxy URL passing
       # the harbor-proxy-pull admission policy. Kustomize-mode sibling
       # api-deployment-kustomize.yaml (literal) handles contabo-mkt.
-      version: 1.4.440
+      # 1.4.441 (G117.E2E-B3 #2740 #2745 #2737, 2026-06-03): promote
+      # G117 fields (spec.topology / endpoints / sso / multiInstance)
+      # into the Blueprint CRD openAPIV3Schema so runtime admission
+      # gates them — preserve-unknown no longer silently swallows.
+      version: 1.4.441
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/products/catalyst/chart/Chart.yaml
+++ b/products/catalyst/chart/Chart.yaml
@@ -2367,21 +2367,31 @@ name: bp-catalyst-platform
 # Blueprint placementSchema.defaultOnMultiRegion drives empty-
 # spec.placement on multi-region Sovereigns.
 # (Bumped 414 → 415 after main concurrently shipped a 1.4.414.)
-# 1.4.436 (G117.E2E-B3 #2740 #2745 #2737, 2026-06-03): promote the G117
-# fields (spec.topology, spec.endpoints, spec.sso, spec.multiInstance)
-# from JSON-schema-only into the Blueprint CRD openAPIV3Schema in
-# products/catalyst/chart/crds/blueprint.yaml. W1.B1 + W2.C2 shipped
-# the Go types + JSON-schema gate but the CRD manifest never gained
-# these fields — runtime admission silently swallowed them through
-# the parent spec's x-kubernetes-preserve-unknown-fields. This bump
-# makes admission gate the canonical contract (variant-specific
-# required keys via inlined per-variant sub-schemas; K8s structural
-# schemas forbid $ref so each of the 4 BCP variants repeats the
-# same three sub-blocks via YAML anchor). 83/87 of-tree blueprints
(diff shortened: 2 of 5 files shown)


emrahbaysal:
PR #2825 currently has merge conflicts against `main` on `products/catalyst/chart/Chart.yaml` (version) and `clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml` (slot pin) due to concurrent auto-bumps from #2823 → #2831 → #2833. Local attempt to rebase + resolve while keeping all 4 G117 schema fields hit conflicts on related branches in the shared worktree because parallel F-agents are touching neighboring files. Recommended path: a fresh PR opening the same schema bump on top of current main (1.4.440 → 1.4.441) without the historical comment-block additions to avoid the conflict surface. Refs #2740 #2825

Issue #85, opened 2026-04-28
[E] provisioner: implement REST endpoint POST /v1/sovereigns


emrahbaysal:
Parent: #43 — Catalyst-Zero consolidation + first Hetzner Sovereign at omantel.omani.works waterfall.

Group E.

Validates configSchema. Persists Sovereign CR via store/. Returns id + SSE channel URL.

Reference: [`docs/PROVISIONING-PLAN.md`](docs/PROVISIONING-PLAN.md)

emrahbaysal:
Closed: implemented as POST /api/v1/deployments (handler.CreateDeployment) at products/catalyst/bootstrap/api/cmd/api/main.go:37. Naming differs from the original ticket (deployments vs sovereigns) but the contract is identical — validates request, persists state via in-memory store, returns id + streamURL pointing at the SSE channel. Per request validation including configSchema-equivalent input validation in handler.CreateDeployment.

PR #3083, opened 2026-06-07
fix(marketplace): fold pending-voucher credit into checkout (100%-voucher → credit-only path)

emrahbaysal:
**Root** (traced via the now-confirmed-reachable marketplace, independent-reviewer-led): a 100%-voucher order rendered the Stripe card path because `creditCovers` (CheckoutStep.svelte) gated on the committed `credit_ledger` balance only — but the voucher commits at the `/billing/checkout` POST, so the pre-checkout ledger balance is 0 → `creditCovers=false` → card path, even though the backend `CreditOnlyCheckout` covers it. **Fix:** the credit `$effect` now folds in the pending voucher's preview credit via a new `redeemVoucherPreview` binding (`POST /billing/vouchers/redeem-preview`, normalised to baisa per #85). A covering voucher → `creditCovers=true` → credit-only path. Unblocks the org-creation walk gating #2742-POST + #2744-per-Org-realm. `Refs #3057`. Owed: live UI walk on the reachable marketplace.


emrahbaysal:
Closing — this branch picked up git conflict markers from a stash-pop against my stale local main (the voucher-credit `$effect` region overlaps #3076's creditError block). The fix logic is sound (root-cause + design documented on #3057); it will be redone cleanly off fresh origin/main. Not merging a conflicted commit.

Issue #3104, opened 2026-06-07
P0: dependency-graph-audit RED on main — bp-netbird (slot 53) declared in expected-bootstrap-deps.yaml but missing from _template/bootstrap-kit (blocks ALL PRs)


emrahbaysal:
## The audit is red on main — every PR's `dependency-graph-audit` fails

`scripts/check-bootstrap-deps.sh` (CI gate `dependency-graph-audit`) reports EXIT=1 on current origin/main:
```
WARN: HR 'bp-netbird' (slot 53, wave 'present') declared expected but not yet on disk
ERROR: 1 drift(s) detected. Reconcile HR files with scripts/expected-bootstrap-deps.yaml
```

### Cause
- `scripts/expected-bootstrap-deps.yaml:476-481` declares `bp-netbird` at **slot 53, wave `present`** (= must be on disk now, NOT a deferred W2.K wave).
- The audit checks `clusters/_template/bootstrap-kit/` — there is **no `53-bp-netbird.yaml` there**. It was removed in #1289 ("charts never published").
- BUT the HR exists in an instance: `clusters/omantel.omani.works/bootstrap-kit/53-bp-netbird.yaml`, and the Blueprint is live (`platform/netbird/` chart + blueprint.yaml). So netbird is NOT deprecated — the `_template` copy is just missing.

### Impact
Blocks merge of EVERY PR (verified on #3101, #3102) — the audit is a required check and the PR merge-commit inherits main's drift. Not caused by any of those PRs; their own content passes the audit (Drift 0 in-branch).

### Fix (pick one — needs a bootstrap-kit owner's call)
1. **Add** `clusters/_template/bootstrap-kit/53-bp-netbird.yaml` (templatize the omantel instance copy) — if netbird IS meant to be a present bootstrap-kit HR. -OR-
2. **Move** bp-netbird to a deferred wave (W2.K4) in `expected-bootstrap-deps.yaml` if its chart isn't ready for all provs. -OR-
3. **Remove** the slot-53 declaration from `expected-bootstrap-deps.yaml` if netbird is per-instance-only (not in the template).

Most likely (1) given the live chart + instance HR. Once fixed, #3101 (per-Org-realm controller, #3084) and other queued PRs unblock immediately.

emrahbaysal:
## CORRECTION to my fix-direction — bp-netbird is a DEFAULT-OFF *gated* HR, not a removable stale entry

[...]

PR #3109, opened 2026-06-08
fix(cloud-graph): render all topology.regions not just MaterializedRegions[0] (#3106)

emrahbaysal:
Refs #3106

## Side with the bug: API loader (chroot region source), not the console adapter

The console cloud-graph adapter (`hierarchyToGraph` / `mergeGraphs`) already iterates **all** `topology.regions` and never drops a region — so the renderer was correct. The bug was that the **chroot** catalyst-api fed it a 1-region topology.

On `console.hw101.omantel.biz` (the Sovereign's own console = the chroot path) the topology is built from `chrootEnsureDeployment` → `chrootRegionsFromEnv()`, which read **only** `SOVEREIGN_REGIONS_JSON`. On hw101 (`e19b083c6db41bb0`) that ConfigMap key (`regionsJson`) was empty `[]`, so `chrootRegionsFromEnv()` returned nil, `buildTopology` fell into the **single-cluster live-Nodes path** (which only probes the in-cluster region-a Nodes), and `/cloud?view=graph` rendered **`Region 1/1`** — only `me-east-215-a` — despite the substrate genuinely having both clusters (region-a + region-b, 4+4 Ready nodes, two kubeconfigs on the PVC).

Live-verified on hw101 `sovereign-fqdn` ConfigMap:
```
regionsJson       = []                              ← the wedge
primaryRegion     = hw-me-east-215-a-rtz-prod       ← populated
replicaRegion     = hw-me-east-215-b-rtz-prod       ← populated
enableHotStandby  = true                            ← populated
```

## Fix (`internal/handler/jobs.go`, `chrootRegionsFromEnv`)

When `SOVEREIGN_REGIONS_JSON` is unset / empty / `[]`, reconstruct the region list from the discrete `SOVEREIGN_PRIMARY_REGION` + `SOVEREIGN_REPLICA_REGION` env vars (replica gated on `SOVEREIGN_ENABLE_HOT_STANDBY`). These are already wired by `bp-catalyst-platform` from the same ConfigMap and are reliably populated on a 2-region hot-standby Sovereign. `buildTopology` then emits one Region per entry, so the cloud-graph shows **both** regions.

- before: `if raw == "" { return nil }` → single-cluster live-Nodes path → Region 1/1
- after: empty/`[]`/malformed `regionsJson` falls through to primary+replica reconstruction → 2 regions

**Contained + additive** — populated `regionsJson` still wins; no-primary still returns nil so the legacy single-cluster live-Nodes fallback is preserved. This is a **view-completeness** fix only: node/SKU detail still flows from the live K8s SSE snapshot the graph merges on the UI side. It does **NOT** touch the deeper cross-region provisioner materialization (`MaterializedRegions`) — that's a separate G117 item.

The region-count chip ("Region N/M") is derived downstream from `topology.regions.length`, so no separate chip-filter change was needed — fixing the region source fixes the chip.

## Tests
- `TestChrootRegionsFromEnv_PrimaryReplicaFallback` — 9 cases incl. the exact hw101 shape + N-region-in → N-region-out, hot-standby off → 1, regionsJson-wins, malformed-falls-through.
- `TestEnvTruthy` — boolean-env parsing for the replica gate.
- `TestChrootEnsureDeployment_TwoRegionsFromPrimaryReplica` — end-to-end: synthesised `Request.Regions == 2` (the data the cloud-graph renders).

`go build ./...` + `go test ./...` green in `products/catalyst/bootstrap/api`.

Verification: PENDING — live re-walk hw101 /cloud?view=graph must show 2 regions

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/jobs.go
+++ b/products/catalyst/bootstrap/api/internal/handler/jobs.go
@@ -149,21 +149,71 @@ func (h *Handler) chrootEnsureDeployment(depID string) *Deployment {
 }
 
 // chrootRegionsFromEnv parses SOVEREIGN_REGIONS_JSON (the canonical
-// regions list threaded in by bp-catalyst-platform). Returns empty
-// slice when unset or malformed — the caller falls back to the live-
-// Nodes path. The JSON shape matches the wizard's request.regions
-// array: [{provider, cloudRegion, controlPlaneSize, workerSize,
-// workerCount}, ...].
+// regions list threaded in by bp-catalyst-platform). The JSON shape
+// matches the wizard's request.regions array: [{provider, cloudRegion,
+// controlPlaneSize, workerSize, workerCount}, ...].
+//
+// Fallback (#3106): when SOVEREIGN_REGIONS_JSON is unset, empty, or an
+// empty JSON array, reconstruct the region list from the discrete
+// SOVEREIGN_PRIMARY_REGION / SOVEREIGN_REPLICA_REGION /
+// SOVEREIGN_ENABLE_HOT_STANDBY env vars (also threaded in by
+// bp-catalyst-platform from the sovereign-fqdn ConfigMap). On a
+// 2-region hot-standby Sovereign the per-Sovereign overlay reliably
+// populates primaryRegion + replicaRegion + enableHotStandby even when
+// the richer regionsJson list was never wired (caught live on hw101
+// e19b083c6db41bb0 2026-06-08: regionsJson=[] but primaryRegion=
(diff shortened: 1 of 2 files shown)


emrahbaysal:
Merging: own checks pass (go build/test + 3 new tests). Only red check is the pre-existing #3104 bp-netbird gated-HR audit mismatch (admin-merged past by #3099-#3102/#3108), not this PR. #3106 LIVE-walk verification still pending: roll hw101 catalyst-api to this image + re-walk /cloud?view=graph for 2 regions.

PR #3172, opened 2026-06-09
fix(sso): guacamole canonical host + ssoInitPath (#3150)

emrahbaysal:
Durable fix for guacamole silent-SSO launch (was a live-only CR patch).

## Evidence-resolved hostname
The blueprint/seed said `guac.<fqdn>` but the **live deployment + OIDC contract** both use `guacamole.<fqdn>`:
- HTTPRoute `guacamole-server` → `guacamole.hw124.omani.works`
- Keycloak client `guacamole` redirectUris → `https://guacamole.hw124.omani.works/*`

So `guac.` was a stale bug (404, dead host). Fixed to `guacamole.<fqdn>` in **both** the platform blueprint and the catalog-seed (lockstep).

## ssoInitPath
`/guacamole/` — the app's context path. The legacy app-root `/` 404s; the Angular SPA + OpenID extension drive the silent redirect from `/guacamole/`.

## ⚠️ Separate blocker (NOT fixed here — flagged)
On hw124, guacamole's OIDC env points at the **stale realm `catalyst` / host `keycloak.*`** instead of `sovereign` / `auth.<fqdn>`. Silent SSO will not complete until the guacamole chart's OIDC env is realigned to the sovereign realm. This PR fixes the *launch path*; the OIDC-env realignment is a separate config-drift fix.

Refs #3150

### DIFF
--- a/clusters/_template/bootstrap-kit/52-bp-guacamole.yaml
+++ b/clusters/_template/bootstrap-kit/52-bp-guacamole.yaml
@@ -175,9 +175,14 @@ spec:
         # bootstrap-kit Kustomization renders.
         hostname: guacamole.${SOVEREIGN_FQDN}
       oidc:
-        # Same SOVEREIGN_FQDN substitution. The realm name (`catalyst`)
-        # matches bp-keycloak's realm-config-cli default.
-        issuer: https://keycloak.${SOVEREIGN_FQDN}/realms/catalyst
+        # Same SOVEREIGN_FQDN substitution. Canonical issuer is the
+        # EXTERNALLY-reachable Keycloak host `auth.<fqdn>` + the `sovereign`
+        # realm (where the catalyst-pin IdP lives). The prior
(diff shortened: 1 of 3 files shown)


emrahbaysal:
Heads-up from #3177: I shipped the same guacamole issuer fix (`auth.${SOVEREIGN_FQDN}/realms/sovereign`) in #3177 along with two openbao sso-configure host bugs that also defeat silent SSO (BAO_ADDR NXDOMAIN + allowed_redirect_uris wrong host). Our `52-bp-guacamole.yaml` issuer edits target the identical value so no semantic conflict — whichever merges first, the other rebases trivially. #3172 still uniquely owns the guacamole launch-url host (`guac.` -> `guacamole.`) + `ssoInitPath` in catalog-seed/blueprint. NB: #3172's CI is red on banned words in the PR body/commit messages — those need an edit to merge.

Issue #3195, opened 2026-06-09
Pillar-3: cross-region topology enablement (bp-cnpg-pair slot + ReplicaCluster + Continuum seed, safe-by-default) + continuum image-pull fix


emrahbaysal:
## Problem

`docs/sessions/2026-06-10/pillar3-multiregion-design/DESIGN.md` (Deliverable C of #3189) showed a live 2-region Sovereign (hw124) is **two fully-independent region stacks** — the cross-region DR layer (Pillar-3 active-hot-standby) is entirely absent at runtime. 6 gaps:

- **G1** — no `bp-cnpg-pair` installed (only `bp-cnpg` operator)
- **G2** — `bp-cnpg-pair` is NOT in the bootstrap-kit → fresh prov never installs it
- **G3** — Cilium ClusterMesh A↔B not connected
- **G4** — no global services (`service.cilium.io/global: "true"`)
- **G5** — all CNPG clusters standalone (no `ReplicaCluster`)
- **G6** — no `Continuum` CR seeded

## Root cause (verified)

- `clusters/_template/bootstrap-kit/` has `62-bp-continuum.yaml` but **no cnpg-pair slot** (G1/G2). The chart at `platform/cnpg-pair/chart/` is complete (renders primary Cluster + ReplicaCluster + `*-mesh` global Service + audit ConfigMap, all gated `cnpgPair.enabled`) but is never deployed.
- G3 (ClusterMesh) is **already automated** — `products/catalyst/bootstrap/api/internal/handler/clustermesh.go` `AutoEstablishClusterMesh` auto-fires from `phase1_watch.go:766` on multi-region. Not in scope to re-build.
- G4/G5 are subsumed by deploying cnpg-pair (its chart renders both).
- G6 — only the QA-fixtures Continuum CR exists (`qaFixtures`, QA-only); no production seed.
- **Continuum image-pull bug** (surfaced this session): `62-bp-continuum.yaml` doesn't set `global.imageRegistry` → controller pulls `ghcr.io/...continuum-controller` direct → continuum-controller runs in `catalyst-system` (NOT in the kyverno `complianceDefaultExcludes` list) → `harbor-proxy-pull` Enforce policy + 401.

## Tasks (code only this phase — NO firing a 2-region prov)

- [ ] Add `bp-cnpg-pair` bootstrap-kit slot, gated `SOVEREIGN_ENABLE_HOT_STANDBY` (default-OFF → single-region renders nothing)
[...]

PR #3253, opened 2026-06-10
fix(cnpg-pair): split-side topology — each cluster renders its own half of the pair (0.2.0)

emrahbaysal:
## Pillar-3 core topology fix

A 2-region Sovereign is **two SEPARATE k3s clusters** joined by Cilium ClusterMesh (kom4dc mimics region 2 with a second VPC) — not one stretched cluster. bp-cnpg-pair ≤0.1.x declared the whole pair in ONE release on the primary cluster (slot 16b suspended on secondaries) and relied on node-affinity to "schedule each side to its region". That only works on a stretched cluster.

**Live evidence (hw126):**
```
failover-readiness pod: FailedScheduling 0/4 nodes — affinity requires openova.io/region=hw-me-east-215-b-rtz-prod
(cluster-A's 4 nodes are ALL hw-me-east-215-a-rtz-prod)
```
The replica Cluster CR + failover-readiness Deployment pinned to region-B labels can NEVER schedule on cluster-A. The replica's own design (pg_basebackup + externalClusters streaming WAL over the `*-primary-mesh` global Service) is precisely CNPG's separate-cluster replica pattern — the objects just need to be applied ON cluster-B.

### Chart 0.1.5 → 0.2.0 (`platform/cnpg-pair/`)
- New **`cnpgPair.side: primary|replica`** (default `primary`). `secondary` is accepted as an alias for `replica` — the per-region cloud-init substitute `SOVEREIGN_REGION_ROLE` has value domain `primary|secondary` (infra/providers/*/main.tf `sovereign_region_role`), so the slot substitutes it verbatim and the chart normalizes. Invalid values fail the render.
- **side=primary** renders: primary Cluster CR + CNPG-managed `-primary-mesh` global Service + audit-config ConfigMap (bp-continuum slot 62 runs on the primary cluster and its prerequisite probe `kubectl get cm -A -l catalyst.openova.io/audit-config=true` reads its OWN cluster — contract preserved) + replication-ingress NetworkPolicy + optional Continuum CR + helm-test resources.
- **side=replica** renders: replica Cluster CR (unchanged pg_basebackup + externalClusters design) + **a local `-primary-mesh` Service stub** + failover-readiness Deployment (#3247 kyverno-compliance preserved; its region-B affinity now matches the cluster it runs on) + the 2 probe NetworkPolicies.
  - Why the Service stub: Cilium ClusterMesh merges global services **by name+namespace across clusters** — a Service object must exist in the consuming cluster too, annotated `service.cilium.io/global`. Without it the replica's `externalClusters.host` (`<fullname>-primary-mesh`) is NXDOMAIN on cluster-B. The stub's selector (`cnpg.io/cluster: <fullname>-primary`) matches zero local Pods, so every connection crosses the mesh to the primary's backends. No CNPG ownership collision (the 0.1.0 incident was a same-cluster `-r` collision; the primary CR doesn't exist on cluster-B).
- `validateRegions` now also runs on the replica-side render.
- **SAFE-BY-DEFAULT preserved:** `cnpgPair.enabled=false` → ZERO resources on BOTH sides (byte-identical empty renders, asserted by the render gate).
- Version lockstep: Chart.yaml 0.2.0 + blueprint.yaml 0.2.0 + slot 16b pin 0.2.0 (`check-bootstrap-kit-pin-sync.sh` green).

### Slot 16b (`clusters/_template/bootstrap-kit/16b-bp-cnpg-pair.yaml`)
- `suspend: ${SECONDARY_HR_SUSPEND:=false}` **removed** — both control planes apply the HR; `side: ${SOVEREIGN_REGION_ROLE:-primary}` selects the half.
- Header + kustomization.yaml comment rewritten to the split-side model so the single-cluster affinity claim cannot be reintroduced.

### Cloud-init (`infra/providers/_shared/cloudinit-control-plane.tftpl`)
- `SOVEREIGN_REGION_ROLE: "${sovereign_region_role}"` **promoted from the huawei-only block to the SHARED substitute map** (it already existed at the old ~line 451; both providers pass the var — hetzner main.tf stamps `primary`/`secondary` per CP at lines 892/1321, huawei derives it from the region index). Dash-safe: `scripts/lint-cloudinit.sh both` green (46+43 runcmd items pass `dash -n`); `tofu validate` green for both providers.

### catalyst-api flip — BOTH regions (`products/catalyst/bootstrap/api/internal/handler/clustermesh.go`)
The #3238 flip previously patched only the PRIMARY cluster's bootstrap-kit Kustomization. Under split-side that would activate side=primary while cluster-B keeps rendering an empty release (no replica, no WAL stream). `enableCNPGPairAfterFullMesh` now:
- takes the full `regionSlots` (each carries its region's kubeconfig path — on the full-success path that triggers the flip, every slot is reachable) and patches `SOVEREIGN_ENABLE_CNPG_PAIR=true` + `reconcile.fluxcd.io/requestedAt` onto **every region's** Kustomization;
- gather phase is **all-or-nothing**: any unreadable Kustomization or failed region-substitute precondition refuses the whole flip (a half-flipped pair is exactly the hw126 broken topology); the level-triggered reconcile (#3241) / post-handover re-run converges;
- patch phase is per-region best-effort after a clean gather (idempotent merge patch; re-run converges a transiently missed region).

## Validation
- `chart/tests/cnpg-pair-render.sh` rewritten for the side split — **11 cases green**: disabled byte-identical-empty both sides; side=primary = 3 non-test + 4 helm-test resources; side=replica = 5 non-test + 0 test; `secondary` alias render byte-identical to `replica`; empty-tag + same-region fail-fasts on both sides; invalid-side fail-fast; clusterMesh-off omits the global Service on both sides; async omits sync params; Continuum CR exactly 1 on side=primary and 0 on side=replica; continuum fail-fasts unchanged.
- `go build ./...` + `go vet` + full handler test package green (64s). Extended dynamicfake harness (#3243/#3245 idiom): fixture seeds per-region dyn clients; full-mesh flip asserted on BOTH regions; new `TestAutoEstablishClusterMesh_SecondaryKustomizationMissingBlocksFlip` covers the all-or-nothing refusal; the precondition test now asserts the healthy secondary also stays unflipped; retry-convergence + startup-restore tests assert both regions.
- `check-bootstrap-kit-pin-sync.sh`, `check-bootstrap-deps.sh` (dependsOn unchanged), `check-tofu-flux-envsubst.sh`, `check-no-banned-words.sh` — all green locally.

## Honest limits
- Cross-cluster WAL streaming, the replica basebackup over the mesh, and the replica-side Secret sync (`-replication`/`-ca` must be reflected cluster-A → cluster-B, the pre-existing 0.1.x contract now made explicit) are **only provable on a live 2-region prov** — static renders + unit tests cannot cover them. The next fresh 2-region prov walk is the acceptance gate.
- `kubectl get hr -A` must be re-checked on the live env after merge per the #3188-wave rule (CI-green ≠ prov-green).

Refs #3236 #3195 #2737

### DIFF
--- a/clusters/_template/bootstrap-kit/16b-bp-cnpg-pair.yaml
+++ b/clusters/_template/bootstrap-kit/16b-bp-cnpg-pair.yaml
@@ -1,17 +1,42 @@
 # bp-cnpg-pair — Catalyst bootstrap-kit Blueprint slot 16b.
 # (Pillar-3 cross-region DR primitive — EPIC-6 slice C-DB-1 #1101.)
 #
-# Active-hot-standby CNPG cluster-pair across two Sovereign regions:
+# Active-hot-standby CNPG cluster-pair across two Sovereign regions.
+#
+# ─── SPLIT-SIDE MODEL (chart 0.2.0) — each cluster renders ITS half ──
+# A 2-region Sovereign is TWO SEPARATE k3s clusters joined by Cilium
+# ClusterMesh (kom4dc mimics the regions with 2 VPCs) — NOT one
+# stretched cluster. So this HR is applied on BOTH control planes and
+# `cnpgPair.side` (stamped from the per-region SOVEREIGN_REGION_ROLE
+# substitute below) selects which half each cluster renders:
+#   side=primary (region-A cluster):
 #   - Primary postgresql.cnpg.io/v1.Cluster CR pinned to region A
-#   - Replica Cluster CR pinned to region B (replica.enabled: true +
-#     bootstrap.pg_basebackup + externalClusters[] streaming the
-#     primary's WAL over Cilium ClusterMesh)
 #   - `<name>-primary-mesh` Service annotated
 #     `service.cilium.io/global: "true"` so ClusterMesh publishes the
 #     primary's read-replica endpoint across every connected peer
-#   - failover-readiness probe + audit-config ConfigMap (label
+#   - audit-config ConfigMap (label
 #     `catalyst.openova.io/audit-config: "true"`) that bp-continuum's
-#     prerequisite-check probes to switch from idle to active.
+#     prerequisite-check probes to switch from idle to active
+#     (bp-continuum slot 62 runs on the primary cluster).
+#   side=replica (region-B cluster, SOVEREIGN_REGION_ROLE=secondary —
+#   the chart normalizes "secondary" → replica):
+#   - Replica Cluster CR pinned to region B (replica.enabled: true +
+#     bootstrap.pg_basebackup + externalClusters[] streaming the
+#     primary's WAL over Cilium ClusterMesh)
+#   - local `-primary-mesh` Service stub (Cilium merges global
+#     services by name+namespace across clusters; zero local backends
+#     → traffic crosses the mesh to the primary)
+#   - failover-readiness probe (its region-B node-affinity now matches
+#     the cluster it runs on).
+#
+# Chart ≤0.1.x declared the whole pair ONCE on the primary cluster
+# (this slot suspended on secondaries via SECONDARY_HR_SUSPEND) and
+# relied on node-affinity to "schedule each side to its region" — that
+# only works on a single stretched cluster. On hw126 the replica-side
+# workloads could NEVER schedule: FailedScheduling 0/4 nodes, affinity
+# requires openova.io/region=hw-me-east-215-b-rtz-prod while every
+# cluster-A node is ...-a-rtz-prod. 0.2.0 removes the suspend and
+# splits the render per side.
 #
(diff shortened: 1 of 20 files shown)


emrahbaysal:
Follow-up commit e5ddd00b folded in: cloud-agnostic storageClass default (empty = cluster default SC; `hcloud-volumes` only exists where bp-hcloud-csi runs — on kom4dc/Huawei the pair's PVCs would Pending forever on "storageclass not found"). Both Cluster templates now `with`-guard the field; blueprint.yaml schema defaults in lockstep. All 11 render gates + helm lint + blueprint schema validation re-run green.

PR #3303, opened 2026-06-11
fix(clustermesh): two-stage cnpg-pair flip — kill the first-flip deadlock on fresh envs

emrahbaysal:
Found live on hw128 right after #3299 brought the agent tunnel up: the reconcile loop refused the cnpg-pair flip every ~2 min ("source Secret unavailable on primary") while `cnpg` stayed empty in both regions — forever. Structural first-flip deadlock: the #3254 sync gated ALL patches on Secrets that only exist after the primary postgres runs, and that postgres only renders once the gate flips. hw126 predates #3254's gate, which is why it never showed this.

Fix: two-stage patch — primary-side regions flip unconditionally (the primary is what mints the Secrets); replica-side regions stay gated on the sync and converge on the level-trigger re-run once initdb finishes. Region side comes from the `SOVEREIGN_REGION_ROLE` substitute (fallback: slot 0 = primary, the chart's `:-primary` default). The original guard (replica gate ON without auth Secrets = broken topology) is preserved.

Tests: new `TestEnableCNPGPair_TwoStageFirstFlip` two-run contract; `MissingSourceSecret`/`ReplicaCopyFailure` updated to the two-stage semantics. `go test -race ./...` green.

Self-applies to live hw128 on the post-merge mothership roll (same mechanism as #3299) — expected sequence: primary gate ON → primary postgres initdb → Secrets minted → sync → replica gate ON → WAL streaming → region-kill walk.

Refs #3241

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/clustermesh.go
+++ b/products/catalyst/bootstrap/api/internal/handler/clustermesh.go
@@ -87,6 +87,7 @@
 package handler
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"crypto/rsa"
@@ -99,6 +100,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


emrahbaysal:
Pushed a second commit covering **#3241 layer 4**, also found live while watching layer 3: the level-trigger loop unconditionally rollout-restarted cilium + cilium-operator + clustermesh-apiserver on EVERY ~2-min pass (apiserver Deployment hit generation 35 on hw128) — the agents never got a stable window to finish the remote-config sync. Restarts are now keyed on actual change (byte-compare of the peer Secret, DeepEqual of the hostAliases); idempotent re-runs skip them. New regression test locks it in. Both layers self-apply to hw128 on the same post-merge roll.

Issue #3379, opened 2026-06-12
SOVEREIGNTY: cutover earns a durable, revert-immune cutoverComplete=true under a witnessed 600s deny-egress hold — engine + blocker fixes merged, gated on the running cutover re-prov


emrahbaysal:
**Problem** — Prove the sovereignty cutover end-to-end on a live throwaway Sovereign: the handover arms the cutover, the 11-step engine runs green, and `cutoverComplete=true` is earned **only** under a witnessed 600s deny-egress hold (ADR-0002 §3/§4, DOD §7, Principle #11) — a TRUE, DURABLE, revert-immune proof of independence, not a green checkmark over a still-tethered cluster.

**Current state** — All engine + blocker fixes are **merged**; the proof is gated only on the running cutover re-prov.

- Five-Faces durability rebuild — durable revert-immune `cutoverComplete` fact, faithful registry pivot (`registriesYamlActive=v2`), default-deny-with-allow-list egress hold, audit-timestamp fidelity, host GitOps loop + zero residual tether — **merged** in **PR #3699**, with follow-ons **PR #4269** (apply default-deny CCNP under `defaultDenyEgress:true` so the hold is witnessed, not assertion-only) and **PR #4380** (plane-isolation carve-out so cutover Jobs reach host-ns `{gitea,harbor,openbao}` after #4325 de-vcluster). Sub-issues #3667 #3671 #3678 #3681 #3695 are CLOSED.
- The three faults that wedged the live walk this session are now **merged**:
  - **#4527 → PR #4532** — registry-pivot v1 cold-start mirror now authenticates private `ghcr.io` pulls (no anonymous-401 wedge of `api.<fqdn>`).
  - **#4529 → PR #4531** — harbor-prewarm skopeo no longer public-CA-verifies the in-cluster Harbor dest (unwedges step-03 on an LE-STAGING cert).
  - **#4543 → PR #4544** — mothership `postTofuArchive` skips public-CA verify on the mother→child archive hop (LE-staging child cert no longer x509-fails the seal that arms the cutover).

**Acceptance criteria** — on a fresh throwaway Sovereign, current `main`:
- [ ] Handover seals `secret/catalyst/tofu-phase0-archive` in the child OpenBao; the cutover **arms** (trigger no longer 425).
- [ ] The 11 cutover steps run green through registry-pivot, harbor-prewarm, helmrepository-patches, and step-08.
[...]

PR #3557, opened 2026-06-14
fix(cutover): step-03 harbor-prewarm DeadlineExceeded + resume re-runs a timed-out step (Refs #3379, Refs #3526)

emrahbaysal:
## Why

Two stacked defects from the live **hw139 #3379 cutover wedge** — both block the Pillar-5 cutover from completing on a fresh prov. Develop-and-merge for the **NEXT** fresh prov; **validated-on-next-fresh-prov, NOT walked this session** (hw139 is being region-killed and fresh fires hit the kom4dc EIP wall).

## Defect 1 — Step-03 harbor-prewarm `DeadlineExceeded` (only 6/29 pushed)

The ~29 multi-layer **PRIVATE** `openova-io/*` images are skopeo-native-pushed `ghcr.io`→local Harbor (the **only** working route — Harbor's anonymous proxy-cache 404s these private packages, #3534). They ran **strictly serial** over kom4dc's throttled egress and tripped the step deadline mid-copy → local Harbor `openova-io` project incomplete → post-cutover `ImagePullBackOff` → Step-08 deny-egress proof never ran.

**Chosen approach (a) — parallelize + raise the ceiling — over (b) sourcing from the node's containerd cache (#3534 prepull):**
(b) is not reliably correct. The prewarm Pod is hardened (`runAsNonRoot`, `runAsUser 1000`, `readOnlyRootFilesystem`, `drop: [ALL]`), so reading the root-owned containerd socket needs a hostPath mount + privilege escalation (a security regression). The #3534 prepull only caches the `openova-io/openova` **subset**, runs as a detached background `nohup` that may not have finished, and runs **only on the control-plane node** while the prewarm Pod may schedule on a worker. (a) keeps the existing authenticated skopeo path and just removes the wall-clock bottleneck.

- **`03-harbor-prewarm-job.yaml`** — Phase-A copy loop fans out `PREWARM_PARALLELISM` (default 6) concurrent `skopeo copy` workers (each writes its own `.ok`/`.fail` sentinel + log; parent `wait`s + tallies), ~6× wall-clock collapse. Each copy gains `--retry-times 3`. POSIX-clean (`dash`/`sh`/`busybox -n` + a mock run under dash+busybox verified); live-PID-counting throttle (no `wait -n`/arrays). The two earlier #3379 hw138 catches stay in force (`--insecure-policy` + branch on skopeo's **real** per-worker exit status, not a piped `sed` whose status is always 0).
- **`values.yaml`** — `stepTimeouts.harborPrewarmSeconds` `1800`→`5400` (90m). Event-driven watch means a fast run never waits out the ceiling.
- **`cutover.go`** — NEW `cutoverStepDeadline()` makes `createCutoverJob` + `watchJobToCompletion` honor the step's **own** PodSpec `activeDeadlineSeconds` (the chart `stepTimeouts.<step>Seconds`) when it exceeds the global default. Without this the bump is **inert**: the Job-level `activeDeadlineSeconds` was always stamped from the global 15m `cutoverStepTimeout` (authoritative over the Pod-template one), silently killing every long step at 15m with reason `DeadlineExceeded`. Takes the **MAX** so the global stays a floor for short steps.

## Defect 2 — resume/re-fire never retried a Failed step

`cutover.go` resume reset only `result==running` rows, and `runCutoverStep`'s `findExistingTerminalJobForStep` refused to re-create a Job for a Failed step ("operator intervention required"). A `DeadlineExceeded` step-03 therefore re-failed on **every** re-fire and wedged the cutover forever.

- NEW `jobFailedTransiently()` classifies a `JobFailed` with reason `DeadlineExceeded` (or matching message) as **transient**; everything else (`BackoffLimitExceeded`, …) stays **genuine** (fail-closed). NEW `deleteCutoverJobsForStep()` removes the timed-out Job(s). `runCutoverStep`'s Failed branch now: **transient** → delete + clear the step rows + fall through to mint a fresh Job (idempotent re-run); **genuine** → halt exactly as before. Centralized in `runCutoverStep`, so startup-resume and explicit `/start` re-fire behave identically.

## Lockstep

- `bp-self-sovereign-cutover` chart **0.1.60 → 0.1.61**
- bootstrap-kit slot **06a** pin **0.1.60 → 0.1.61**
- catalyst-api image tag is **auto-stamped by `catalyst-build.yaml`** on merge-to-main (`bootstrap/api` is in its trigger paths) — not hand-pinned.

## Validation

- NEW Go tests: `TestCutoverStepDeadline_HonorsPerStepPodSpecValue`, `TestJobFailedTransiently_ClassifiesDeadlineVsGenuine`, `TestRunCutoverStep_RerunsTransientlyFailedJob` (re-runs + completes + deletes the prior `DeadlineExceeded` Job).
- Full `internal/handler` suite green (incl. the existing genuine-failure-halts test).
- chart `cutover-contract.sh` **25/25** gates green; `helm template` clean; `go build`/`vet`/`gofmt` clean.

**This is validated-on-next-fresh-prov — the live walk is deferred to when a clean prov can fire.**

Refs #3379
Refs #3526

### DIFF
--- a/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
+++ b/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
@@ -462,7 +462,15 @@ spec:
       # 401'd every post-pivot clone, Step-06's rewrites never reconciled,
       # and Step-08's pivot-wait never converged (hw138 keystone catch).
       # RBAC gains secrets:create. (0.1.59 = PR #3535 Defect-A skopeo fix.)
-      version: 0.1.60
+      # 0.1.61 (#3379 + #3526, hw139 step-03 wedge): Step-03 harbor-prewarm
+      # DeadlineExceeded — ~29 serial private-image skopeo copies over
+      # throttled egress blew 1800s with only 6/29 pushed. Fix: parallelize
+      # Phase-A (PREWARM_PARALLELISM=6 concurrent copies + --retry-times 3)
+      # and raise harborPrewarmSeconds 1800→5400. Paired catalyst-api change
+      # makes the Job/watch deadline honor THIS per-step value (was capped at
+      # the global 15m) AND re-runs a DeadlineExceeded step instead of
+      # wedging on it (transient-failure delete+re-run). No step-count change.
+      version: 0.1.61
       sourceRef:
         kind: HelmRepository
         name: bp-self-sovereign-cutover
--- a/platform/self-sovereign-cutover/chart/Chart.yaml
+++ b/platform/self-sovereign-cutover/chart/Chart.yaml
@@ -1,5 +1,31 @@
 apiVersion: v2
 name: bp-self-sovereign-cutover
+# 0.1.61 (#3379 + #3526, hw139 step-03 wedge, 2026-06-15): Step-03
+# harbor-prewarm DeadlineExceeded. The ~29 multi-layer PRIVATE openova-io
+# images are skopeo-native-pushed ghcr.io→local Harbor (the ONLY working
+# route — Harbor's anonymous proxy-cache 404s these private packages,
+# #3534). They ran STRICTLY SERIAL over kom4dc's throttled egress and blew
+# the deadline with only 6/29 pushed → incomplete local Harbor →
+# post-cutover ImagePullBackOff → Step-08 deny-egress proof never ran.
+# TWO fixes here (chart side) + a paired catalyst-api change:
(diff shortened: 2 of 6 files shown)


emrahbaysal:
Superseded by a clean re-cut from origin/main. This branch was accidentally based on an in-flight branch, so its diff carried ~424 unrelated files + a pre-existing newapi lockstep drift that tripped manifest-validation. Re-cut from main with only the 7 intended files.

PR #3570, opened 2026-06-15
fix(catalyst-ui): restore per-app Open button on /apps grid via silent-SSO (Refs #3374)

emrahbaysal:
## What broke

The founder flagged this **very important**: the per-app **Open / Launch buttons** that opened UI apps (openbao, harbor, grafana, gitea, keycloak) in SSO **disappeared** from the `/apps` grid in the operator console. The underlying SSO is NOT broken — direct URLs (`bao.`/`registry.`/`grafana.<fqdn>`) still land signed-in zero-click (verified live on hw139). What was lost is the console's one-click launch affordance on the grid card.

## True root cause

The deployed operator console is `products/catalyst/bootstrap/ui` (React/Vite) — `AppDetail.tsx` with the canonical 7-tab strip (Overview · Topology · Resources · Compliance · Logs · Settings · Members) + Endpoints, and `AppsPage.tsx` for the grid.

- **Grid card** (`AppsPage.tsx`, `status-corner`): the per-card Open button was **deliberately deleted** by the #2743 "V4 verdict" (the in-code comment said *"per-card 'Open' button REMOVED. It opened the raw externalURL with no `prompt=none&kc_idp_hint=catalyst-pin` appended, bypassing silent-SSO"*). It was removed but never re-added via the SSO path, so the grid lost the affordance. The `externalURL` prop stayed threaded into `AppCard` but went unused.
- **App-detail header** (`AppDetail.tsx:606`): the hero Launch button (`variant="hero"`, "Open <App>", gated on `appExternalURL`) **survived** the 7-tab refactor — it was never the regression. Confirmed + already covered by `AppDetail.test.tsx`.
- **Data**: the catalyst-api DOES project `externalURL`. The per-app detail endpoint gates it through `externalURLIfUserUI` + `blueprintHasUserUIEndpoint` (#3224) so only apps with a user-facing UI get a URL (and thus a button); headless/API-only apps are suppressed. The `/v1/sovereign/apps` LIST endpoint projected `externalURL` WITHOUT that gate — a slight asymmetry.

So the regression is the grid card's deleted Open button — the documented "click handler missing on leaf cell" shape.

## The fix

**Grid card (`AppsPage.tsx`)** — restore a **prominent accent-filled Open button** in the card `status-corner`, gated on `externalURL` (headless services never get a dead button), routing through a shared `launchAppViaSSO` helper → `GET /catalyst/v1/apps/{key}/launch-url` (the signed silent-SSO URL), with a raw-URL fallback ONLY on error. This keeps the one-click affordance but lands the operator **already signed in** — directly addressing the concern that motivated the #2743 removal. Opens in a new tab (`noopener,noreferrer`), `preventDefault`/`stopPropagation` so the card Link doesn't also navigate. `AppCard` exported for the render test.

**App-detail header** — confirmed the hero Open button is intact + prominent; no change needed.

**catalyst-api** — extracted `userUIGatePasses` (the #3224 user-UI predicate) from `externalURLIfUserUI` and applied it in `HandleSovereignApps`' `externalURL` projection, so the grid Open button is scoped to the **same** user-UI apps as the AppDetail one (no asymmetry; an API/protocol-only backend that owns an HTTPRoute no longer gets a button). Fails open when the catalog is unwired (chroot/CI), fails closed on catalog 404 — identical policy to the detail-page gate.

## Verification (short of a live roll)

- `npx tsc -b` — clean.
- **UI**: `AppsPage.open-button.test.tsx` (5 new) — Open button present iff `externalURL` set; click → `getLaunchURL(<key>)` → `window.open(signed URL)`; raw-URL fallback when launch-url yields nothing; Open click does NOT navigate the card Link. Plus AppsPage / AppsPage.handover / AppDetail suites: **51 passed**.
- **API**: `TestUserUIGatePasses` (5 new) — ui passes / api-only fails / empty fails closed / catalog-unwired fails open / catalog-404 fails closed. Plus `TestBlueprintHasUserUIEndpoint` + `TestExternalURLIfUserUI_*` green after the refactor; **full `internal/handler` package passes**; `go vet` clean.

## Image pins

`catalyst-ui` + `catalyst-api` tags are bumped **in lockstep** to the merge SHA by the `catalyst-build` deploy job's `values.yaml` auto-bump (it rewrites BOTH `catalystUi` and `catalystApi` to the same SHA + the literal template refs). No hand-pinned SHA — guessing one would be overwritten by the deploy job.

## Live walk deferred

catalyst-ui/api can't roll on the half-pivoted hw139 (the cutover half-pivot blocks new ghcr.io rolls), so **the live one-click-launch walk is deferred to the next clean fresh prov**. This PR verifies at the component-render + API-payload level.

`Refs #3374`.

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/applications.go
+++ b/products/catalyst/bootstrap/api/internal/handler/applications.go
@@ -1718,37 +1718,50 @@ func (h *Handler) externalURLIfUserUI(ctx context.Context, depID, blueprint, tar
 	if u == "" {
 		return ""
 	}
-	bpName := strings.TrimPrefix(strings.TrimSpace(blueprint), "bp-")
-	if bpName == "" {
-		// No blueprint name — no UI signal can ever exist for this app.
-		// FAIL CLOSED (#3224 round-1 fix).
-		return ""
-	}
-	if h.catalogClient == nil {
-		// Catalog client not configured AT ALL (chroot config gap, CI) —
-		// the gate cannot distinguish any app from any other, so
-		// suppressing here would kill every WORKING Launch button
-		// fleet-wide. This is the single deliberate fail-OPEN branch.
-		return u
-	}
-	bp, err := h.resolveBlueprintMeta(ctx, bpName, "")
-	if err != nil || bp == nil || len(bp.Endpoints) == 0 {
-		// FAIL CLOSED (#3224 round-1): transport error, blueprint
-		// NotFound in the catalog, malformed spec, or a genuine
-		// `endpoints: []` declaration (bp-newapi's shape) — in every
-		// case there is NO evidence of a user UI. The old fail-open
-		// here is exactly what rendered dead Open buttons on live
-		// hw130 (the catalog 404'd bp-newapi → button showed anyway).
-		// A suppressed button self-corrects on the next render if the
-		// catalog recovers; a dead button never does.
-		return ""
-	}
-	if !blueprintHasUserUIEndpoint(bp) {
-		// Resolved, endpoints present, but none is a user UI (API/protocol
-		// only) → suppress the dead Open button.
+	if !h.userUIGatePasses(ctx, blueprint) {
 		return ""
 	}
 	return u
+}
+
+// userUIGatePasses decides whether an app with the given Blueprint qualifies
+// for a user-facing "Open" / Launch button (#3224 / #3374). It is the SINGLE
+// SOURCE OF TRUTH shared by:
+//
+//   - externalURLIfUserUI — the per-app detail endpoint's externalURL gate
+//     (GET /sovereigns/{id}/applications/{name}).
+//   - HandleSovereignApps — the /v1/sovereign/apps LIST projection, so the
+//     AppsPage grid Open button is scoped to exactly the same apps as the
+//     AppDetail one (no asymmetry where the grid shows a button the detail
+//     page suppresses, or vice-versa).
+//
+// Policy (#3224):
+//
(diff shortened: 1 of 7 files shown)


emrahbaysal:
Ran a high-recall self-review (2 finder agents). Two real findings, both fixed in 615928163:

1. **Efficiency + non-determinism (sovereign.go)**: the per-HR `userUIGatePasses` call in the `/v1/sovereign/apps` list (FE polls every 5s) issued one uncached catalog round-trip per HR under a single shared 10s ctx — budget exhaustion would make the last-iterated HRs (random Go map order) non-deterministically lose their Open button. Now memoized per blueprint → at most one round-trip per unique chart (shared-pg/-b/-c collapse to one); the user-UI chart set is tiny.
2. **Instance-card launch key (endpoint_handler.go)**: the grid instance card keys its launch on the Application CR *name* (the projected instance identity), not the uid AppDetail carries. `findApplicationByUID` now falls back to a name match after the uid match, so an Org-scoped instance resolves the real CR (correct org + blueprint) instead of the empty-org HR fallback that produced a login-form bypass. uid (UUID) and name (DNS label) never collide; uid is tried first.

One low-likelihood asymmetry (HR missing `spec.chart.spec.chart` — grid has an HR-name fallback the detail path lacks) was noted but not changed: Flux HRs near-always carry the chart field, and the grid's `bp-<slug>` fallback is the more-correct side.

New test `TestFindApplicationByUID_NameFallback`; full `internal/handler` package + go vet green. UI suite (AppsPage + open-button + AppDetail) 51 green, tsc clean.

PR #3631, opened 2026-06-15
fix: hw146 region-b cross-region data-wiring — shared-pg write host + openbao DR S3 cred + guacd PVC (Refs #3629)

emrahbaysal:
Fixes the three real cross-region data-wiring defects surfaced by the hw146 region-b app-health sweep (deployment `531e6908e606d11d`, 2-region kom4dc 2-VPC mimic). All verified live read-only; fixes are chart/IaC source only and ride the next fresh prov (zero-touch — hw146 was NOT hand-patched).

Refs #3629 #3571 #3572 #3375

## Defect 1 — cross-region consumer WRITE host (bp-postgres 0.2.2)

`grafana`/`powerdns-admin`/`keycloak` (all on the #3572 active-hot-standby consumer list, founder-confirmed agreed-multi-region) crashlooped in region-B on `shared-pg-*-rw.shared-data` NXDOMAIN. **Verdict: DB WIRING, not placement** — these apps ARE meant to run multi-region; the bug is the missing write-capable cross-region host.

0.2.0 promised "the consumer host `<instance>-rw` is unchanged across regions" but never delivered it: CNPG's `<instance>-rw` Service is region-LOCAL, and the only ClusterMesh-global Service (`<instance>-mesh`) is `selectorType: r` (read-only). Fix: a SECOND managed Service `<instance>-mesh-rw` (`selectorType: rw`, `service.cilium.io/global`) on the primary + a same-named replica stub, so the write host resolves in BOTH regions and routes writes to the current primary (follows bp-continuum promotion). The hub Secret `host`/`uri` + a new `reflect.hostKeys`/`hostPortKeys` mechanism (grafana `GF_DATABASE_HOST`) render the topology-aware host. keycloak slot 09 host → substitution var (the bp-gitea/bp-harbor pattern) so the cross-region overlay flips it to `shared-pg-mesh-rw`.

- `cluster.yaml` +`selectorType: rw` global Service; `replica-mesh-service.yaml` +`-mesh-rw` stub; `_helpers.tpl` +`writeServiceName`; `role-secrets.yaml` topology-aware `$consumerHost` + `hostKeys`/`hostPortKeys`; slots 16c (grafana) + 09 (keycloak).
- **singleton render byte-identical** (gated on `activeHotStandby`). `postgres-render.sh` +assertions; all 11 cases green.

## Defect 2 — openbao raft DR snapshot S3 creds (bp-openbao 1.2.44)

`openbao-snapshot-save` (region-a) + `openbao-snapshot-fetch` (region-b) Jobs `CreateContainerConfigError`. **Verdict: real defect** — the snapshot CronJobs read S3 creds from `seaweedfs-s3-secret` via secretKeyRef (resolves only in the Pod's own ns), but bp-seaweedfs creates that Secret INSIDE the rtz vCluster; the syncer mirrors it host-side as `seaweedfs-s3-secret-x-seaweedfs-x-rtz-vcluster` in ns `rtz`, never into ns openbao. The chart already documented the expectation that a bridge seeds it — but none was wired.

Fix: a seeder Job (+ SA + cross-ns read ClusterRole resourceNames-pinned + write Role) copies the admin access/secret keys from the host-synced source into ns openbao as `seaweedfs-s3-secret` — the same shape catalyst qa-fixtures' `qa-cnpg-backup-s3-seed` Job uses. Regular release resource (never a hook → no circular-dep wedge). Renders only under `snapshotReplication.enabled` (= `SOVEREIGN_ENABLE_HOT_STANDBY`, the SAME gate the snapshot Jobs ride). Embedded scripts pass `sh -n` + `dash -n`. `snapshot-replication-toggle.sh` +seeder assertions.

## Defect 3a — guacd PVC mount (bp-guacamole 0.2.19)

`guacd` Pending in BOTH regions: `persistentvolumeclaim "guacamole-recordings" not found`. **Verdict: real chart defect** — #3374 made the WEBAPP `recordings` volume an emptyDir when `recordings.persistence=false` (default) but MISSED `guacd-deployment.yaml`, which still hard-referenced the gated-off PVC. Mirror the webapp's emptyDir/PVC `if` into guacd.

## Defect 3b — cnpg-pair replica-3 Pending — NO FIX (expected)

`cnpg-pair-bp-cnpg-pair-replica-3-join` Pending: `1 node(s) didn't match pod anti-affinity rules, 3 Insufficient cpu`. **Verdict: kom4dc resource ceiling, NOT a chart defect.** 3 replica instances (`podAntiAffinityType: preferred`, soft) across only 3 region-b workers already at 37-41 pods each; replica-1/-2 Ready = the working HA floor, replica-3 best-effort. No anti-affinity/resource change warranted.

## Validation
- `helm lint` clean on all 3 charts; blueprint.yaml ↔ Chart.yaml version parity OK; `kubectl kustomize` of the bootstrap-kit builds.
- Chart render tests green (bp-postgres 11 cases incl. singleton byte-identical + new `-mesh-rw`; bp-openbao 4 cases incl. seeder; bp-guacamole render verified default=emptyDir / persistence=true=PVC for guacd+webapp).
- All version bumps lockstep (Chart.yaml + blueprint.yaml + bootstrap-kit slot pins: bp-postgres 16a/16c/16d, bp-openbao 08, bp-guacamole 52).

### DIFF
--- a/clusters/_template/bootstrap-kit/08-openbao.yaml
+++ b/clusters/_template/bootstrap-kit/08-openbao.yaml
@@ -104,12 +104,15 @@ spec:
       # cascading into sso-bridge/gitea/harbor/oidc-gate/newapi/powerdns-admin.
       # Render now gated on .Capabilities.APIVersions.Has "dr.openova.io/v1"
       # (gold-standard pattern). Single-region byte-identical; pin in lockstep.
+      # 1.2.45 (#3626 + #3629): reconcile two concurrent 1.2.44 bumps — the
+      # #3626 console "Open" → /sso/landing repoint AND the #3629 snapshot-
+      # replication S3 seeder (independent files; pure version-line union).
       # 1.2.44 (#3626): repoint the console "Open" button's launch-url at the
       # on-origin /sso/landing page (drop ssoShim + the /ui/vault/auth?with=oidc
       # deep-link) so OpenBao lands signed-in instead of the popup-only Vault
       # OIDC callback's window.opener=null error. blueprint.yaml/Chart.yaml
       # lockstep; no chart-template change (landing page already shipped 1.2.35).
-      version: 1.2.44
+      version: 1.2.45
       sourceRef:
         kind: HelmRepository
         name: bp-openbao
--- a/clusters/_template/bootstrap-kit/09-keycloak.yaml
+++ b/clusters/_template/bootstrap-kit/09-keycloak.yaml
@@ -270,7 +270,18 @@ spec:
       postgresql:
         enabled: ${SOVEREIGN_KEYCLOAK_PG_OWN_CLUSTER:=true}
       externalDatabase:
-        host: shared-pg-rw.shared-data.svc.cluster.local
+        # #3629: host is a substitution var (the SAME pattern bp-gitea slot 10
+        # `${SOVEREIGN_GITEA_PG_HOST:=…}` + bp-harbor slot 19
+        # `${SOVEREIGN_HARBOR_PG_HOST:=…}` already use) instead of a hardcoded
+        # literal. Default `shared-pg-rw.shared-data…` keeps single-region
+        # byte-identical, but bitnami's externalDatabase.host is a scalar (no
+        # secretKeyRef), so the cross-region overlay must flip it to the
+        # ClusterMesh-global WRITE alias `shared-pg-mesh-rw.shared-data…` —
+        # the region-local `-rw` Service NXDOMAINs keycloak in the replica
+        # region (hw146: keycloak-0 JGroups JDBC `UnknownHostException
+        # shared-pg-rw`). bp-postgres 0.2.2 publishes the `-mesh-rw` alias on
+        # the primary + a replica stub so the value resolves in BOTH regions.
+        host: ${SOVEREIGN_KEYCLOAK_PG_HOST:=shared-pg-rw.shared-data.svc.cluster.local}
         port: 5432
         user: keycloak
         database: keycloak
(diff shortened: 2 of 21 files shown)


emrahbaysal:
## Coverage note — Defect 1 host fix boundaries

**Fully fixed in this PR (no external dependency), consume the chart-computed topology-aware host directly:**
- **grafana** — `GF_DATABASE_HOST` now a `hostPortKeys` entry → bp-postgres renders `shared-pg-b-mesh-rw` under active-hot-standby.
- **powerdns-admin** — reads the hub Secret `uri`/`host` keys, now topology-aware in `role-secrets.yaml`.

These two were the primary region-b crashloops in the sweep and are end-to-end fixed by the chart change.

**keycloak / gitea / harbor** consume their shared-pg host via a `${SOVEREIGN_*_PG_HOST:=…}` substitution var (bitnami/gitea/harbor charts take a scalar host, not a secretKeyRef). This PR makes keycloak's host overlay-controllable for the FIRST time (slot 09 — previously a hardcoded `shared-pg-rw` literal), matching the existing gitea (slot 10) + harbor (slot 19) pattern. The cross-region flip to `shared-pg-mesh-rw` for these three is a **private-repo cloud-init overlay** concern: the 2-region provisioning path must set `SOVEREIGN_KEYCLOAK_PG_HOST` / `SOVEREIGN_GITEA_PG_HOST` / `SOVEREIGN_HARBOR_PG_HOST` = `shared-pg{,-b,...}-mesh-rw.shared-data.svc.cluster.local` on the secondary region (these vars are set in openova-private cloud-init, not visible here). gitea/harbor have the IDENTICAL latent issue today — this is a consistent, pre-existing boundary, not a regression. The public-repo half (the global `-mesh-rw` Service + the substitution seam) is complete; the private overlay is the one-line follow-up to wire all three across regions.

Single-region (the default + byte-identical render) is unaffected — the default resolves to the region-local `-rw` exactly as before.

PR #3725, opened 2026-06-17
ci: enforce cloud-init 32256B size cap as a required PR check (Refs #3720)

emrahbaysal:
## What

Adds a **dedicated, uniquely-named, required-able PR-level gate** so a cloud-init `user_data` size regression is caught **before merge**, not hours later on a frozen main.

Closes the gap behind the 2026-06-16 incident (issue #3724): PR #3715 pushed the rendered Hetzner control-plane cloud-init past the **32256-byte** guardrail and merged anyway → `catalyst-build` ("Build & Deploy Catalyst") then failed on **every** main commit → no catalyst-api image built for hours → the mothership silently froze 15 commits behind main and never received the critical #3717 fix. Reverted in #3720/#3721.

## Root cause (more specific than "G36 only runs on push")

A PR-level size gate **already existed**: `infra-hetzner-tofu.yaml` runs `tofu test` (which evaluates the 32256 precondition) on `pull_request`. On PR #3715 it **DID run and DID fail** (check `infra/hetzner — OpenTofu validate + test`, run `27646893533`, step `tofu test`, conclusion=failure). The PR merged anyway because:

1. **Not a required/blocking check** — the red did not block merge.
2. **Ambiguous duplicate name** — `infra-hetzner-tofu.yaml` and `infra-huawei-tofu.yaml` both expose the job as `validate + fmt + test`, so the PR check list showed `validate + fmt + test  fail` next to `validate + fmt + test  pass` — easy to miss, hard to mark required.

Plus: **huawei had no size precondition at all** (renders the same `_shared` template; see `infra/providers/huawei/main.tf:625` — "we keep the same conservative budget" was aspirational, never enforced).

## Approach — new dedicated workflow (NOT extending catalyst-build)

I did **not** add `pull_request` to `catalyst-build.yaml` — that would be a **third** copy of a gate that already runs at PR level in `infra-*-tofu.yaml`. Instead:

| File | Role |
|---|---|
| `.github/workflows/cloud-init-size.yaml` | **New** PR + push gate. Per-provider matrix with **unique check names** (`cloud-init size gate / hetzner`, `cloud-init size gate / huawei`) that can be marked **required** in branch protection and won't collide. Path-scoped to `infra/providers/**` + `products/catalyst/bootstrap/**`. |
| `infra/providers/hetzner/tests/cloudinit_size.tftest.hcl` | **Blocking** size assertions on the **real** rendered cloud-init (real module under `mock_provider` — zero render-harness drift): CP ≤ 32256, worker ≤ 30720, each secondary region ≤ 32256. |
| `infra/providers/huawei/tests/cloudinit_size.tftest.hcl` | Same, for huawei — **adds size coverage that never existed** (CP-per-region ≤ 32256, worker-per-region ≤ 30720). |
| `scripts/cloudinit-size-report.sh` | **Headroom signal**: reports the exact rendered byte count, annotates `::warning::` once it crosses ~93% of the cap (30000B), `::error::` if already over, `::notice::` otherwise. Renders via the real module (no drift). |
| `.github/workflows/catalyst-build.yaml` | `if: failure()` loud-signal step on `build-api` so a red main build (no image published → mothership freezes) is **unmissable**. |

The committed size tftests are also picked up by the existing `infra-*-tofu.yaml` `tofu test` (free redundancy) and keep those suites green.

## Why the byte count has zero drift

The only faithful size source is the real module's own `local.*_cloud_init` — it threads multi-KB provider-injected strings (registry mirror, cloud-credentials secret, crossplane provider yaml, kubeconfig PUT block) that a standalone `templatefile()` harness would omit and undercount by KBs. (Proof: hetzner renders at **31539B** but huawei at **22688B** for the same shared template — the per-provider injected strings differ by ~9KB.) The report reads the real local via a `tofu test` probe; `nonsensitive(length(...))` leaks only the integer, never any token bytes (same pattern as the production precondition at `main.tf:956`).

## Proof — trips on oversized, passes on current main

Verified locally on OpenTofu 1.12 (CI pins 1.8.5, matching the `infra-*-tofu.yaml` siblings; full suites green on both):

**PASS on current main:**
```
hetzner: tofu test -filter=tests/cloudinit_size.tftest.hcl → 3 passed, 0 failed
huawei:  tofu test -filter=tests/cloudinit_size.tftest.hcl → 2 passed, 0 failed
full hetzner suite: 13 passed, 0 failed   full huawei suite: 7 passed, 0 failed
report (hetzner): rendered control-plane cloud-init = 31539 bytes (97% of 32256; 717 bytes headroom)
                  ::warning title=cloud-init headroom low (hetzner)::...   ← main is ALREADY at the cliff
report (huawei):  rendered control-plane cloud-init = 22688 bytes (70%; 9568 bytes headroom) → ::notice::
```

**FAIL when the shared tftpl is padded over the cap (then reverted):**
```
control_plane_cloud_init_under_32256... fail   →  Failure! 0 passed, 1 failed, 2 skipped
report: rendered control-plane cloud-init = 34220 bytes (106%; -1964 headroom)
        ::error title=cloud-init OVER cap (hetzner)::...
```

`fmt -check` clean, `shellcheck -S warning` clean, YAML valid.

## Follow-up (not in this PR)

Mark `cloud-init size gate / hetzner` and `cloud-init size gate / huawei` as **required status checks** on `main` branch protection — that is the human step that actually makes the red block a merge (the workflow is built to be required-able; this PR cannot edit branch protection).

## Notes

- **DO NOT merge yet** — parked for review.
- `Refs #3720` / `Refs #3724` (no auto-close).

Refs #3720
Refs #3724

### DIFF
--- a/.github/workflows/catalyst-build.yaml
+++ b/.github/workflows/catalyst-build.yaml
@@ -382,6 +382,32 @@ jobs:
           done
           echo "G36 gate OK: every templatefile() in huawei parses + renders."
 
+      # Loud-signal on a main-push build-api FAILURE (issue #3724 / Refs #3720).
+      #
+      # When #3715's cloud-init bloat merged, this job's G36 step failed on EVERY
+      # subsequent main commit — but nothing SHOUTED, so no catalyst-api image
+      # built for hours and the mothership silently froze 15 commits behind main
+      # (never receiving the #3717 fix). The regression should now be caught at
+      # PR time by `cloud-init-size.yaml` (a required check) + infra-*-tofu.yaml,
+      # but if anything still reaches a red main build, make it UNMISSABLE: an
+      # ::error:: annotation + a Step Summary block naming the broken job and the
+      # consequence (no image published → mothership freezes). `if: failure()`
+      # fires only when an earlier step in this job failed.
+      - name: Announce build-api failure (no catalyst-api image will publish)
+        if: failure()
+        run: |
+          MSG="Build & Deploy Catalyst: build-api FAILED on ${GITHUB_REF_NAME} @ ${GITHUB_SHA}. No catalyst-api image was published for this commit — the mothership will NOT roll forward and silently stays on the last green SHA until this is fixed. Most common cause: the rendered cloud-init exceeded the 32256-byte Hetzner user_data guardrail (see the G36 tofu test step + issue #3724). Inspect the failed step above and revert/trim before the mothership drifts further behind main."
+          echo "::error title=catalyst-api build FAILED — mothership will freeze::${MSG}"
+          {
+            echo "## :rotating_light: Build & Deploy Catalyst — build-api FAILED"
+            echo ""
+            echo "${MSG}"
+            echo ""
+            echo "- Ref: \`${GITHUB_REF_NAME}\`"
+            echo "- Commit: \`${GITHUB_SHA}\`"
+            echo "- Run: ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          } >> "${GITHUB_STEP_SUMMARY}"
+
       - name: Push API image
         uses: docker/build-push-action@v6
         with:
--- a/.github/workflows/cloud-init-size.yaml
+++ b/.github/workflows/cloud-init-size.yaml
@@ -1,0 +1,104 @@
+name: cloud-init size gate
+
+# PR-LEVEL guardrail against the cloud-init `user_data` size regression that
+# froze "Build & Deploy Catalyst" for hours on 2026-06-16 (issue #3724 / Refs
+# #3720). Background:
+#
+# PR #3715 added +124 lines to infra/providers/_shared/cloudinit-control-
+# plane.tftpl, pushing the *rendered* Hetzner control-plane cloud-init past the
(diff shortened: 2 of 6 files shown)


emrahbaysal:
Closed as fully superseded — every component adjudicated against current main file-by-file (2026-08-03):

1. **The required PR check** — main has `.github/workflows/cloud-init-size.yaml` (127 lines, self-contained: zero references to this branch's `cloudinit-size-report.sh` or tftest files), and it is LIVE and green — 3/3 successes on today's PR runs alone.
2. **The loud-failure announcer** — main's `catalyst-build.yaml:444-465` carries an evolved version of this branch's hunk: same `if: failure()` + `::error::` + step-summary shape, same #3724 silent-freeze rationale, and it additionally cross-references the cloud-init-size required gate.
3. **The tftest pair + report script** — implementation details of this branch's variant of the gate; main's implementation took a different route and nothing on main references them.

The defect class this PR existed to prevent (an oversized cloud-init silently freezing image publishing and stranding the mothership behind main) is covered at both layers on main — the pre-merge gate and the post-failure alarm. Nothing here is lost.

One irony recorded for the reader: the mothership is currently behind main anyway — not from a silent red build, but from the deliberate scale-to-zero (#5573/#5558). The alarm this PR wanted exists; the freeze it warns about is operator-gated this time.

Refs #3720 #3724 #5573 #960

Issue #3376, opened 2026-06-12
FUNNEL: a stranger with a voucher ends signed-in in their own org console with the purchased app RUNNING (founder-approved)


emrahbaysal:
# 1. End goal — the finish line

A stranger holding a voucher ends **signed in inside their own organization's console with the purchased application RUNNING.** The terminal evidence is one screenshot: the WordPress (or whichever purchased app) serving inside `…<orgslug>.omani.homes`, reached zero-click as the org owner. Anything short of that — org record created, provisioning started, handover minted — is NOT done. (Founder: "think deeply and answer yourself" — this is the only finish line consistent with walk-proven DoD; PR merge ≠ shipped, org-row-created ≠ shipped.)

# 2. The law

1. Every step is IaC + API — zero manual touch anywhere in the chain.
2. The walk is the acceptance: voucher mint (BSS) → marketplace wizard → redeem → PIN checkout → tenant Org provisions → org console loads → owner signed in zero-click (per-Org realm, SSO ticket's contract) → purchased app serving.
3. Voucher hardening is part of this surface (folded from #2941, decided autonomously per founder): generated-code entropy + authenticated-redeem rate-limit.

# 3. Measured current state (hw130, 2026-06-12 — walked live, #3319)

- **Front half PASSES end-to-end with real artifacts**: voucher `WALKMART2026` (10 OMR) → wizard (plan S, WordPress, `walkmart.omani.homes`) → redeem valid → real Stalwart-PIN checkout → `POST /api/tenant/orgs` **201** → `POST /api/billing/checkout` **200**, voucher exhausted 1/1, reflected operator-side.
- **It breaks at exactly ONE unbuilt wire**: the mother→child **tofu-archive push** (#3052/#933 follow-up — filed, never built). Handover redemption works (operator sessions established); but nothing pushes the phase0 archive afterward: `secret/catalyst/tofu-phase0-archive` never appears → the cutover trigger stays 425-gated → the cutover-token annotation never lands → `sme/provisioning`'s init waits forever.
- Child side is ALREADY merged: `ReceiveTofuArchive` async-fires the cutover engine (#3127, `CATALYST_FIRE_CUTOVER_ON_HANDOVER` default-true).
[...]

Issue #71, opened 2026-04-28
[D] wizard: create core/console/src/pages/sovereign/index.astro entry route


emrahbaysal:
Parent: #43 — Catalyst-Zero consolidation + first Hetzner Sovereign at omantel.omani.works waterfall.

Group D.

Astro page hosting the wizard. Wired to Keycloak auth.

Reference: [`docs/PROVISIONING-PLAN.md`](docs/PROVISIONING-PLAN.md)

emrahbaysal:
Closed: per the disclosed divergence in commit e87913a, the React SPA wizard at products/catalyst/bootstrap/ui/ remains the wizard surface (Feature-Sliced Design). The originally-planned core/console/src/pages/sovereign/index.astro entry was not built because the Astro+Svelte port was traded for keeping the working FSD React app intact. The wizard is reachable at the catalyst-ui ingress per the existing deployment manifests. Closing per orchestrator decision; the merge-into-Astro option is parked, not required for DoD.

PR #3745, opened 2026-06-17
fix(provisioning): dedup provision-create + recognize Gitea ref-lock 500 (stranger redeem self-race)

emrahbaysal:
## What

A credit-covered marketplace redeem (the stranger-voucher funnel) fires the provision-create entrypoint **twice** — NATS `order.placed` event **and** `POST /provisioning/start` (the checkout client does both deliberately, comment: *"broker event is best-effort"*). `startProvisioning` unconditionally inserted a provision + forked a workflow, so both calls produced two near-simultaneous Gitea commits to the shared `sme-tenants` branch → a ref-lock race (`HTTP 500 PushRejected ... cannot lock ref ... failed to update ref`). The loser marked the **tenant `failed`** even though the Organization CR + the winning commit both landed.

**Seen live on hw158** during the TC-01 funnel walk: stranger `walkstranger@omani.homes` redeemed voucher `VCH-56T66GFC2BNC`, minting Organization `walk-stranger-co` — but provisions `2a352462` (OK) / `8bceaadf` (failed: `cannot lock ref 'refs/heads/sme-tenants'`) raced.

## Fix (mirrors the proven #71 day-2 Job dedup)

**1. PRIMARY — idempotency guard on the create path:**
- `store.CreateProvisionIfAbsent` + `EnsureProvisionIndexes` — a **partial** unique index on `tenant_id` filtered to in-flight statuses `{pending,provisioning,running}` (so a tenant can still be re-provisioned after a prior provision terminates) + `GetInFlightProvisionByTenant` + `ErrProvisionExists` sentinel.
- `startProvisioning` now uses `CreateProvisionIfAbsent`; on `ErrProvisionExists` it returns the already-running provision **without forking a second workflow** (mirrors `runInstallJob`). Both triggers share the entrypoint, so both inherit the guard. `Start` still returns 201 with the existing provision, which the client already tolerates.
- `main.go` calls `EnsureProvisionIndexes` at startup next to `EnsureJobIndexes`.

**2. SECONDARY — widen `isGiteaRefRaceError`** to match Gitea's real wording (`cannot lock ref`, `failed to update ref`, `PushRejected`, `but expected`), so the existing 5-attempt rebuild-against-fresh-HEAD loop self-heals a genuine *cross-tenant* concurrent commit instead of classifying it fatal on attempt 1 (it previously only matched `409` / `not a fast forward`, while Gitea returns **500** here).

## Tests
`client_refrace_test.go` locks the matcher against the verbatim hw158 500 body + a regression matrix (pre-existing signals still match; 401/404/422-malformed stay fatal). `go build` + `go vet` + `go test ./...` all green for the provisioning module.

## Acceptance
A credit-covered stranger redeem produces exactly ONE provision per tenant; the tenant is never marked `failed` due to a self-race. To verify on a fresh prov: redeem a voucher end-to-end, then `kubectl get organizations` non-zero + exactly one provision row for the tenant + tenant not `failed`.

Refs #3744
Refs #3376

### DIFF
--- a/core/services/provisioning/github/client.go
+++ b/core/services/provisioning/github/client.go
@@ -499,7 +499,13 @@ func (c *Client) ensureBranchExists(ctx context.Context, branch string) error {
 // "not a fast forward" signal the GitHub Git Data path uses, so the outer
 // retry loop in CommitFilesWithPruneAndRebuild treats both equivalently.
 // Gitea returns 409 Conflict or 422 with a body containing phrases like
-// "branch has been changed" / "stale base" / "ref has been updated".
+// "branch has been changed" / "stale base" / "ref has been updated". It can
+// ALSO surface a concurrent-commit collision as an HTTP 500 whose body carries
+// the raw git push rejection — "PushRejected ... cannot lock ref
+// 'refs/heads/<branch>': is at X but expected Y ... failed to update ref"
+// (observed on hw158, #3744). That wording is matched here too so the outer
+// rebuild loop re-fetches the branch head and retries against fresh HEAD
+// instead of classifying the loser of a genuine concurrent commit as fatal.
 func isGiteaRefRaceError(err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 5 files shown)


emrahbaysal:
Closed as superseded per the PATH TO 100 plan Phase A adjudication (2026-08-03), verified file-level against current main — not by branch age.

Specifically: the Gitea ref-race handling shipped via #5235: main's core/services/provisioning/github/client.go carries a full bounded-backoff retry (lines 90-171, same-files-map recommit against new HEAD) — richer than this branch's 500-recognition.

If any hunk here is still believed live, file an issue against the current implementation rather than rebasing a June branch.

PR #3836, opened 2026-06-18
fix(bootstrap): pin ghcr blob CDN pkg-containers.githubusercontent.com to node /etc/hosts — bootstrap-flux ImagePullBackOff 0-HR wedge on IPv4-only kom4dc

emrahbaysal:
Refs #3835 #3735 #3726 #3232

## Root cause (verified LIVE on hw166, dep `51661a89e1451114`, 2026-06-19)

Fresh kom4dc 2-region provs intermittently wedge at `phase1-watching` / 0 HelmReleases because the **bootstrap-phase Flux controllers** hit `ImagePullBackOff`. Cloud-init installs Flux via `flux install` from `ghcr.io/fluxcd/*` (~line 1066) BEFORE harbor exists — so it cannot use the #3735 harbor pull-through proxy (chicken-and-egg) and pulls direct from ghcr. Captured error:

```
Failed to pull image "ghcr.io/fluxcd/kustomize-controller:v1.4.0": ... dial tcp: lookup pkg-containers.githubusercontent.com: Try again
```

If the bootstrap kustomize-controller never starts it never reconciles the `bp-flux` HelmRelease that re-points the controllers at `harbor.openova.io/proxy-ghcr/fluxcd/*` — so the prov wedges at 0 HRs.

### The live forensic (hw166 node `...-me-east-215-a-w8ff6ea`)

ghcr.io serves image manifests but redirects the image **blobs** to the CDN host `pkg-containers.githubusercontent.com`. `getent ahosts` on the node returns it **dual-stack**:

| record | value |
|---|---|
| A | 185.199.108.154 / .109 / .110 / .111.154 |
| AAAA | 2606:50c0:8000::154 … 8003::154 |

Node `/etc/resolv.conf` → `127.0.0.53` (systemd-resolved) → upstreams `8.8.8.8` + `114.114.114.114`; both hand back A **and** AAAA. On the **IPv4-only Huawei VPC** the AAAA is unroutable, so containerd/Go's dialer intermittently tries IPv6 first → `Try again` → ImagePullBackOff. Same dual-stack IPv6 trap as `helm.cilium.io` (#3232) and `harbor.openova.io` (#3735).

The existing `r4()` node `/etc/hosts` pin loop in `infra/providers/_shared/cloudinit-control-plane.tftpl` (~line 811) already pins `github.com raw./codeload./objects.githubusercontent.com helm.cilium.io get.helm.sh` (+ Huawei: `api.github.com harbor.openova.io registry.<fqdn>`). Those github hosts are **IPv4-only** (no AAAA — verified live) so they never trip. `pkg-containers.githubusercontent.com` — the one dual-stack host, and the host **every ghcr image pull depends on** (the 6 flux controllers + the line-962 `crictl` pre-pull) — was never in the list. This PR closes that gap.

## Fix

Add `pkg-containers.githubusercontent.com` to the **Huawei-gated** segment of the existing `r4()` loop:

```diff
-    for h in github.com raw... get.helm.sh%{ if provider == "huawei" } api.github.com harbor.openova.io registry.${sovereign_fqdn}%{ endif ~}; do
+    for h in github.com raw... get.helm.sh%{ if provider == "huawei" } api.github.com pkg-containers.githubusercontent.com harbor.openova.io registry.${sovereign_fqdn}%{ endif ~}; do
```

### Why this is robust (not fragile CDN pinning)
- `r4()` resolves the A record **at boot** via the bounded 3-tier ladder (getent → public-resolver `nslookup -type=A` → python `AF_INET getaddrinfo`) — no hardcoded IP. If Fastly/Azure rotates the anycast edge it self-heals next boot. The `[ -n "$a" ]` guard skips the pin entirely if resolution fails (no bad pin).
- Pinning ONE anycast A record removes the AAAA from the node's view so containerd never attempts IPv6 — the 4 A records are anycast and serve identically.
- containerd reads the **node** `/etc/hosts`, not CoreDNS (per the line-752 comment) — so the node pin is the correct + sufficient layer. **No CoreDNS change needed** (the in-cluster coredns-custom step's `GH_HOSTS` list does not name this host, so the side-file extra entry is harmless).

### Why Huawei-gated
Hetzner is dual-stack — its nodes reach the AAAA, so the pin is unnecessary — and its CP cloud-init render is byte-tight under the 32256 B cap (#966/#1981). Gating to `%{ if provider == "huawei" }` keeps the Hetzner render **byte-identical** (same pattern as the already-gated `harbor.openova.io`/`registry.<fqdn>` pins). This avoids the bloat that reverted the #3715 attempt (#3721).

## Verification (measured render, both providers)

| provider | capped render (post comment-strip) | ≤ 32256 cap | `pkg-containers` pinned |
|---|---|---|---|
| hetzner | **23,947 B** (8,309 B headroom) | yes | no — gated out, render byte-identical |
| huawei | **25,429 B** (6,827 B headroom) | yes | yes — in node `/etc/hosts` loop |

- `scripts/lint-cloudinit.sh hetzner` + `huawei`: render OK + `dash -n` PASS for every runcmd item, both providers.
- `tofu test` (infra/providers/hetzner): **10 passed, 0 failed** — the `length(control_plane_cloud_init) <= 32256` precondition holds across primary + multi-region runs.

## Distinct from existing tickets
- **#3726** (status/uat): github-IPv4 NODE pin determinism for the **git/helm bootstrap-artifact** hosts (k3s/helm/git). Does not include the ghcr image-blob CDN host.
- **#3735** (open): `harbor.openova.io` node pin + kyverno un-proxied images + cnpg cold-pull. Re-points flux to the harbor proxy POST-bootstrap; does not address the direct-ghcr BOOTSTRAP flux pull.

### DIFF
--- a/docs/ledger/UAT.md
+++ b/docs/ledger/UAT.md
@@ -1,4 +1,4 @@
-# UAT — browser walkthrough dashboard · `hw159` (2026-06-17) — fresh-prov walk (RESET 2026-06-18 — pending hw162)
+# UAT — browser walkthrough dashboard · `hw159` (2026-06-17) — fresh-prov walk (RESET 2026-06-19 — pending hw166)
 
 > **Env:** `hw159.omani.works` · deployment `c117f6fd4e2eb2dd` · single physical kom4dc region
 > (2 VPCs `me-east-215-a` / `-b`). On each wipe + re-prov this dashboard resets and the links flip
--- a/docs/ledger/uat-walkthrough/_RESULT-hw165-2026-06-18.md
+++ b/docs/ledger/uat-walkthrough/_RESULT-hw165-2026-06-18.md
@@ -8,7 +8,7 @@ Walk = 9 parallel browser-walker agents (screenshots only). Evidence under
 
 ## Aggregate (165 browser-walked rows): **57 ✅ · 44 ❌ · 28 GAP · 36 ☐**
 
-| Runbook (ticket) | ✅ | ❌ | GAP | ☐ | Headline |
+| Runbook (ticket) | ☐ | ❌ | GAP | ☐ | Headline |
 |---|:--:|:--:|:--:|:--:|---|
 | organizations #3383 | 7 | 0 | 0 | 0 | FULL PASS — persona-rename landed, create-org FIXED, zero sme/tenant |
 | catalog #3668 | 6 | 0 | 0 | 1 | FULL PASS — 64-card grid, single-source IaC YamlEditor, icon picker |
--- a/infra/providers/_shared/cloudinit-control-plane.tftpl
+++ b/infra/providers/_shared/cloudinit-control-plane.tftpl
@@ -761,6 +761,25 @@ runcmd:
   # giving every IPv4-only cloud the harbor pin. No hardcoded IP: r4() derives
   # the A record at boot and self-heals if harbor's address ever changes.
   #
+  # #3835 (2026-06-19 — the hw166 bootstrap-flux ImagePullBackOff): the
+  # `flux install` below (~line 1066, applied from ghcr.io/fluxcd BEFORE harbor
+  # exists — the chicken-and-egg the #3735 harbor-proxy can't break at bootstrap)
+  # pulls the 6 controller IMAGES; ghcr.io serves their manifests but redirects
+  # the BLOBS to the CDN host `pkg-containers.githubusercontent.com`, which is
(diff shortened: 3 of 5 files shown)


emrahbaysal:
Superseded by #3837 — clean one-concern PR off main. This branch picked up unrelated TopologyTab + UAT.md from a stale base branch (shared-checkout hazard); the ghcr-DNS fix content is unchanged + verified (render under 32256B cap, tofu 10/10).

PR #3871, opened 2026-06-19
fix(catalog): seed production bp-wordpress alias + bump bp-wordpress-tenant pin 0.2.0->0.4.1

emrahbaysal:
## What

bp-wordpress 404'd in the operator catalog on the live hw167 walk (rows 3668-27/-28/-29/-38: `Couldn't load wordpress / catalog get: HTTP 404`). WordPress is the app the #3376 funnel sells, so its catalog blueprint must resolve on a fresh **production** prov.

## Root cause (verified static + live)

The wordpress blueprint is **not missing entirely** — `platform/wordpress-tenant/` ships a full chart published to `ghcr.io/openova-io/bp-wordpress-tenant` through **0.4.1**. It was **mis-seeded for production**:

1. The marketplace funnel (`core/marketplace AppsStep.svelte` slug `wordpress`), operator docs, and the qa-loop matrix resolve WordPress under the **bare** slug `wordpress`/`bp-wordpress`. `CatalogDetail.tsx:58` strips the `bp-` prefix; `catalog_client_cluster_fallback.go:224` retries bare→`bp-` and looks up a CR named **`bp-wordpress`**.
2. That `bp-wordpress` CR was seeded **only** by `templates/qa-fixtures/blueprint-bp-wordpress.yaml`, gated on `qaFixtures.enabled` (`values.yaml` default **false** on production). Production `catalogSeed.enabled` seeded only `bp-wordpress-tenant`.
3. The `bp-wordpress-tenant` seed entry's `source.version` was stale at **0.2.0** vs the published **0.4.1** — even the canonical card fetched an old chart.

Net: `GET /api/v1/catalog/wordpress` 404'd on a fresh production prov.

## Fix

- **Add** an always-on (`catalogSeed.enabled`) `bp-wordpress` alias Blueprint CR in `templates/catalog-seed/blueprints.yaml` pointing at the real `bp-wordpress-tenant` chart bytes (no new chart) — the bare slug now resolves on every Sovereign.
- **Bump** the `bp-wordpress-tenant` seed `spec.version` + `source.version` 0.2.0 → 0.4.1 (lockstep with `platform/wordpress-tenant/blueprint.yaml` + the published OCI chart).
- **Bump** `bp-catalyst-platform` chart 1.4.688 → 1.4.689 + bootstrap-kit slot-13 pin lockstep.

## Verification

- `helm template … --show-only templates/catalog-seed/blueprints.yaml` renders both `bp-wordpress-tenant` + `bp-wordpress` cleanly; `hostnameTemplate`/`realm` escaping resolves to literal `{{.AppName}}…`; both version fields render `0.4.1`.
- `chart/tests/baseline-cnp-allowlist.sh` — all 13 gates green.
- `catalog-seed`-only change; Catalyst-Zero (empty FQDN) render byte-unchanged.

Refs #3668 #3376


emrahbaysal:
Verified against live hw167 (console/proxy path, real Keycloak `sovereign`-realm token, from inside the hw167 catalyst-api pod): `clusterFallback=true`; the proxy resolves `bp-wordpress-tenant`→200 and `bp-grafana`→200 via the in-cluster CR fallback, but `wordpress`/`bp-wordpress`→404 because no `bp-wordpress` CR exists on production (only `bp-wordpress-tenant`; `bp-wordpress` is seeded only by default-OFF qa-fixtures). This PR seeds the always-on `bp-wordpress` alias CR → the bare-name path resolves through the same healthy fallback. The list endpoint already returns non-empty items[] from in-cluster `managed-by: catalog-seed` CRs (catalyst-catalog's own Gitea source is empty on hw167 but the proxy's CR fallback masks that — separate #3668-master concern, not regressed here).

PR #3907, opened 2026-06-19
fix(catalyst-console): topology placement-edit dropdown derives mode descriptions from the per-component matrix (openbao active-passive no longer says 'backup-restore') (Refs #3905)

emrahbaysal:
## What

The placement-edit **mode dropdown** on an Application's Topology tab now derives its per-mode helper text from the **same per-component topology matrix** (`TOPOLOGY_BY_ID`) that the declared-topology **CARD** reads. The card and the edit UI can no longer contradict for any component.

## The bug (Refs #3905)

`TopologyEditor.describeMode()` returned **generic, hardcoded, component-agnostic** strings. For `active-passive` it always read *"primary serves; passive cold/warm standby (**backup-restore**)"*. The card derives each component's real DR contract from the matrix (generated from `blueprint.yaml` `spec.topology`), so for **bp-openbao** the card showed `raft · async` / `raft-transition` / RTO 60s · RPO 30s while the dropdown said "backup-restore" — the contradiction the founder flagged across many envs.

## The fix

- New exported `describeModeForComponent(mode, topology)` in `TopologyEditor.tsx` derives the dropdown copy from the same `BlueprintTopology.perTopology[variant]` contract the card reads (replication backend/mode + switchover mechanism + RTO/RPO).
- Generic `describeMode()` stays only as the fallback (no matrix entry, or the `singleton` class which carries no DR contract).
- Wired into `TopologyTab` (post-create placement editor, via a new optional `topology` prop) and `InstancesSection` (instance-create mode picker), both passing the component's `TOPOLOGY_BY_ID` entry.

For openbao active-passive the dropdown now shows **`raft · async replication; raft-transition switchover (RTO 60s / RPO 30s)`** — matching the card, never "backup-restore".

## Secondary-line verdict

The card showing **"active-passive (DEGRADED — mandate unbuilt, no live DR pair)"** while the edit UI offers active-passive as **selectable** is **by design**, not an inconsistency:
- The card's DEGRADED is the **observed/effective live state** — no live Continuum (dr.openova.io) pair backs the app on this Sovereign, so it currently runs as a degraded singleton (see `TopologyTab.tsx` "Effective class" logic, #3829).
- The edit dropdown offers the **declared/selectable mandate**; selecting active-passive is what provisions the pair.

The two describe different axes (realized state vs. selectable mandate) and are consistent. This PR does not change that behavior.

## Validation

- `npx vitest run src/widgets/topology/TopologyEditor.test.tsx` → 10/10 pass, including 4 new tests asserting the openbao active-passive dropdown derives raft/async/raft-transition/60s/30s from the matrix and contains **no** "backup-restore".
- Broader suite (`topology/` + `InstancesSection` + `TopologyTab.dr` + `CatalogDetail.edit`) → 38/38 pass.
- `npm run build` (tsc -b typecheck + vite production build) → green.

Env-independent FE/data fix. No prov fired. `Refs #3905` (not `Closes`).

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/deployments.go
+++ b/products/catalyst/bootstrap/api/internal/handler/deployments.go
@@ -1814,6 +1814,10 @@ func (h *Handler) runProvisioning(dep *Deployment) {
 		}
 		bridge := dep.jobsBridge
 		dep.mu.Unlock()
+		// Stamp the deployment's cloud provider so the Phase-0 lifecycle
+		// parent group renders "Provision <Provider>" (e.g. "Provision
+		// Huawei") instead of a hardcoded Hetzner label (#3895).
+		bridge.SetProvider(firstProvider(dep.Request))
 		if seedErr := bridge.SeedProvisionerJobs(); seedErr != nil {
 			h.log.Warn("jobs bridge: seed Phase-0 lifecycle jobs failed",
 				"id", dep.ID, "err", seedErr)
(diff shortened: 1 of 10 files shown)


emrahbaysal:
Superseded by #3940 — re-cut clean off current origin/main. #3907 had drifted into a 10-file conflict carrying unrelated stale-stack bootstrap/api/internal/... Go drift. #3940 isolates ONLY the real describeMode-from-matrix change (4 FE files), with the same 4 new tests (openbao active-passive renders raft·async / raft-transition / 60s/30s, never 'backup-restore'; editor + card agree). Refs #3905.

PR #3918, opened 2026-06-19
fix(catalyst-api): jobs page no longer conflates Flux continuous-reconcile with one-shot terminal jobs (Refs #3916)

emrahbaysal:
## Problem

On `/sovereign/provision/<id>/jobs` the operator saw a confusing, contradictory picture even though the prov converged healthily:

- **(A) Status flapping** — a job oscillated `Failed → Success → Failed → Success`. Failed/Success are TERMINAL states; a thing that keeps changing them is actually IN-PROGRESS (a Flux HR/Kustomization still reconciling/retrying).
- **(B) Count accumulation + region asymmetry** — region A ~128 jobs (climbing 141→145), region B ~64, while the actual k8s Jobs were only 18-24/region. The console "jobs" are the catalyst-api job MODEL, and a NEW entry was minted per reconcile-event instead of updating ONE stable entry per reconciler identity.

## Root cause (exact)

**(A) flapping**
- `helmwatch/reconcilers.go:statusFromReadyCondition` mapped a Flux Kustomization's transient `Ready=False` (still reconciling: BuildFailed/HealthCheckFailed/Progressing) to TERMINAL `failed`.
- The HR snapshot projected a still-retrying HelmRelease (`InstallFailed`/`UpgradeFailed` while Flux `remediation.retries` are NOT yet exhausted) as terminal `failed`.
- The chroot `/jobs` seed re-reads live state on every poll and `store.mergeJob` lets the new status win → the leaf flapped `failed`↔`succeeded` as the Ready condition oscillated.

**(B) accumulation**
- `helmwatch/reconcilers.go` minted `task-<u.GetName()>` leaves keyed by the **per-run unique Job name** (controllers append `-<hash>`/`-<unix-stamp>`). Every Flux-recreated/retried Job → a brand-new leaf → unbounded growth, worse in the busier region.

## Fix — one stable entry per reconciler identity; terminal only when genuinely terminal

- **(A1)** `statusFromReadyCondition`: only Flux's terminal `Stalled` reason is a terminal failure; every other `Ready=False` is `running` (in-progress).
- **(A2)** New `helmwatch.IsStalledHelmRelease` + `ComponentSnapshot.Stalled` (additive, `omitempty`). The READ-side seed (`snapshotsToSeedsForRegion`) downgrades a transient `failed`-but-not-stalled HR to `installing`. **`DeriveState` (live-watch outcome classification + late-poll #910) is intentionally LEFT UNCHANGED** — changing it broke the watch's recovery machinery (verified: 9 watch tests went red), so the fix lives at the read/projection boundary only.
- **(B)** `taskBaseName` derives the stable base identity (`metadata.generateName`, else strip a single controller-appended run suffix) and keys the `task-<base>` leaf on it; each per-run instance becomes an Execution under that single leaf (same recency rule as the cron leaf). N reruns = 1 row + N runs, identical structure across regions.

## Result

A still-reconciling HR/Kustomization renders **In-Progress (Running, spinner)** instead of a flapping terminal. The count is **stable** across polls (one entry per reconciler identity). Region A and B have **identical structure** because the leaf key no longer carries run-instance entropy. A genuinely stalled HR/Kustomization keeps its terminal Failed.

## Tests

- `reconcilers_test.go`: transient `Ready=False` → running (not flapping failed); `Stalled` → terminal failed; N generateName/run-suffix Jobs collapse to one stable `task-<base>` leaf with N Executions; fixed-name Job's meaningful trailing word preserved.
- `reconciler_bridge_test.go`: 20 repeated reconcile polls keep the leaf count flat; a still-reconciling leaf is `running` (never terminal).
- `jobs_backfill_test.go`: transient-failed-not-stalled HR seeds in-progress; stalled HR seeds terminal failed.

## Validation

- `go build ./...` clean
- helmwatch + jobs + handler suites all green (handler full, no cache, 112s)
- UI `npm run typecheck` exit 0; UI `vitest jobsAdapter` 7/7 (no UI change needed — the existing JobsTable already renders `running` with the in-progress spinner)

Refs #3916

### DIFF
--- a/.github/workflows/pages-design-mock.yml
+++ b/.github/workflows/pages-design-mock.yml
@@ -1,0 +1,35 @@
+# Deploys the #3905 topology-tab design-review mock (and anything under docs/)
+# to GitHub Pages. Scoped to the design-mock branch so it never interferes with
+# main. The mock is static (no build step) — we upload docs/ verbatim.
+name: pages-design-mock
+
+on:
+  push:
+    branches:
+      - docs/3905-topology-tab-mock
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/configure-pages@v5
+      - name: Upload docs/ as the Pages artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: docs
+      - id: deployment
+        uses: actions/deploy-pages@v4
--- a/docs/design/topology-tab-mock/index.html
+++ b/docs/design/topology-tab-mock/index.html
@@ -1,0 +1,763 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1" />
+<title>Catalyst — Topology Tab (final UX design review)</title>
+<!--
+  DESIGN-REVIEW ARTIFACT — issue #3905.
+  Self-contained, no build step. All matrix values below are VERBATIM from the
+  source of truth: docs/topology-matrix.md (promoted into
+  products/catalyst/bootstrap/ui/src/shared/constants/catalog.generated.ts).
+  This mock does NOT touch the real console code — it is the agreement surface
+  the founder signs off before code lands.
+
+  THE BUG THIS FIXES (the founder's fury): the topology CARD derives
+  openbao active-passive from the matrix (raft · async, raft-transition,
+  60s/30s, warm-standby — CORRECT). The "Change placement" EDIT dropdown
+  used a generic hardcoded describeMode() that said "cold/warm standby
+  (backup-restore)" — WRONG, matrix-blind, and logically contradicting the
+  card. THE FIX: BOTH the card AND the editor read the SAME perTopology
+  contract. One source, zero contradiction.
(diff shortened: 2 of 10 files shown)


emrahbaysal:
Reconciled into the agreed #3925 Jobs-view model by #3928. That PR is cut from this branch, so #3918's `taskBaseName` identity keying + `statusFromReadyCondition` sticky-state discipline carry over **verbatim** — and #3920's `isDay2ScannerJob` detection is **re-targeted** from exclude → collapse (each scanner becomes ONE identity-keyed row with its run-history, never 600 rows and never dropped, per `jobs-convergence-monitor-model.md` §4 Surface B). Hold/supersede this in favour of #3928.

emrahbaysal:
Superseded by #3930. Its `taskBaseName` keying + `statusFromReadyCondition` sticky-terminal discipline (only Flux's terminal `Stalled` is a genuine terminal failure; every other `Ready=False` is in-flight reconcile) + the `Stalled`/read-side anti-flap (`helmwatch.go` + `jobs_backfill.go`) are folded INTO #3930, which lands the full #3925 Jobs-view identity-collapse cleanly off current `main` (MERGEABLE, 16 files). This branch was stacked on the drifted chain that left #3928 conflicting at 200 files; #3930 re-cuts clean. Recommend closing in favour of #3930.

emrahbaysal:
Closed as superseded per the PATH TO 100 plan Phase A adjudication (2026-08-03), verified file-level against current main — not by branch age.

Specifically: the Flux continuous-reconcile conflation was eliminated at a deeper level by the #3646 typed-Kind rebuild (generic reconciler ingestion: cron/job/kustomization/reconciler-deploy as distinct Kinds, de-merged JobsPage). Main carries its own evolved jobs_backfill.go. Same stray-file contamination as #3920.

If any hunk here is still believed live, file an issue against the current implementation rather than rebasing a June branch.

PR #3932, opened 2026-06-19
fix(catalyst-api): Open-button externalURL + Topology placement for mgmt-vCluster apps post-#3642

emrahbaysal:
## Problem (grounded live on hw171, dep `3963e9267c10a90d`, catalyst-api/ui:fb423be)

The AppDetail "Open" button + AppsPage card button **vanished** and the **Topology placement read back empty** for EVERY front-door app (gitea/grafana/harbor/keycloak/openbao/guacamole/newapi). `GET /api/v1/sovereign/apps` + the per-app detail endpoint returned `externalURL='' placement=''` for every app — including `bp-openbao`/`bp-grafana`/`bp-gitea` whose front doors are live.

The UI is correct + deployed: `AppDetail.tsx` renders the Open CTA only when `apiApp.externalURL` is set, and reads `apiApp.placement`. **The backend projection was the bug.**

## Root cause — #3642 host→mgmt-vCluster migration regression (same family as #3927)

Confirmed live on hw171:

| What | Where it lives now |
|---|---|
| HelmReleases (gitea/grafana/openbao/…) | host ns `mgmt`, `spec.targetNamespace=gitea\|grafana\|openbao\|…` |
| **HTTPRoutes** (the front-door source) | host ns **`mgmt`** (`mgmt/gitea`, `mgmt/grafana`, `mgmt/openbao`, `mgmt/harbor`, `mgmt/keycloak`…) |

The loft-sh syncer mirrors each in-vCluster HTTPRoute back onto the HOST, landing it in the vCluster HOST sync namespace (`hostNamespace: mgmt` in bp-mgmt-vcluster), **NOT** the app's `targetNamespace`.

- `lookupExternalURL` (detail) + the `hrURL` join (list) both hard-required `route.namespace == targetNamespace` (the pre-#3642 invariant, when an app and its route shared a host namespace) → the `mgmt/*` route matched nothing → empty URL → no Open button.
- `synthesiseAppFromHelmRelease` (the HR-synth path ALL these CR-less bootstrap apps take, since they have no Application CR) never set `resp.Placement` → empty Topology placement.

## Fix (catalyst-api only — lookup logic is env-independent)

- **`routeNamespaceMatchesApp()`** widens the route-namespace filter to also accept the well-known vCluster host sync namespaces (`mgmt`/`dmz`/`rtz`) — **never arbitrary** namespaces, so the #3150 cross-namespace false-positive guard survives. The downstream name / host-leftmost-label / backendRef match rules are collision-free across apps (each owns a distinct `<release>.<fqdn>` host).
- The **list-endpoint `hrURL` join** probes `targetNamespace` first, then the sync namespaces.
- **`placementFromHR()`** projects a canonical placement class (singleton/active-passive/active-hot-standby/active-active) for HR-synth apps from `spec.values.placement` | `spec.values.topology.mode` | the topology label | `singleton` default — normalised via `placementForTopology` (#3375 one-vocabulary).
- The **#3224 user-UI suppression** for genuinely-headless apps stays intact.

## Tests
- `mgmt-vcluster synced route resolves` (via sync-ns + via hostname-leftmost label)
- `unrelated namespace still excluded after mgmt widening` (the #3150 guard survives)
- `TestRouteNamespaceMatchesApp` table + `TestPlacementFromHR` table
- `go build ./...` + `go vet` + full handler suite (94s) all green.

## Sibling regression found (NOT in this PR — distinct, larger surface)
The AppDetail **Resources/Logs tabs** query `targetNamespace` (`gitea`) for pods the syncer mirrors to host ns `mgmt` as `<pod>-x-gitea-x-mgmt-vcluster`, so they also come back empty post-#3642. Needs in-vCluster→host-mirror namespace + label-mangling resolution (not a simple namespace allow-list) — tracked separately.

## Verification note
The lookup-logic fix + unit tests are env-independent and proven here. Full confirmation that `externalURL`/`placement` re-populate on the live API requires a catalyst-api deploy carrying this commit (the running pod is `fb423be`).

Refs #3642 #3375 #3150 #3687


emrahbaysal:
Superseded by #3933 — re-cut of the same 3-file fix cleanly onto current `origin/main`. The merge conflict on this branch came entirely from a now-superseded duplicate of the Jobs-view identity-collapse work that already landed on main via #3930 (`5676e475d`). The externalURL/placement fix logic in #3933 is byte-identical (295+/29-, same 3 files). Closing in favour of #3933.

PR #3737, opened 2026-06-17
fix(bootstrap-kit): revert keycloak host-bridge — SHARED_PG terminal convergence wedge (keycloak-database-secret can't cross host→vCluster on OSS vcluster 0.21.0) [Refs #3736]

emrahbaysal:
## What & why

Fixes the **terminal fresh-prov convergence wedge on `SOVEREIGN_ENABLE_SHARED_PG=true`** — the last ~11 HelmReleases never converge because keycloak wedges.

**Root cause (verified live, hw157 shape).** Since #3642/#3703 keycloak installs **inside the `vc-mgmt` vCluster** (`09-keycloak.yaml` `kubeConfig → vc-mgmt`; `placement.yaml` `vcluster: mgmt`). On SHARED_PG keycloak flips to `externalDatabase` mode and mounts the reflector-pushed `keycloak-database-secret` (`existingSecret`). But slot 16a mints that Secret in the **HOST** ns `shared-data`, the host emberstack-reflector pushes it to the **HOST** ns `keycloak`, and keycloak-0 runs **inside vc-mgmt** — a different apiserver. Nothing bridges the Secret across the boundary:

```
keycloak-0  MountVolume.SetUp failed:
  secret "keycloak-database-secret-x-keycloak-x-mgmt-vcluster" not found  (8m+)
```

→ no OIDC seed into OpenBao → every downstream `*-sso-oidc-credentials` ExternalSecret (harbor/grafana/hubble/guacamole) `SecretSyncedError` → the SSO-consuming HRs hang.

**Why the proper bridge isn't available here.** The host→vCluster delivery mechanism (`sync.fromHost.secrets`) **does not exist on the OSS vcluster 0.21.0** we run — source-verified against tag-pinned loft-sh/vcluster: the field is absent from `SyncFromHost` until **0.23.0**, and 0.21.0's `UnmarshalYAMLStrict` + `additionalProperties:false` schema **reject the unknown key at config-parse (breaking the syncer)**. (From 0.23.0 it IS genuinely OSS, not Pro-gated.) An in-vCluster ExternalSecret can't reconcile either — ESO is a host-minimum substrate and `sync.toHost.customResources` is Pro-gated/inert on OSS.

## The fix — revert keycloak to HOST placement (the #3733 grafana-revert precedent)

[...]

PR #3950, opened 2026-06-20
fix(bp-mgmt-vcluster): #3642 spine — replicate gitea-pg-rw/-ro Services host→vc-mgmt (0.2.28)

emrahbaysal:
Refs #3948 #3642 #3737 — follow-up to #3949 (the SERVICE half of the same gitea spine wedge).

#3949 fixed the gitea-pg-app SECRET so the gitea pod could mount its DB password — but the pod then CrashLooped one layer deeper:

```
configure-gitea: dial tcp: lookup gitea-pg-rw.gitea.svc.cluster.local: no such host
```

gitea ships its OWN dedicated `gitea-pg` CNPG Cluster (host ns `gitea`). After #3642 the gitea pod runs in vc-mgmt and dials `gitea-pg-rw.gitea.svc.cluster.local:5432` (`values.yaml gitea.database.HOST`), but the CNPG `-rw`/`-ro` Services live host-side only → NXDOMAIN inside vc-mgmt → gitea `Init:CrashLoopBackOff` → gitea-http never serves → `catalyst-gitea-token-mint` hook retries forever → bp-catalyst-platform install hangs → **console never comes up** (hw172, dep `b50ba61d9cb9e157`).

**Fix:** add `gitea/gitea-pg-rw` + `gitea/gitea-pg-ro` to `replicateServices.fromHost` — the exact SERVICE-half gap already covered for guacamole-pg + newapi-pg. Additive (no-op until the host-seams HR renders the gitea-pg Cluster).

Lockstep: chart + blueprint + slot pin (`58-bp-mgmt-vcluster`).

### DIFF
--- a/clusters/_template/bootstrap-kit/58-bp-mgmt-vcluster.yaml
+++ b/clusters/_template/bootstrap-kit/58-bp-mgmt-vcluster.yaml
@@ -182,7 +182,7 @@ spec:
       # `harbor-core.harbor.svc` — resolves Harbor again post-#3642 (it now runs
       # INSIDE vc-mgmt; the host name NXDOMAIN'd → cutover wedged at step-02 on
       # hw171). Same proven host-service-aliases mechanism as gitea/keycloak/openbao.
-      version: 0.2.27
+      version: 0.2.28
       sourceRef:
         kind: HelmRepository
         name: bp-mgmt-vcluster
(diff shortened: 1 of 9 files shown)


emrahbaysal:
Superseded by #3953 (the comprehensive #3642 secret+service cross-boundary cascade fix). Your intent is folded: the gitea-pg-rw/-ro SERVICE-half replication is now part of the bp-mgmt-vcluster 0.2.28 replicateServices.fromHost block (alongside harbor-pg + the rtz cross-vc valkey/seaweedfs mirrors). Closing as superseded to avoid the bp-mgmt-vcluster version-pin collision.

PR #3951, opened 2026-06-20
fix(bp-catalyst-platform): gitea-token-mint readiness probe is 403-tolerant (1.4.723)

emrahbaysal:
Refs #3948 #3642 — follow-up to #3949/#3950, removes a 14min dead-wait on the console-up critical path.

The `catalyst-gitea-token-mint` pre-install hook's readiness loop used `curl -sSf /api/v1/version`. With gitea's `REQUIRE_SIGNIN_VIEW` enabled, the unauthenticated probe returns **HTTP 403** (`Only signed in user is allowed to call APIs`) — and `-f` (fail on ≥400) means the loop NEVER breaks, so the hook spun the full 168×5s=**14min** budget on every prov before falling through to the mint. Step 4 (the actual mint) is authenticated (`-u GU:GP`) so it succeeded anyway — the 14min was pure dead time delaying `bp-catalyst-platform` install → console.

**Fix:** treat ANY HTTP answer (incl. 403) as 'API reachable'; only a connect failure (NXDOMAIN/refused → curl code `000`/empty) keeps waiting.

**Verified live on hw172** (dep `b50ba61d9cb9e157`): once gitea served, the token minted immediately and `bp-catalyst-platform` install proceeded to deploy the console/api workloads.

Lockstep: chart + slot pin (`13-bp-catalyst-platform`).

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1052,7 +1052,9 @@ spec:
       #   Jobs so the Kyverno flux-managed (Enforce) policy admits them in
       #   catalyst-system. Pre-1.4.722 the pre-install hook was DENIED →
       #   console install failed pre-install on every Enforce Sovereign (hw172).
-      version: 1.4.722
+      # - 1.4.723 (#3948): gitea-token-mint readiness probe is 403-tolerant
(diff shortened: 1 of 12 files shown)


emrahbaysal:
Superseded by #3953 (the comprehensive #3642 secret+service cross-boundary cascade fix). Your intent is folded: the gitea-token-mint 403-tolerant readiness probe is now in bp-catalyst-platform 1.4.724. Closing as superseded to avoid the bp-mgmt-vcluster version-pin collision.

PR #4015, opened 2026-06-21
fix(catalyst-api): deliver region-b kubeconfig to in-cluster catalyst-api — kill false topology 'singleton' at its DELIVERY root (Refs #4000 #3986)

emrahbaysal:
Deliver the region-b (secondary) kubeconfig to the in-cluster catalyst-api so the placement resolver sees both regions — the real DELIVERY root of the false topology singleton (deeper than the #4001 on-disk self-heal, which no-ops when the kubeconfig is never delivered). The in-cluster pod previously started clusterCount:0 / sovereigns:1 with an empty kubeconfigs dir. Refs #4000 #3986.

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/deployment_handover_export.go
+++ b/products/catalyst/bootstrap/api/internal/handler/deployment_handover_export.go
@@ -157,12 +157,33 @@ func (h *Handler) exportSecondaryKubeconfigsToChild(dep *Deployment, fqdn, depID
 	dep.mu.Lock()
 	regions := append([]string(nil), regionKeysForExport(dep)...)
 	dep.mu.Unlock()
+	dir := secondaryKubeconfigsDir()
+
+	// #4000 — UNION the spec-derived region keys with the keys ACTUALLY on
+	// disk. The spec keys are reconstructed as `<CloudRegion>-<idx>`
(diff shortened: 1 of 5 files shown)


emrahbaysal:
## ✅ Live 2-target proof captured on hw174 (after env image-pull recovered)

Chroot rebooted, loaded the persisted private-IP-rewritten region-b kubeconfig at boot:
```
"k8scache: data plane started" sovereigns:2
36 × "informer synced" cluster:ea30d1d816f2eee2-me-east-215-b-1
```
**bp-alloy placement flipped singleton → active-active (2 targets):**
```json
{"targets":[
  {"region":"hw-me-east-215-a-rtz-prod","cluster":"ea30d1d816f2eee2","role":"Primary"},
  {"region":"hw-me-east-215-b-rtz-prod","cluster":"ea30d1d816f2eee2-me-east-215-b-1","role":"Primary"}
],"derivedFromRuntime":true}
```
grafana/keycloak/openbao stay 1-target HONESTLY because region-b is mid-converging (live HR 37 True / 22 False / 7 pending): region-b alloy pods exist (DaemonSet) → 2 targets; region-b grafana pods absent → 1 target. bp-catalyst-api correctly stays 1 target (singleton control-plane). Full evidence on #4000.

Issue #3998, opened 2026-06-20
fix(cloud): network+security coverage gap — Gateway/HTTPRoute/NetworkPolicy/CiliumNetworkPolicy invisible; Load Balancers misrepresents the real EIP+Gateway front door


emrahbaysal:
## Problem
The Cloud view (`/cloud?view=graph` + per-kind list) shows **0** for resources that are LIVE on the Sovereign — and for Load Balancers shows a number that is technically correct but hides how the platform is actually fronted. Same family as #3987 (per-kind list empty) and the CSI/local-path gap: the view does not represent kinds that exist.

## Ground reality (live, hw173 / dep 7bb723da8da06047, region-a)
| Kind | Live count | Cloud-view page today | Verdict |
|---|---|---|---|
| Gateway (gateway.networking.k8s.io) | **1** (`kube-system/cilium-gateway`) | **none** | REAL GAP — the actual front door is invisible |
| HTTPRoute | **15** | **none** | REAL GAP — the routes serving console/keycloak/gitea/harbor are invisible |
| NetworkPolicy | **8** | **none** | REAL GAP — security posture invisible |
| CiliumNetworkPolicy | **5** | **none** | REAL GAP |
| CiliumClusterwideNetworkPolicy | 0 | none | (none live yet) |
| Load Balancers (cloud-provisioned) | 0 ELB on Huawei | page sources `topology.loadBalancers[]` | CORRECT-but-MISLEADING |
| Services | 182 (1 type=LoadBalancer = clustermesh-apiserver) | page exists | OK after #3987/#3997 |
| Ingress | 0 | page exists | HONEST 0 (platform uses gateway-api, not Ingress) |

## How the platform is actually reachable (the "if 0 LBs, how?" answer)
NOT via a cloud load balancer. Front door = **Cilium Gateway** (gateway-api, PROGRAMMED=True) + **15 HTTPRoutes**, exposed via a **Huawei EIP DNAT'd to the node port** (kom4dc nodeIPAM/DNAT pattern). On Hetzner the same deploy provisions a real `hcloud` LB (would show >=1). The LB page reads only cloud-provider `topology.loadBalancers[]`, which is empty on Huawei → 0/0, hiding the real datapath. The one actual `type=LoadBalancer` Service (`clustermesh-apiserver`, 212.72.24.30) is also not shown.

## Scope (ONLY-ADD; no regression to the existing #3970/#3981/#3987 explorer)
[...]

PR #4019, opened 2026-06-21
fix(cloud-view): surface the REAL front-door LB from the deployment record, not the empty XRC layer (Refs #3998 #3987)

emrahbaysal:
## What

The Catalyst console cloud-view Load Balancers page rendered **0/0** on a real Sovereign even though the front door is live — because the live chroot path (`buildRegionFromLiveNodes`, the branch every running Sovereign hits since catalyst-api runs in-cluster) hardcoded `LoadBalancers: []LoadBalancer{}`, ignoring the deployment record's `Result.LoadBalancerIP`. The empty Crossplane XRC layer (`LoadBalancerClaim`) was never the right source — on a real Sovereign that whole OpenOva day-2 XRC layer is 0 instances; infra is 100% OpenTofu-provisioned.

## Fix (Path A — the approved lighter path)

- **Source `loadBalancers[]` from where the data actually lives** — the deployment record (`Result.LoadBalancerIP`) — in BOTH the declared (mothership) path and the live chroot path, via a shared `frontDoorLBs()` helper. Never the empty XRC layer.
- **Structured contract**: add `Listeners[]` / `Targets[]` to the Go `LoadBalancer` type so it matches the UI `LoadBalancerSpec` (`LBListener` / `LBTarget`) — the listeners column now renders the real `tcp:80/443/6443` instead of `—`.
- **Front-door descriptor**: add `FrontDoorKind` (`cloud-lb` | `gateway-eip`) so the view explains the real ingress datapath provider-agnostically — a real cloud LB (Hetzner) vs an EIP DNAT'd to the Cilium Gateway NodePort (Huawei kom4dc). Surfaced as a "Front door" row in the LB detail drawer.
- **Worker nodes** are already sourced from live K8s (not XRCs) — left exactly as-is.
- Network/security kinds (Gateway / HTTPRoute / NetworkPolicy / CNP / CCNP) for #3998 scope item 1 were already wired on the base branch.

## MUST-PRESERVE (no regression)

- The populated `regions` / `clusters` / `networks` tree (sourced from the deployment record) is untouched.
- The k9s-style per-kind cloud-list explorer + every existing column is untouched — this is ONLY-ADD on the LB seam.
- Honest 0s stay 0: the LB slice is empty (no fabricated row) when `LoadBalancerIP` isn't known yet.

## Before / After (cloud-view diff)

| Surface | Before | After |
|---|---|---|
| regions / clusters / networks | populated | **unchanged** |
| Worker Nodes | live K8s nodes | **unchanged** (still live K8s) |
| Load Balancers (live chroot path) | **0 rows** | **1 row** — real EIP/LB from the record, with listeners + front-door kind |
| Load Balancers listeners column | `—` (no `listeners[]` in JSON) | `tcp:80, tcp:443, tcp:6443` |
| LB detail "Front door" | absent | `EIP → Cilium Gateway (NodePort)` (Huawei) / `Cloud load balancer` (Hetzner) |

## Tests

- 3 new Go loader tests: live-chroot regression guard (non-empty LB from record + `gateway-eip` for Huawei + 2 live nodes), declared `cloud-lb` path, and no-placeholder guard.
- `go build ./...` + `go vet ./...` green; full `internal/infrastructure` + `internal/handler` packages green.
- UI `tsc -b` typecheck green; `cloud-network` + `cloud-list` vitest 60/60 green. (Unrelated pre-existing UI test failures in PinInput6 / ProvisionPage / Settings / wizard are untouched by this PR.)

## Backlog

Path B (provisioning instantiates the Crossplane XRCs so the XRC layer truly mirrors the OpenTofu infra — the heavier object-model-live decision) is filed as **#4018** for founder A-vs-B steer; not built here.

## Live verify

Best-effort on hw174 (`ea30d1d816f2eee2`); the durable proof is the orchestrator's cycle-2 fresh prov. hw174's chroot catalyst-api image may currently be ImagePullBackOff (ghcr-blob DNS flake on the IPv4-only kom4dc VPC) — if the live console is env-blocked, that is not this fix failing.

Refs #3998 #3987

### DIFF
--- a/products/catalyst/bootstrap/api/internal/infrastructure/topology_loader.go
+++ b/products/catalyst/bootstrap/api/internal/infrastructure/topology_loader.go
@@ -398,17 +398,6 @@ func buildRegionFromLiveNodes(ctx context.Context, in LoaderInput) (Region, bool
 	if in.Region != "" {
 		clusterID = "cluster-" + in.DeploymentID + "-" + in.Region
 	}
-	cluster := Cluster{
-		ID:            clusterID,
-		Name:          clusterName,
-		Version:       "v1.31",
-		Status:        in.Status,
-		NodeCount:     len(allNodes),
-		VClusters:     loadVClusters(ctx, in),
-		LoadBalancers: []LoadBalancer{},
-		NodePools:     pools,
-		Nodes:         allNodes,
-	}
 	// Provider selection: prefer the wizard-declared provider; if the
 	// chroot's synthesised deployment has no Provider (G14, Refs
 	// #2551), sniff the first Node's provider hints before defaulting.
@@ -416,6 +405,9 @@ func buildRegionFromLiveNodes(ctx context.Context, in LoaderInput) (Region, bool
 	// "huawei://..." / "hcloud://...") + sometimes a label like
 	// "topology.kubernetes.io/cloud-provider". Default "hetzner" is
 	// kept only as the absolute last resort.
+	//
+	// Resolved BEFORE the cluster is built so frontDoorLBs can pick the
+	// correct FrontDoorKind (gateway-eip on Huawei vs cloud-lb elsewhere).
 	provider := in.Provider
 	if provider == "" {
 		for _, n := range list.Items {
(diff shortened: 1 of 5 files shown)


emrahbaysal:
**DONE** (backlog refactor 2026-06-24). This issue IS the merged Path-A implementation PR (#3998): cloud-view Load Balancers now source from the deployment record (`Result.LoadBalancerIP` via `frontDoorLBs()`) with structured Listeners/Targets + FrontDoorKind, never the empty XRC layer. UAT rows **203/204 ✅** (LB + worker-node counts render real values live). Closing.

PR #4066, opened 2026-06-21
feat(provisioning): storage class is a choosable user input with per-provider CSI defaults (#4057)

emrahbaysal:
## What & why

Founder point #1 (verbatim): *"storage class is an INPUT that needs to be chosen by the user and like all the other choosable options it needs to have its defaults as well."* Connects to the #3971 / #892 governance: local-path is FORBIDDEN; the per-provider cloud CSI must be the cluster-default StorageClass.

## The Huawei EVS CSI auth gap — already resolved on main, LIVE-PROVEN on hw182

The "EVS CSI can't authenticate" problem from the prior branches is **already fixed on `main`** (the `idc` no-Keystone AK/SK bypass, merged via #4013) and is **live-working on hw182 (dep `b78e8dcd13e6e6d6`) right now**:

- EVS CSI controller 2/2 + node 6/6, **0 restarts, 5h28m uptime** — no CrashLoop.
- Driver log: `Successfully obtained volume list, size: 189` — real SDK-HMAC-SHA256-signed EVS API calls returning 189 actual volumes. **Zero APIGW.0101 / Keystone / 401 errors.**
- `evs-ssd` is the single cluster-**default** StorageClass (`storageclass.kubernetes.io/is-default-class: "true"`), provisioner `evs.csi.huaweicloud.com`, reclaimPolicy Retain.
- cloud-config secret carries `idc=true` + correct AK/SK/project-id/region/cloud/auth-url.
- **35+ PVCs Bound on evs-ssd.** The 4 Pending PVCs are `WaitForFirstConsumer` waiting on CPU-starved pods (`0/6 nodes: Insufficient cpu`) — that is #4055 (under-provisioned Huawei workers), **not** a CSI fault.

So the chart/cluster half of #3971/#892 (EVS chart 1.0.3 idc bypass, ghcr-pull imagePullSecret, snapshotclass-off, dependsOn bp-reflector, both CSIs default-class, k3s `--disable=local-storage`, K23 Kyverno ENFORCE deny) is **done on main**. The remaining gap was the **INPUT**: `SOVEREIGN_CNPG_STORAGE_CLASS` was a hardcoded per-provider literal with no operator override. This PR closes that gap.

## Changes

**Backend (catalyst-api)**
- `provisioner.Request.StorageClass` (+ `RegionSpec.StorageClass`) — optional input.
- `DeriveStorageClass` / `defaultStorageClassForProvider` — one source of truth: explicit value verbatim, else the per-provider durable CSI default (`hetzner`→`hcloud-volumes`, `huawei`→`evs-ssd`). Never empty, never local-path.
- `writeTfvars` emits `default_storage_class`; `Validate` folds `Regions[0]` into the umbrella; `store.Redact` / `ToProvisionerRequest` persist it; `State()` surfaces the effective class for the Settings page.

**Infra (tofu)**
- `default_storage_class` string var on both providers (validation rejects `""` and `"local-path"`); per-provider default keeps a fresh prov byte-identical.
- cloud-init `SOVEREIGN_CNPG_STORAGE_CLASS` now renders `${default_storage_class}` instead of the hardcoded `hcloud-volumes` / `evs-ssd` literals.

**UI (wizard)**
- `storageClass` on `WizardState` + `setStorageClass` + an optional "Default storage class" input in `StepProvider` (per-provider placeholder) + the POST body in `StepReview`.

**No-chart-relies-on-local-path**
- The one remaining `local-path` literal (`13-bp-catalyst-platform.yaml` QA-fixtures `cnpgStorageClass`) flips `local-path` → empty so a QA Sovereign's qa-fixture CNPG uses the cluster-default durable class (the K23 Enforce policy would otherwise deny a local-path PVC).

## Verification

- `go build ./...`, `go vet ./...` clean.
- New Go tests: 14 cases (per-provider default, explicit honored, per-region fold, umbrella-wins, local-path never emitted). `provisioner` + `store` + `handler` suites green.
- New tofu cloud-init render assertions: default (`hcloud-volumes` / `evs-ssd`) + operator override land in `SOVEREIGN_CNPG_STORAGE_CLASS`, never `local-path`. **Both providers' tofu test suites pass: hetzner 12/12, huawei 7/7.**
- `scripts/lint-cloudinit.sh both` renders + `dash -n` passes for both providers.
- EVS chart `helm template` (enabled, default-class) renders `evs-ssd` + `is-default-class: "true"` + `provisioner: evs.csi.huaweicloud.com`. bootstrap-kit `kustomize build` clean.
- UI `tsc -b --noEmit` clean.

## Confirmations
- **local-path is FORBIDDEN**: tofu var validation rejects it; k3s `--disable=local-storage` + K23 Kyverno ENFORCE deny (on main); the one QA literal removed here.
- **per-provider CSI is the cluster DEFAULT**: `hcloud-volumes` (Hetzner) / `evs-ssd` (Huawei), annotated `is-default-class: "true"` (on main); the operator-chosen class threads through the same seam.
- **No Hetzner regression**: hcloud-csi default + tofu tests unchanged-green; the var defaults to `hcloud-volumes` on Hetzner.

Refs #4057 #3971 #892

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1493,7 +1493,16 @@ spec:
       cnpgPrimaryRegion: ${QA_CNPG_PRIMARY_REGION:-hz-fsn-rtz-prod}
       cnpgReplicaRegion: ${QA_CNPG_REPLICA_REGION:-hz-fsn-rtz-prod}
       cnpgImage: ${QA_CNPG_IMAGE:-ghcr.io/cloudnative-pg/postgresql:16.4-1}
-      cnpgStorageClass: ${QA_CNPG_STORAGE_CLASS:-local-path}
+      # #4057/#3971 — empty default ⇒ the qaFixtures CNPG Cluster CR OMITS
+      # storageClass (the chart's `{{- with .Values.qaFixtures.cnpgStorageClass }}`
+      # guard), so CNPG uses the cluster-DEFAULT durable cloud CSI class
+      # (hcloud-volumes / evs-ssd). The former `local-path` default is now
+      # FORBIDDEN — k3s runs `--disable=local-storage` and the K23 Kyverno
+      # ENFORCE policy DENIES any PVC referencing local-path, so a QA Sovereign
+      # left on the old default had its qa-fixture CNPG PVCs rejected. An
+      # operator may still pin a specific class via QA_CNPG_STORAGE_CLASS, but
+      # local-path is never a valid value.
+      cnpgStorageClass: ${QA_CNPG_STORAGE_CLASS:-}
       cnpgStorageSize: ${QA_CNPG_STORAGE_SIZE:-1Gi}
       # Kyverno baseline policies (Fix #37). disallow-privileged-containers
       # ships in Enforce mode; the other 18 baseline policies in Audit so
--- a/infra/providers/_shared/cloudinit-control-plane.tftpl
+++ b/infra/providers/_shared/cloudinit-control-plane.tftpl
@@ -645,7 +645,14 @@ write_files:
             # that NAMES the class instead stays Pending (not frozen) until the
             # class exists, then binds — race-immune. On Hetzner the CSI default
             # is hcloud-volumes.
-            SOVEREIGN_CNPG_STORAGE_CLASS: "hcloud-volumes"
+            #
+            # #4057 — the class is now the operator-chosen
+            # `default_storage_class` tofu var (catalyst-api
+            # Request.StorageClass, defaulted to the per-provider durable CSI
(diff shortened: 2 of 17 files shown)


emrahbaysal:
CI note: every functional check is green (integration kind+gitea, test, validate+fmt+test ×2, all Playwright, dash -n cloud-init render, manifest-validation, drift, pin-sync, dependency-graph, scan, vendor-coupling).

The single red 'Reject banned words in PR body' is a STALE check-run (started 17:00:28) — I rewrote the body to drop the forbidden word 'blocker', and the gate re-evaluated GREEN twice afterward (runs 16:58:50 and 17:03:00 both SUCCESS on this same SHA). `gh pr checks` surfaces the older run, but the latest run per check name (what branch protection uses) is SUCCESS and the live body contains zero banned words. A fresh push would clear the display but re-burns the ~12-min integration suite for a cosmetic-only reason; flagging here instead.

PR #4151, opened 2026-06-22
fix(bp-wordpress-tenant): apt sandbox-setuid fails under dropped-ALL-caps init → APT::Sandbox::User=root (0.4.5)

emrahbaysal:
## Problem (diagnosed LIVE on the omantel.biz demo Org, kom4dc)

The per-Org WordPress app could not converge on a fresh Sovereign prov: the `wp-plugin-install` initContainer was stuck in **CrashLoopBackOff**.

`wp-plugin-install` runs `apt-get update && apt-get install -y --no-install-recommends unzip` to extract two payloads onto the wp-content PVC: the `openid-connect-generic` SSO plugin and the `pg4wp` Postgres adapter drop-in.

The container's `securityContext` (from `.Values.wordpress.containerSecurityContext`) **drops ALL Linux capabilities**. apt's `http` transport method tries to drop privileges to the `_apt` user via `setgroups`/`setegid`/`seteuid` — which require `CAP_SETUID`/`CAP_SETGID`. With caps dropped, those syscalls fail:

```
setgroups 65534 failed - Operation not permitted
Method http has died unexpectedly!
E: Sub-process http returned an error code (112)
```

→ apt exits non-zero → init CrashLoopBackOff → the app never reaches Running.

This was surfaced **only after #4147** set `runAsUser:0` (which fixed the earlier file-permission errors writing `/var/cache/apt`, but exposed this apt-sandbox-setuid failure underneath). This fix **stacks on top of #4147**.

## Fix (cleanest, no extra Linux caps)

Pass `-o APT::Sandbox::User=root` to **both** apt-get invocations in the `wp-plugin-install` init command:

- `apt-get update` → `apt-get -o APT::Sandbox::User=root update`
- `... apt-get install ... unzip` → `... apt-get -o APT::Sandbox::User=root install ... unzip`

This is the canonical container-friendly apt setting: apt keeps running as root (no setuid syscall is attempted at all), so it works under a dropped-ALL-capabilities securityContext **without loosening it**. Preferred over re-adding `SETUID`/`SETGID` caps, which a restrictive Kyverno baseline may reject on a different cluster. `runAsUser:0` (from #4147) is kept — apt still needs to write `/var/cache/apt`.

**Live validation:** re-adding `SETUID`/`SETGID`/`DAC_OVERRIDE` caps unblocked apt on the cluster (confirming the diagnosis), but the durable fix uses `APT::Sandbox::User=root` so it carries NO extra caps and stays portable.

## Changes

- `platform/wordpress-tenant/chart/templates/deployment.yaml` — both apt-get calls in the `wp-plugin-install` init heredoc now carry `-o APT::Sandbox::User=root`, with an inline comment explaining the apt `_apt` setuid drop vs the dropped-ALL-caps securityContext.
- `chart/Chart.yaml` + `blueprint.yaml` — version bumped **0.4.4 → 0.4.5 in lockstep** (manifest-validation/lockstep CI gate) + changelog entry.

## Validation

- `helm template` confirms both apt-get calls render with `-o APT::Sandbox::User=root`.
- All three chart test scripts pass (`active-hot-standby-render.sh`, `observability-toggle.sh`, `oidc-config.sh`).

Refs #4139
Refs #3858

### DIFF
--- a/platform/wordpress-tenant/blueprint.yaml
+++ b/platform/wordpress-tenant/blueprint.yaml
@@ -6,7 +6,7 @@ metadata:
     catalyst.openova.io/category: tenant-app
     catalyst.openova.io/section: pts-7-org-tenant
 spec:
-  version: 0.4.4
+  version: 0.4.5
   card:
     title: WordPress (per-Organization)
     summary: |
--- a/platform/wordpress-tenant/blueprint.yaml
+++ b/platform/wordpress-tenant/blueprint.yaml
@@ -6,7 +6,7 @@ metadata:
     catalyst.openova.io/category: tenant-app
     catalyst.openova.io/section: pts-7-org-tenant
 spec:
-  version: 0.4.5
+  version: 0.4.6
   card:
     title: WordPress (per-Organization)
     summary: |
--- a/platform/wordpress-tenant/blueprint.yaml
+++ b/platform/wordpress-tenant/blueprint.yaml
@@ -6,7 +6,7 @@ metadata:
     catalyst.openova.io/category: tenant-app
     catalyst.openova.io/section: pts-7-org-tenant
 spec:
-  version: 0.4.6
+  version: 0.4.7
   card:
     title: WordPress (per-Organization)
     summary: |
--- a/platform/wordpress-tenant/chart/Chart.yaml
+++ b/platform/wordpress-tenant/chart/Chart.yaml
@@ -114,7 +114,7 @@ name: bp-wordpress-tenant
 #   apt's sandbox keeps it running as root (no setuid syscall), needs NO extra
 #   Linux caps (portable under a restrictive Kyverno baseline). Diagnosed live on
 #   the omantel.biz demo Org (kom4dc). Stacks on top of #4147 (runAsUser:0).
-version: 0.4.6
+version: 0.4.7
 appVersion: "6"
 description: |
   Catalyst Blueprint scratch chart for in-vcluster WordPress, one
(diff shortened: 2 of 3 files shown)


emrahbaysal:
## Scope expanded — full init-chain convergence (validated live on omantel.biz demo Org / kom4dc)

This PR now fixes the complete `wp-plugin-install` init-container failure chain (each bug masked the next; all observed live):
1. **apt-sandbox setuid** — apt drops to `_apt` via setgroups/seteuid, forbidden under dropped-ALL-caps → 'Operation not permitted' / http method died (112).
2. **apt egress** — kom4dc can't reach `deb.debian.org` ('Ign … InRelease' → 'Unable to locate package unzip').
   → **Fix 1+2: extract via PHP `ZipArchive`** (always in the wordpress php image) instead of `apt install unzip` — no apt, no caps, no Debian-mirror egress.
3. **OIDC plugin slug 404** — `openid-connect-generic` → 404; correct slug is **`daggerhart-openid-connect-generic`**.
4. **pg4wp tag 404** — pinned `v3.7.0` doesn't exist; latest release is **`v3.3.1`**.
5. **fatal chown** — `chown -R 33:33 /pvc` as uid 33 can't re-own the fsGroup-mounted `/pvc` root → init Error. Removed (php_unzip already extracts 33-owned files).

After these, **both init containers reach `ready=true` (0 restarts)** live.

### Remaining (separate follow-up, NOT in this PR)
The **main wordpress container** then fails one layer deeper: the official image's entrypoint `tar`-extracts WP core and hits `tar: ./wp-content: Cannot change mode to rwxrwxr-x: Operation not permitted` on the wp-content PVC mount (uid 33 can't chmod the root/fsGroup-owned mount). That's a WP-image-entrypoint vs PVC-perms issue (fsGroup / fsGroupChangePolicy or a perms initContainer), tracked separately — this PR's scope is the init chain.

Refs #4139 Refs #3858

Issue #4206, opened 2026-06-23
Secondary region: per-app CNPG PVCs (guacamole-pg/gitea-pg/harbor-pg) stuck Pending with storageClassName="" — created before evs-ssd CSI default landed


emrahbaysal:
## Symptom (live, omantel.biz dep `4635277cae4ffed9`, region-B `me-east-215-b`)

Three per-app CNPG clusters in the region-B mgmt-vCluster have their `-1` PVC stuck `Pending` for 44h:

```
catalyst-system   guacamole-pg-1   Pending   <unset>   44h
gitea             gitea-pg-1       Pending   <unset>   44h
harbor            harbor-pg-1      Pending   <unset>   44h
```

Scheduler: `0/6 nodes are available: pod has unbound immediate PersistentVolumeClaims`.

The CNPG cluster (e.g. `guacamole-pg`) is wedged `Setting up primary` / `Creating primary instance guacamole-pg-1` with 0 instances Ready → the guacamole `jdbc-seed` post-install Job loops `waiting for postgres (N/60)` → DeadlineExceeded → bp-guacamole HR install fails + remediation-uninstalls in a loop. Same cascade wedges gitea + harbor.

## Root cause

The PVCs were created with **`storageClassName: ""`** (empty string) — verified raw:

```
region-B  guacamole-pg-1  .spec.storageClassName = []   (empty)   → Pending
region-A  guacamole-pg-1  .spec.storageClassName = [evs-ssd]      → Bound (35h)
```

An empty-string `storageClassName` explicitly **opts out of dynamic provisioning** (it is NOT the same as omitting the field, which would defaulting-admission-stamp the cluster default). So even though region-B's `evs-ssd (default)` StorageClass IS present and healthy, the PVC will never bind. The CNPG cluster spec is byte-identical in both regions (`{"resizeInUseVolumes":true,"size":"5Gi"}` — no storageClass pinned), so the divergence is at **PVC-creation time**: region-B created these PVCs **before** the `evs-ssd` default StorageClass existed (a fresh-prov ordering race), so the default-StorageClass admission plugin stamped `""` permanently.

This is the P0-STORAGE / CSI-default-ordering class (Refs #3971), distinct from the #4158 SSO/DB secret-delivery work — region-A's identical PVCs bound fine; only region-B (provisioned in a window before its CSI default landed) is affected.

## Fix direction (not yet implemented)

[...]

PR #4214, opened 2026-06-23
fix(d31): name CNPG storageClass + decode Gitea contents envelope — region-B Pending PVCs + org-overlay corruption (Refs #4206 #3785 #4060)

emrahbaysal:
## D31 multi-region placement/storage systemic root

Fixes the systemic root blocking three issues + the live Sovereign "Degraded" banner on omantel.biz dep `4635277cae4ffed9`. Root-caused live across both regions.

### Layer 1 — #4206: region-B per-app CNPG PVCs Pending 46h (empty storageClassName)
The host-shared CNPG `Cluster` CRs (`gitea-pg` / `harbor-pg` / `guacamole-pg`, bootstrap-kit-crs slots 10/19/52) rendered `storage:` with **no** `storageClass`. An *omitted* field lets the admission-default plugin freeze `""` the instant a PVC is created **before** the per-provider default StorageClass (`evs-ssd` on Huawei) lands — the fresh-prov ordering race the cloud-init comments already document — and that `""` is **permanent** (opts out of dynamic provisioning) → PVCs Pending 46h while the healthy `evs-ssd (default)` SC sits unused.

**Fix:** NAME the class via the existing per-provider durable-CSI var `SOVEREIGN_CNPG_STORAGE_CLASS` (`evs-ssd` on Huawei / `hcloud-volumes` on Hetzner; the tofu `default_storage_class` var validation rejects `""` and `local-path`, so it can never render empty/ephemeral). Threaded into the `bootstrap-kit-crs` Kustomization `postBuild.substitute` (it previously only passed `SOVEREIGN_FQDN`) + the three `Cluster` CR `storage:` blocks (`placement.yaml` source-of-truth + regenerated `bootstrap-kit-crs` slots via `render-slot-placement.py`). **Supersedes the #4107 OMIT shape** (which correctly rejected the empty-string class but left the omit-race open).

### Layer 2 — #3785/#4206: the TRUE systemic root of the org-overlay corruption
`core/services/provisioning/github/client.go` `ReadFile` returned the **raw Gitea contents-API JSON envelope** (`{"content":"<base64>","encoding":"base64",…}`) verbatim — Gitea ignores the `Accept: application/vnd.github.raw+json` header that only GitHub.com honours. The org-tenant `rebuild` closure read this envelope, spliced the new Org subdir onto it, and committed the JSON-wrapped string back as the **parent** `clusters/<fqdn>/org-tenants/kustomization.yaml`. Flux/kustomize cannot parse it → the **whole org-tenants tree (incl. the demo Org `7283eb4a`) stopped reconciling** and its namespace was pruned (stuck Terminating on 3 `application.apps.openova.io` finalizers). This is the live "wordpress HR Ready=False → Sovereign DEGRADED" trail's deepest cause.

**Fix:** `ReadFile` now decodes the Gitea envelope (base64, newline-stripped) to raw content, with a fall-through for genuine GitHub raw bodies. New regression test `TestReadFile_GiteaContentsEnvelope_Decoded` covers the 200+base64 path — the prior contents test only covered 404/create, so this slipped through.

### Layer 3 — #4060: CNPG-pair provider-aware storageClass
Verified **already fixed** on main — `generateCNPGPair` is provider-aware via `cnpgStorageClass()` (Huawei→`evs-ssd`, Hetzner→`hcloud-volumes`), wired from `CLOUD_PROVIDER`. Unit tests `TestCNPGPair_StorageClass_{Huawei,Hetzner,DefaultsToHetzner}` + `TestCNPGStorageClass_Resolver` pass. No code change; covered here for the D31 story.

### Demo Org hot-standby
The demo customer WordPress is a **single-region singleton** (North Star). The org overlay already defaults `SOVEREIGN_ENABLE_HOT_STANDBY` **OFF** (`organization_gitops.go:579-593`, with defence-in-depth) — so a fresh prov renders **no** cross-region `wordpress-db-replica`. The live env carried a *stale* `pg.activeHotStandby.enabled=true` from an older image render; dropped via the live org-overlay repair (the misplaced replica's pgbasebackup pod was Pending forever on region-B node affinity in a region-A namespace → HR Ready=False → DEGRADED banner).

## Validation
- `tofu validate` (hetzner provider): **Success — configuration valid** (cloud-init tftpl edit).
- `go build ./...` + `go vet`: clean.
- `go test ./github/` (incl. new ReadFile test) + `go test ./gitops/` (#4060 provider tests) + org-overlay hot-standby tests: **pass**.
- Live before/after on omantel.biz documented in the issue threads.

## DoD
- region-B Pending PVCs → bound (chart fix prevents recurrence on every fresh prov)
- demo wordpress HR → Ready (replica dropped; namespace recovered)
- Sovereign banner → not Degraded
- org-overlay corruption → fixed at root (no recurrence on next provision)

Refs #4206 #3785 #4060

### DIFF
--- a/clusters/_template/bootstrap-kit-crs/10-gitea.yaml
+++ b/clusters/_template/bootstrap-kit-crs/10-gitea.yaml
@@ -150,6 +150,7 @@ spec:
       memory: 512Mi
   storage:
     size: 5Gi
+    storageClass: ${SOVEREIGN_CNPG_STORAGE_CLASS}
   enableSuperuserAccess: true
   monitoring:
     enablePodMonitor: false
--- a/clusters/_template/bootstrap-kit-crs/19-harbor.yaml
+++ b/clusters/_template/bootstrap-kit-crs/19-harbor.yaml
@@ -106,6 +106,7 @@ spec:
       memory: 512Mi
   storage:
     size: 5Gi
+    storageClass: ${SOVEREIGN_CNPG_STORAGE_CLASS}
   enableSuperuserAccess: true
   monitoring:
     enablePodMonitor: false
--- a/clusters/_template/bootstrap-kit-crs/52-bp-guacamole.yaml
+++ b/clusters/_template/bootstrap-kit-crs/52-bp-guacamole.yaml
@@ -88,6 +88,7 @@ spec:
       memory: 512Mi
   storage:
     size: 5Gi
+    storageClass: ${SOVEREIGN_CNPG_STORAGE_CLASS}
   enableSuperuserAccess: true
   monitoring:
     enablePodMonitor: false
--- a/clusters/_template/bootstrap-kit/placement.yaml
+++ b/clusters/_template/bootstrap-kit/placement.yaml
@@ -412,14 +412,20 @@ spec:
                   maintenance_work_mem: 64MB
                   log_statement: ddl
                   log_min_duration_statement: "1000"
-              # #4107 — OMIT storageClass entirely. An empty-string storageClass
-              # is NOT "cluster-default" — it means "use NO StorageClass", which
-              # produces an immediate-binding PVC with no provisioner that stays
-              # Pending forever (FailedBinding: "no persistent volumes available
-              # for this claim and no storage class is set"). Omitting the field
-              # is what makes cnpg fall back to the default StorageClass (evs-ssd
-              # on Huawei, hcloud-volumes on Hetzner). local-path remains
-              # forbidden via the forbid-local-path-storage Kyverno policy.
+              # #4206 (supersedes the #4107 OMIT) — NAME the storageClass via
+              # the per-provider durable-CSI var instead of omitting it. #4107
+              # correctly rejected an EMPTY-string class (which opts out of
+              # provisioning → Pending forever), but OMITTING the field is also
+              # unsafe: the admission-default plugin freezes "" the instant a
+              # PVC is created BEFORE any default SC exists (the fresh-prov
+              # ordering race), and that "" is PERMANENT → the gitea/harbor/
+              # guacamole-pg-1 PVCs stuck Pending<empty> 46h (region-B, dep
+              # 4635277cae4ffed9). Naming the class makes the PVC RACE-IMMUNE:
+              # it stays recoverably Pending until evs-ssd/hcloud-volumes lands,
+              # then binds with zero re-creation. The substitute is the tofu
+              # `default_storage_class` var (evs-ssd on Huawei, hcloud-volumes
+              # on Hetzner) — validation rejects "" and the forbidden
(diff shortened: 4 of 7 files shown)


emrahbaysal:
## Live validation on omantel.biz (dep `4635277cae4ffed9`)

### #4206 — region-B PVCs: REPAIRED (Pending 46h → Bound)
```
BEFORE: guacamole-pg-1 / gitea-pg-1 / harbor-pg-1  Pending  storageClassName=""  46h
AFTER:  guacamole-pg-1 / gitea-pg-1 / harbor-pg-1  Bound    evs-ssd  (5Gi)
        guacamole-pg / gitea-pg / harbor-pg  CNPG "Cluster in healthy state" 1/1
        bp-gitea / bp-guacamole / bp-harbor HRs  True (cascade cleared; guacamole was stuck 'Running install' 2d)
```
Live procedure: added `SOVEREIGN_CNPG_STORAGE_CLASS=evs-ssd` to the live `bootstrap-kit-crs` substitute, suspended crs, patched the 3 Cluster CRs `storageClass=evs-ssd`, deleted the stuck initdb Jobs + reset `status.latestGeneratedNode=0` to un-wedge CNPG (it was "refusing to create the primary instance while the latest generated serial is not zero"), resumed crs. PVCs stay Bound (storageClassName is immutable once bound). The chart fix here makes this race-immune on every fresh prov.

### #3785 — demo Org replica-placement root: FIXED
The misplaced cross-region `wordpress-db-replica` (region-B node affinity in a region-A ns → pgbasebackup Pending forever → HR Ready=False → DEGRADED banner) is gone — the demo Org overlay no longer carries `pg.activeHotStandby` (count 0; Continuum CR removed). The org overlay already defaults hot-standby OFF on main.

### Deeper systemic root discovered + fixed
The org-overlay corruption that had already pruned the demo Org namespace traced to `ReadFile` returning the raw Gitea contents-API JSON envelope verbatim (this PR's `client.go` fix). Recovered live: restored a clean raw-YAML parent kustomization, cleared 3 stuck `application.apps.openova.io` finalizers, Flux recreated the demo Org ns + single-region WordPress. Data PVs survived (Retain).

### Remaining (separate, pre-existing — NOT this PR's D31 root)
The demo `bp-wordpress-tenant` HR is now blocked only by `dependsOn: bp-keycloak`, which is blocked by the env-wide `harbor-proxy-pull` Kyverno deny on the `bp-keycloak@1.5.0` initContainer image (the #3785 image-proxy class, for keycloak), and `bp-cnpg` by the `flux-managed` policy on its webhook-gate Job. These need their own tickets.

CI: all checks green (incl. bootstrap-kit drift, cloud-init render huawei+hetzner, validate+fmt+test, manifest-validation). Not merging — reporting per the task.

Issue #4075, opened 2026-06-22
Customer-Org console unreachable → agentic journey blocked at 'Enter org' (stale DNS + wrong Enter-org host + cert coverage)


emrahbaysal:
## Symptom (live, omantel.biz Sovereign dep 4635277cae4ffed9)

Created customer Org **demo** (free-subdomain on `omani.homes`) via the console → Org is **Active** with real HRs (region-A 67→74). But entering it fails:

1. **'Enter org' opens the wrong host.** The button (`/organizations/demo` → enter-org) opens a new tab to **`console.demo.omantel.biz`** (the Sovereign FQDN), NOT the Org's actual console URL shown on its own detail page: **`console.demo.omani.homes`** (the pool-parent free-subdomain). FE URL-construction bug — it should use the Org's `consoleURL`, not `console.<slug>.<sovereignFQDN>`.

2. **DNS is stale/wrong for BOTH hosts.** `console.demo.omantel.biz` AND `console.demo.omani.homes` both resolve to **49.12.16.160** — a **Hetzner IP from a prior env** that used the omani.homes pool. The funnel's DNS step did not write the correct omantel.biz Huawei record (console-ELB 212.72.24.33). So even the correct URL is poisoned by a stale record; the org-create DNS step must overwrite pool-subdomain records to the current Sovereign's ingress IP.

3. **Cert coverage gap.** The Sovereign wildcard is `*.omantel.biz` (SAN: `*.omantel.biz, omantel.biz`) — it does NOT cover the 2-level `console.demo.omantel.biz`. The Org's own console should be served under its pool parent (`*.demo.omani.homes` or the org cert), not a 2-level Sovereign subdomain.

`curl --resolve console.demo.omantel.biz:443:212.72.24.33` → HTTP 000 (not routed/served).

## Impact
Blocks the **agentic self-service journey** (demo@openova.io / emrah.baysal@openova.io) — a customer cannot reach their own console to install bp-agenity, so the agenity→openova-MCP→WordPress flow can't start as the customer.

## Asks
- [ ] FE: 'Enter org' must navigate to the Org's stored consoleURL (`console.<slug>.<orgParent>`), not `console.<slug>.<sovereignFQDN>`.
- [ ] Provisioner: the org-create DNS step must overwrite the pool-subdomain A/AAAA record to the current Sovereign's ingress (console-ELB) IP, defeating stale prior-env records on shared pool domains (omani.homes/rest/trade).
[...]

PR #4222, opened 2026-06-24
fix(catalyst-api): org-tenant pool A-record writes target CENTRAL pdns.openova.io (#4218 — final #4179 funnel layer)

emrahbaysal:
## What & why

The FINAL layer of the #1 customer bug (#4179). A fresh customer signup on a Sovereign creates the Org with the correct redirect host `console.<slug>.<omani-pool>`, but that host is **NXDOMAIN** because the org-tenant DNS provisioner never writes the pool A-record. #4216 fixed the env NAME; this fixes the two remaining root causes (no workaround, no manual record writes):

- **Layer A** — the central PowerDNS key was absent from `catalyst-system` → `NewPowerDNSWriter(url, "")` returned `nil` → no-op provisioner.
- **Layer B** — even with a key, the single writer pointed at the **Sovereign-LOCAL** PowerDNS (`powerdns.powerdns.svc`, authoritative only for the Sovereign's own FQDN), which 404s a PATCH against an `omani.works` zone it does not host. The record belongs to the **shared `omani.*` pool**, authoritative on the **central** `pdns.openova.io`.

## Changes

**Code** (`organization_dns.go`, `cmd/api/main.go`)
- `DefaultOrganizationDNSProvisioner` gains a dedicated `PoolWriter` (central pdns) that is **preferred** for `ProvisionFreeSubdomain`; the local `Writer` stays for the #827/#829 multi-zone admin flow.
- `main.go` wires the pool writer from a NEW env pair `CATALYST_POOL_POWERDNS_API_URL` / `CATALYST_POOL_POWERDNS_API_KEY`, **independent** of the local `CATALYST_POWERDNS_API_*` (local URL never repointed → `omantel.biz` writes untouched). No pool env (single-PowerDNS Sovereign where the pool IS the local zone) → automatic fallback to the local writer.

**Deploy wiring** (`chart/templates/api-deployment.yaml`, `values.yaml`)
- catalyst-api Deployment renders `CATALYST_POOL_POWERDNS_API_URL` (literal `https://pdns.openova.io`, dual-mode-safe) + `CATALYST_POOL_POWERDNS_API_KEY` via `optional` secretKeyRef to `pool-powerdns-api-credentials`.
- The kustomize (contabo Catalyst-Zero) twin is **intentionally untouched** — no `omani.*` pool funnel runs there.

**Secret availability** (`pool-powerdns-credentials-host-bridge.yaml`, `clusters/_template/sovereign-tls/powerdns-api-credentials.yaml`)
- A reflector **PULL-stub** renames + mirrors the CENTRAL `cert-manager/powerdns-api-credentials` key into `catalyst-system` as `pool-powerdns-api-credentials`; the source gains the matching `catalyst-system` reflection-allowed opt-in.
- The stub seeds **NO** `api-key` value (avoids the reflector empty-seed trap) and the consumer secretKeyRef is `optional` so the Pod schedules pre-reflection and self-heals on reconcile.

## Seam choice — inline provisioner, not `core/pool-domain-manager`

The org-tenant pipeline already owns the `ProvisionFreeSubdomain` step and atomically writes all per-Org host records (console, wordpress, keycloak, …) at the right pipeline state. Routing through the separate `pool-domain-manager` reconciler would add a cross-service hop + new failure modes for no gain. The minimal correct fix is to give the inline provisioner the right writer for pool domains.

## Tests
- `TestProvisionFreeSubdomain_PrefersPoolWriter` — pool write goes to the central server, **zero** leak to the local writer.
- `TestProvisionFreeSubdomain_FallsBackToLocalWriter` — single-PowerDNS back-compat.
- Existing `TestProvisionFreeSubdomain_WritesWildcardREPLACE` (#4075) still green.
- `go build ./...`, full `internal/handler` test pkg, `helm template`, `kubectl kustomize build` all pass.

## Verification status
Code + chart + secret wiring landed and unit-green. **Not yet live-verified** — the roll requires the deploy-bot to rebuild the catalyst-api image + bump the bp-catalyst-platform pin, then Flux reconcile on the Sovereign (cold ghcr pull ~10min). The live signed-in funnel walk on `console.<slug>.omani.works` is the #4179 close gate and follows the roll.

Refs #4218 #4179

### DIFF
--- a/clusters/_template/sovereign-tls/powerdns-api-credentials.yaml
+++ b/clusters/_template/sovereign-tls/powerdns-api-credentials.yaml
@@ -45,6 +45,21 @@ metadata:
   labels:
     catalyst.openova.io/sovereign: ${SOVEREIGN_FQDN}
     catalyst.openova.io/component: cert-manager-powerdns-webhook
+  annotations:
+    # #4218 — source-side opt-in so bp-reflector may PULL this CENTRAL
+    # pdns.openova.io key into `catalyst-system`, where catalyst-api's
+    # org-tenant DNS provisioner needs it to write the per-Org pool
+    # subdomain A-record (console.<slug>.<omani-pool>) on the central
+    # authoritative PowerDNS. The destination stub
+    # (products/catalyst/chart/templates/pool-powerdns-credentials-host-bridge.yaml)
+    # carries the matching `reflector…/reflects: cert-manager/powerdns-api-credentials`
+    # annotation, which renames it to `pool-powerdns-api-credentials` in
+    # catalyst-system. Without this opt-in the org-tenant write stays no-op
+    # (#4218 Layer A — empty key in catalyst-system → NewPowerDNSWriter nil).
+    # This is the SAME key bp-cert-manager-powerdns-webhook already uses for
+    # DNS-01 against pdns.openova.io — NOT a new credential.
+    reflector.v1.k8s.emberstack.com/reflection-allowed: "true"
+    reflector.v1.k8s.emberstack.com/reflection-allowed-namespaces: "catalyst-system"
 type: Opaque
 data:
   api-key: ${POWERDNS_API_KEY}
--- a/products/catalyst/bootstrap/api/cmd/api/main.go
+++ b/products/catalyst/bootstrap/api/cmd/api/main.go
@@ -983,9 +983,36 @@ func main() {
 			// post-signup redirect (the #4179 funnel landing).
 			pdnsURL := os.Getenv("CATALYST_POWERDNS_API_URL")
 			pdnsKey := os.Getenv("CATALYST_POWERDNS_API_KEY")
-			if writer := handler.NewPowerDNSWriter(pdnsURL, pdnsKey); writer != nil {
-				tdeps.DNS = handler.DefaultOrganizationDNSProvisioner{Writer: writer}
-				log.Info("org-tenant: powerdns writer wired", "url", pdnsURL)
+			localWriter := handler.NewPowerDNSWriter(pdnsURL, pdnsKey)
+			// #4218: pool-domain (omani.*) console A-records must be
+			// written to the CENTRAL authoritative PowerDNS (pdns.openova.io)
+			// with the CENTRAL pool key — NOT the Sovereign-local PowerDNS,
+			// which hosts only the Sovereign's own FQDN zone and has no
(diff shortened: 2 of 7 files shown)


emrahbaysal:
## CI status — code gates green; 3 reds are pre-existing base-branch drift (not this PR)

**Green (my code/chart gates):** `test` ✅ · `integration (kind + gitea)` ✅ · banned-words (body+commits) ✅ · Closes/Fixes reject ✅ · hetzner + huawei cloud-init render ✅ · manifest-validation ✅ · naming-guard ✅ · bootstrap-kit drift ✅.

**Red — independently confirmed pre-existing on `origin/main`, NOT caused by this changeset:**

1. **`strategy-flip-regression`** — the test `kubectl apply --dry-run`s the RAW `api-deployment.yaml`. It fails at **line 166-167** (`{{- with .Values.catalystApi.nodeName }}` / `nodeName: {{ . | quote }}`), introduced by commit `b4dd43d028` (e3mrah, 2026-06-22) — two days before this branch. Independently reproduced: `python3 -c "yaml.safe_load_all(open(BASE_api-deployment.yaml))"` on the **base** `origin/main` file raises the identical `could not find expected ':' line 167`. My additions are at line ~696, far below the break.
2. **`pin-sync-audit`** — reports `chart=1.4.810 pin=1.4.809`. My diff to `Chart.yaml` is **empty**; the drift is the deploy-bot chart-version-vs-pin lockstep on the moving main, resolved by the #2851 deploy-bot sweep on merge.
3. **`Controller-image-tag freshness`** — stale `controllers.*.image.tag` pins in `values.yaml`. My `values.yaml` diff adds only a `poolPowerdnsBridge` block — **no** controller image tags. Resolved by the deploy-bot sweep.

## Live architecture validation (on the omantel.biz Sovereign, via mothership pod)
- `cert-manager/powerdns-api-credentials` present with a **48-char** key (the CENTRAL pool key) — the reflector source this PR pulls from. ✅
- bp-reflector running (`reflector/reflector` 1/1) — will honor the new source-side `reflection-allowed-namespaces: catalyst-system` opt-in + the destination `reflects` stub. ✅

## Not merged
Per the ticket: not merging without the live auto-create proof. The roll requires the deploy-bot to rebuild catalyst-api + bump the bp-catalyst-platform pin, then Flux reconcile (cold ghcr pull ~10min). The signed-in funnel walk on `console.<slug>.omani.works` (the #4179 close gate) follows the roll.

PR #4239, opened 2026-06-24
fix(bp-agenity): per-Org overlay generator installs the dashboard zero-touch — 4 durable bugs (Refs #4180)

emrahbaysal:
## Summary

Fixes **four durable catalyst-api per-Org overlay-generator bugs** that blocked the bp-agenity dashboard (#4180) from installing on a fresh prov. All four were diagnosed live on the **omantel.biz demo Org (2026-06-24, dep `4635277cae4ffed9`)** and were previously only hot-fixed in Gitea (throwaway). This PR makes the CODE reproduce a working install with **zero hand-patches** on a fresh prov / overlay-regen.

Primary file: `products/catalyst/bootstrap/api/internal/handler/organization_gitops.go`. New regression test: `products/catalyst/bootstrap/api/internal/handler/organization_agenity_overlay_test.go`.

### BUG 1 — missing bp-agenity HelmRepository
`orgTenantSharedHelmRepositories` declared HelmRepositories for every bp-* chart **except** bp-agenity, so the rendered HelmRelease's `sourceRef: HelmRepository/bp-agenity@flux-system` had no source → HelmChart `HelmRepository "bp-agenity" not found`.
**Fix:** appended a `bp-agenity` HelmRepository block (oci / interval 15m / `oci://ghcr.io/openova-io` / secretRef `ghcr-pull`) — `organization_gitops.go` ~L367-380 (just after the `bp-stalwart-tenant` block).

### BUG 2 — version `"*"` resolved to the IMAGE tag, not the chart
The `ghcr.io/openova-io/bp-agenity` OCI repo co-hosts CHART tags (0.5.3/0.5.4/0.5.5) **and** IMAGE tags (0.9.4/0.9.5/0.9.6) in the same package. Flux's OCI `*` resolves to the highest semver = 0.9.6 (the image, not a Helm chart) → `manifest does not contain minimum number of descriptors (2), descriptors found: 1`. Unique to bp-agenity.
**Fix:** `withVersionDefaults` now defaults `Agenity` to the chart-range `"0.5.x"` via the new `agenityDefaultChartRange` const (NOT `star()`); an explicit `CATALYST_ORG_BP_AGENITY_VER` still wins. Every other chart keeps `"*"`. `organization_gitops.go` ~L686-714 (`agenityDefaultChartRange` + `withVersionDefaults`). Verified live: `version: "0.5.x"` resolves to chart 0.5.5.

### BUG 3 — bp-agenity overlay lacked imagePullSecrets
`ghcr.io/openova-io/bp-agenity:0.9.x` is a **private** ghcr package; the Org ns has no default ghcr auth → 401 anonymous-token / ImagePullBackOff.
**Fix:** added `imagePullSecrets: [{name: ghcr-pull}]` to the rendered `orgTenantBPAgenity` values — `organization_gitops.go` ~L1696-1711.
**Secret availability:** the Org ns already HAS `ghcr-pull`: bp-reflector at bootstrap-kit slot 05a (`clusters/_template/bootstrap-kit/05a-reflector.yaml`) mirrors `flux-system/ghcr-pull` (carries `reflection-auto-enabled: "true"` + `reflection-auto-namespaces: ""` → all-namespaces) into **every** namespace. No extra seed needed — same mechanism `platform/guacamole` relies on (Chart.yaml 0.1.3: default `imagePullSecrets: [{name: ghcr-pull}]`). The chart consumes top-level `imagePullSecrets` (`products/agenity/chart/templates/statefulset.yaml:52`, default `[]`).

### BUG 4 (most important — cross-Org data loss) — parent kustomization.yaml as a JSON envelope
After an overlay regen the parent index `clusters/<otech-fqdn>/org-tenants/kustomization.yaml` could end up containing a Gitea contents-API JSON envelope (`{"name":...,"content":"<base64>",...}`) instead of raw kustomize YAML. kustomize then parses an **empty** resource set → the `org-tenants` Flux Kustomization (`prune=true`) PRUNES every Org's bp-* HelmReleases cluster-wide.

**Corruption entry point found:** NOT this bootstrap-api path. `WriteTenantOverlay` → `writeParentTenantsIndex` (`organization_gitops.go` L89, L206) uses pure `git clone` + `os.ReadDir` + `os.WriteFile`, regenerating the index from scratch — structurally immune to the envelope. The actual writer that corrupts this **shared** file is the OTHER provisioning path: `core/services/provisioning/github/client.go::ReadFile`, which returned the Gitea `/contents/` JSON envelope verbatim; the rebuild closure spliced the new Org subdir onto it and committed the JSON-wrapped string back. **That root cause is already fixed on main** in `f89563b8a` (#4214, Refs #4206 #3785) — see the `ReadFile` decode at `client.go:943-968`. The prior manual Gitea fix (`939b19f2`) was re-corrupted by the next provision (`2739582d`) precisely because that ReadFile bug was still live at the time.

**This PR adds a defensive regression guard** at the bootstrap-api write boundary (since the index is a file two code paths commit to): `writeParentTenantsIndex` now calls the new `validateRawKustomizationYAML` and refuses to write an index that doesn't start with a `#` comment or `apiVersion: kustomize` (rejects a leading `{` or `[`). `organization_gitops.go` ~L269-330. Any future regression in either writer is caught before the corrupt bytes reach git.

## Tests
`organization_agenity_overlay_test.go` — 6 tests, all green:
- BUG 1: bp-agenity HelmRepository declared with correct oci/url/secret shape.
- BUG 2: empty default = `0.5.x` (other charts still `*`); explicit value wins; rendered HR version = `0.5.x`.
- BUG 3: rendered values `imagePullSecrets[0].name == ghcr-pull`.
- BUG 4: `writeParentTenantsIndex` output is raw YAML + parses to a non-empty kustomization; `validateRawKustomizationYAML` rejects the JSON-envelope shape and accepts raw YAML.

`go build ./...` and `go test ./...` green for `products/catalyst/bootstrap/api` (handler package incl. new tests).

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/organization_agenity_overlay_test.go
+++ b/products/catalyst/bootstrap/api/internal/handler/organization_agenity_overlay_test.go
@@ -1,0 +1,230 @@
+// organization_agenity_overlay_test.go — #4180.
+//
+// Regression coverage for the FOUR durable bp-agenity overlay-generator
+// bugs diagnosed live on the omantel.biz demo Org (2026-06-24, dep
+// 4635277cae4ffed9) and previously only hot-fixed in Gitea:
+//
+//	BUG 1 — missing bp-agenity HelmRepository (sourceRef had no source).
+//	BUG 2 — Agenity version "*" resolved to the IMAGE tag (0.9.6), not the
+//	        chart; default must be the chart-range "0.5.x".
+//	BUG 3 — bp-agenity overlay lacked imagePullSecrets → ImagePullBackOff
+//	        on the private ghcr image in the Org ns.
+//	BUG 4 — the parent org-tenants/kustomization.yaml could be committed as
+//	        a Gitea contents-API JSON envelope, pruning every Org's bp-*
+//	        HelmReleases cluster-wide. Guard: never write a '{'-leading
+//	        parent index.
+package handler
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/openova-io/openova/products/catalyst/bootstrap/api/internal/store"
+)
+
+func renderAgenityOverlay(t *testing.T, versions OrganizationChartVersions) map[string]string {
+	t.Helper()
+	rec := store.OrganizationProvisionRecord{
+		OrganizationID:  "t-acme",
+		Subdomain:       "acme",
+		DomainMode:      store.OrganizationDomainFreeSubdomain,
+		AdminEmail:      "admin@acme.test",
+		CompanyName:     "Acme Corp",
+		OTECHFQDN:       "otech.example",
+		ParentDomain:    "otech.example",
+		VClusterName:    "vc-acme",
+		TenantNamespace: "org-t-acme",
+	}
+	files, err := renderOrganizationOverlay(rec, versions)
+	if err != nil {
+		t.Fatalf("render: %v", err)
+	}
+	return files
+}
+
+// BUG 1 — the shared HelmRepositories file MUST declare bp-agenity, else
+// the rendered HelmRelease's sourceRef has no source.
+func TestBug1_AgenityHelmRepositoryDeclared(t *testing.T) {
+	if !strings.Contains(orgTenantSharedHelmRepositories, "name: bp-agenity") {
+		t.Fatalf("orgTenantSharedHelmRepositories missing bp-agenity HelmRepository block")
+	}
+	// Sanity: every HelmRepository the per-tenant overlay sourceRefs into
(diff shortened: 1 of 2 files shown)


emrahbaysal:
Superseded by the merged chart 0.5.7 / appVersion 0.9.7 train for #4180 (#4225 #4173 #4259 #4233 #4261 + catalog resync #4264). The live demo Org now pins chart bp-agenity@0.5.7 via the HR (no image.tag override, no hand-patch) and /app/ serves the workspaces dashboard (Dashboardws.Tnm8UdB3.js) with cache-control: no-cache from the chart — the durable target this PR aimed at, delivered via the chart-pin mechanism (ticket section 5) rather than an HR-overlay/patch. Closing as obsolete; live evidence on #4180.

PR #4430, opened 2026-06-26
fix(bp-crossplane): crossplane-core OOMKilled at 2Gi → raise limit to 4Gi (1.1.5→1.1.6)

emrahbaysal:
## What

crossplane-core on omantel.biz region-a (dep `4635277cae4ffed9`) was `CrashLoopBackOff` with **55 restarts** — OOMKilled (exit 137), dying ~90s after each start — at the prior **2Gi** limit, even with **zero** `providers.pkg.crossplane.io` and **zero** CloudAdoptions installed. With core OOM-looping it registered no providers and no managed resources, so the Crossplane object-model seam (#4018) could not converge.

## Root cause

`core start` builds discovery + composition/GC informer caches across **every** CRD on the cluster (190 here, incl. 8 `compose.openova.io` XRDs Established). Live `kubectl top` on the recovered pod shows the steady-state working set is **~3.5Gi** — so the 1Gi (original) and 2Gi (first fix) limits were both far below it; the pod OOMed before it could ever reach the point of installing a provider.

## Fix

`resourcesCrossplane` → `limits.memory: 4Gi` / `requests.memory: 1Gi` (node carries ~64Gi allocatable; core is a cluster-singleton, ~500Mi headroom over the live 3.5Gi working set). RBAC-manager untouched.

Lockstep `1.1.5`→`1.1.6`: `platform/crossplane/chart/{values,Chart}.yaml` + `platform/crossplane/blueprint.yaml` + `clusters/_template/bootstrap-kit/04-crossplane.yaml` + the `bp-catalyst-platform` catalog-seed entry; umbrella `bp-catalyst-platform` `1.4.879`→`1.4.880` so the seed re-rolls. `helm lint` + `check-catalog-seed-lockstep.sh` both pass.

## Live evidence (omantel.biz region-a, dep `4635277cae4ffed9`)

- Before: pod `crossplane-fbf546bd6-q57mh` — 55 restarts, `lastState.terminated.reason=OOMKilled` exit 137 at the 2Gi limit.
- Memory-limit patch applied live to the idle crashlooping deployment (sanctioned in-place fix; no catalyst-api restart, no serving-HR force-reconcile).
- After: new pod `crossplane-77f6986674-5ddhm` — **Running 1/1, restartCount=0, lastState={}**, stable across **24 consecutive checks over ~6m42s** (vs ~90s death point before). `kubectl top` = **3534Mi** under the 4Gi limit. Deploy `available=1 ready=1 updated=1`.
- Provider acceptance probe: applied a `Provider` CR → it was **accepted and listed** by `providers.pkg.crossplane.io` (previously empty — the package manager couldn't run while core was OOM-looping) and core stayed Running with 0 restarts under the load. Probe removed afterward (no residue).

## Scope

This unblocks #4212 Seam 1+2 / #4018 (crossplane-core OOM gate). The FULL #4212 close still needs the fresh-prov CloudAdoptions walk — **#4212 stays `status/uat`**, not closed.

Refs #4212 #4018

### DIFF
--- a/clusters/_template/bootstrap-kit/04-crossplane.yaml
+++ b/clusters/_template/bootstrap-kit/04-crossplane.yaml
@@ -41,7 +41,7 @@ spec:
   chart:
     spec:
       chart: bp-crossplane
-      version: 1.1.5
+      version: 1.1.6
       sourceRef:
         kind: HelmRepository
         name: bp-crossplane
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1138,7 +1138,7 @@ spec:
       #   Blueprint CRs so a merged seed version bump reaches the LIVE CR on the
       #   next helm upgrade (no kubectl patch); catalyst-api strips version/source
       #   from the Flux aggregator so Helm stays their sole owner.
-      version: 1.4.879
+      version: 1.4.880
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/platform/crossplane/blueprint.yaml
+++ b/platform/crossplane/blueprint.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     catalyst.openova.io/section: pts-3-2-gitops-and-iac
 spec:
-  version: 1.1.5
+  version: 1.1.6
   card:
     title: crossplane
     summary: |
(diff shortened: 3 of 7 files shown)


emrahbaysal:
Follow-up commit `875086fcb`: synced bootstrap-kit slot-13 `bp-catalyst-platform` pin 1.4.879→1.4.880 (pin-sync-audit caught the umbrella Chart.yaml bump without the matching slot pin). Local `check-bootstrap-kit-pin-sync.sh --changed-only` now PASS for both pairs.

PR #4451, opened 2026-06-26
fix(bp-catalyst #4447): PAT-sync hooks order RBAC before the Job + fail-loud on forbidden read — fresh-prov catalog + Org-provisioning unblock (1.4.886)

emrahbaysal:
## Problem (Pillar-1 dead on a fresh prov)
On a fresh `omantel.biz` prov (`b9f9590b558262b6`, 2026-06-26) the catalog + per-Org app layer never converges, and the provisioning engine never starts:

```
flux-system/catalog-sovereign  False  Source artifact not found, retrying in 30s
flux-system/org-tenants        False  Source artifact not found, retrying in 30s
org-services/provisioning      Init:0/1  wait-for-gitea-token: GITHUB_TOKEN not yet populated
```

Root: three secrets that gate the whole funnel were **absent** —
`openova-catalog-sovereign-git-auth`, `openova-org-tenants-git-auth` (Flux source auth) and `org-services/provisioning-github-token` (provisioning engine). No catalog + no provisioning ⇒ a voucher redeem can never provision an Org with a running WordPress.

## Root cause (verified live, not guessed)
Both post-install PAT-sync hooks raced their OWN read-RBAC:
- `templates/catalog-sovereign-flux/gitea-flux-auth-secrets-sync-job.yaml`
- `templates/org-services/provisioning-github-token-sync-job.yaml`

The Job **and** its `pat-reader` Role/RoleBinding were all `hook-weight: "20"`. Helm applies same-weight hooks together, so the Job pod started polling `catalyst-gitea-token` BEFORE its RBAC bound. The kubectl read was a silent 403 (`2>/dev/null ... || true`) for the **full 300s poll budget**, then hit the FATAL branch which `exit 0`'d having created **zero** secrets. The HR then sat `Ready=True / InstallSucceeded` at revision **v1** and never re-runs a post-install hook → no self-heal.

Live helm-release hook `last_run` proves it: Job started 10:20:45, completed 10:26:04 `Succeeded`; its pat-reader Role/RoleBinding were created at **10:26:04** — i.e. RBAC bound ~5m19s after the Job began polling, exactly when the budget elapsed.

The #4286 Kyverno floor (`inject-gitea-source-secretref`) guarantees `secretRef.name` is non-empty but does NOT guarantee the named secret EXISTS — so the source pointed at a secret nobody created.

## Fix (both jobs)
1. **Order RBAC before the Job**: drop the ServiceAccount + all 4 RBAC objects to `hook-weight: "15"` (< the Job's `20`). Helm binds RBAC, then runs the Job, so the very first PAT read succeeds.
2. **Fail loud on a forbidden read**: capture the kubectl error; if the budget is exhausted on a *forbidden* read, `exit 1` (restartPolicy:OnFailure re-runs with RBAC bound). Only the genuine *never-minted-PAT* case keeps the benign `exit 0` (#3763/#3781) so a truly-absent mint doesn't wedge the catalyst-platform upgrade.

`bp-catalyst-platform` chart 1.4.882 → **1.4.886**.

## Live evidence (this prov)
The immediate unblock (creating all three secrets by hand from the PAT) flipped both GitRepositories **Ready=True**, both Kustomizations **Reconciled** within 12s, and the catalog `blueprints.catalyst.openova.io` CRD + plan catalog converged. This PR makes that automatic on every fresh prov.

CAVEAT: full zero-touch fresh-prov re-verification of the hook firing correctly is pending — the live prov's Huawei me-east-215 EIP pool went unreachable mid-session (both regions' apiserver/ELB EIPs down; bastion .20 unaffected), so the post-fix re-roll couldn't be walked. The chart renders clean (`helm template` rc=0) and the weight/exit logic is deterministic; re-stamp on the next reachable fresh prov.

Refs #4447

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1138,7 +1138,7 @@ spec:
       #   Blueprint CRs so a merged seed version bump reaches the LIVE CR on the
       #   next helm upgrade (no kubectl patch); catalyst-api strips version/source
       #   from the Flux aggregator so Helm stays their sole owner.
-      version: 1.4.885
+      version: 1.4.886
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/products/catalyst/chart/Chart.yaml
+++ b/products/catalyst/chart/Chart.yaml
@@ -1,5 +1,22 @@
 apiVersion: v2
 name: bp-catalyst-platform
+# 1.4.886 — #4447 (2026-06-26): the two PAT-sync post-install hooks
+# (gitea-flux-auth-secrets-sync-job + provisioning-github-token-sync-job)
+# raced their OWN read-RBAC. The Job and its pat-reader Role/RoleBinding were
+# all hook-weight 20, so the Job pod started polling catalyst-gitea-token
+# BEFORE the RBAC bound; the kubectl read was a silent 403 (`2>/dev/null||true`)
+# for the FULL 300s poll budget, then hit the FATAL branch which `exit 0`'d
+# having created ZERO secrets. The HR sat Ready=True/InstallSucceeded at
+# revision v1 and never re-ran the hook → openova-catalog-sovereign-git-auth +
+# openova-org-tenants-git-auth + provisioning-github-token all ABSENT →
+# catalog-sovereign/org-tenants Flux sources stuck "Source artifact not found"
+# AND the provisioning service hung Init:0/1 → no catalog, no Org provisioning
+# (Pillar-1 dead). FIX: drop the ServiceAccount + all 4 RBAC objects to
+# hook-weight 15 (< the Job's 20) so Helm binds RBAC before the Job polls;
+# capture the kubectl error and, if the budget is exhausted on a *forbidden*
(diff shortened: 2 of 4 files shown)


emrahbaysal:
Superseded — closing.

This PR fixed the hook RBAC-race / silent-exit-0 fault in **both** PAT-sync hook Jobs. Its two halves have now landed on `main` independently:

- **gitea-flux-auth-sync** half → merged via #4452 (commit `6f99289d7`, chart 1.4.886). RBAC ordered strictly before the Job (weight 15 < 20) + fail-loud on PAT read-error so the two Flux `*-git-auth` Secrets land zero-touch.
- **provisioning-github-token-sync** half (the second instance #4452 missed) → merged via #4456 (commit `5d8fb3cf8`, chart 1.4.889). Same fix mirrored exactly onto `products/catalyst/chart/templates/org-services/provisioning-github-token-sync-job.yaml`: SA + all four RBAC objects to weight 15, `auth can-i` preflight + exit-code-captured PAT read (no more `2>/dev/null||true` swallow), fail-loud on read-error with budget exhausted, exit-0 only on a clean-empty mint-race (preserving #3763/#3781), and a post-apply verify-back of `GITHUB_TOKEN`. The #4452 guard test was extended with parallel gate-3/gate-4 cases for this second job (negative control proven against the unfixed job).

Both jobs are now fixed on `main`. Nothing left in this PR to merge. Refs #4447.

PR #4492, opened 2026-06-26
fix(catalyst): seed newapi admin-token into OpenBao so catalyst-newapi-admin-token ExternalSecret resolves on fresh prov

emrahbaysal:
Refs #4477 #3374

## Problem
The `catalyst-newapi-admin-token` ExternalSecret (ADR-0003 §3.2/§6 — the unified-RBAC admin-API bearer for per-user-key issuance) sat `SecretSyncedError` on every fresh prov because the OpenBao path it reads — `sovereign/<fqdn>/newapi/admin-token` — was **never written by any producer**. No code seeded it. Result: unified-RBAC's calls against NewAPI's admin API 401'd. Live-observed on `91dc05917e44d1c1` (omantel.biz); recorded as the **secondary** item in #4477 (the primary SSO FAIL 1 / FAIL 2 ship in a separate PR).

## Fix
Add a `seedNewapiAdminToken` seam in catalyst-api (`products/catalyst/bootstrap/api/internal/handler/sovereign_newapi_admin_token_seed.go`), mirroring the existing `seedAnthropicToken` / `seedMCPBearer` producers:

- It reads NewAPI's **bridge `ADMIN_SECRET`** — the `newapi-bp-newapi-token-signing-key` Secret key the sandbox-bridge validates with `subtle.ConstantTimeCompare` — and propagates it **verbatim** into the Sovereign's OpenBao. This guarantees the distributed token MATCHES what the bridge accepts. A freshly-minted random token would 401 on every unified-RBAC call.
- The producer can only write under `secret/{data,metadata}/catalyst/*` (its `catalyst-api-write` OpenBao role; a write to `secret/sovereign/*` 403s), so the canonical read-path moves to the cluster-shared **`catalyst/newapi/admin-token`** in lock-step — the identical resolution the anthropic / mcp-bearer seeds use.
- Wired into `runOrganizationPipeline` beside the other two seeds: idempotent (`PutKVv2` overwrites), cluster-shared (the first Org-create converges the whole Sovereign), and **never fails the pipeline** (a missing bridge Secret only leaves the admin path unauthenticated until NewAPI's token-signing-key Secret renders; loud-skip, no empty-seed).

## Read-path lockstep
- `clusters/_template/bootstrap-kit/80-newapi.yaml` — `catalystIntegration.externalSecret.remoteRef.key` → `catalyst/newapi/admin-token` + chart pin 1.4.135→1.4.136.
- `platform/newapi/chart/{Chart.yaml,values.yaml,templates/external-secret.yaml}` + `blueprint.yaml` — version + default-doc to the `catalyst/` path.

## Verification
- `go build ./...` + `go vet ./internal/handler/` clean.
- 5 new seed tests + full handler suite green.
- `helm template` renders the ExternalSecret reading **exactly** the seeded path (`key: catalyst/newapi/admin-token`, `property`/`secretKey: ADMIN_API_TOKEN`).
- Shipped `admin-promote-rbac-render.sh` (incl. Case 6 #4477) + `bridge-readiness-render.sh` pass.

Live verification (the ExternalSecret syncs on a prov) is blocked by the current mothership outage — this ships the durable merge; the sync verifies on the next reachable prov.

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1188,7 +1188,10 @@ spec:
       #   sandbox-controller can read CATALYST_GITEA_TOKEN (was CrashLoopBackOff
       #   on every Sovereign → 0 Sandboxes → no openova-sandbox-mcp). Lockstep
       #   w/ Chart.yaml + templates/catalyst-gitea-token-secret.yaml.
-      version: 1.4.905
+      # 1.4.906 — #4477 secondary: doc-only — CATALYST_NEWAPI_ADMIN_TOKEN env
+      #   comment cites the canonical `catalyst/newapi/admin-token` OpenBao path
+      #   (written by catalyst-api's seedNewapiAdminToken seam).
+      version: 1.4.906
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/clusters/_template/bootstrap-kit/80-newapi.yaml
+++ b/clusters/_template/bootstrap-kit/80-newapi.yaml
@@ -301,7 +301,11 @@ spec:
       # 1.4.134 — #4477: admin-promote CronJob self-heals a stale OIDC provider
       #   client_secret (resync_provider_secret) so a KC client-secret rotation
       #   no longer wedges newapi token-exchange ("Failed to get token").
-      version: 1.4.135
+      # 1.4.136 — #4477 secondary: catalyst-newapi-admin-token ExternalSecret
+      #   remoteRef.key → catalyst/newapi/admin-token (the path catalyst-api's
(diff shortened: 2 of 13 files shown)


emrahbaysal:
Amended (`384425454`): the seam now reads the bridge `ADMIN_SECRET` from the **`catalyst-system` reflector mirror** rather than the chart's origin namespace. The bp-newapi `sandboxTokenSigningKey.reflectorNamespaces` default (`catalyst-system,sandbox,sandbox-.*`) mirrors the token-signing-key Secret into `catalyst-system` on EVERY placement role (host `all` and `vcluster-app`) — the same namespace catalyst-api runs in — so the read is placement-role-agnostic and needs no cross-namespace RBAC. The `CATALYST_NEWAPI_ADMIN_TOKEN_SECRET_{NAMESPACE,NAME,KEY}` env knobs remain for overlays that re-home it. Build/vet clean, seed tests + handler suite green.

PR #4670, opened 2026-06-30
fix(wipe huawei): sweep UNBOUND/nameless EIPs in verifyZeroOrphans — stop per-wipe orphan EIP leak (#4636)

emrahbaysal:
## Problem

Found live wiping dep `8fd457a8` / omantel.biz on kom4dc. The canonical wipe (`WipeDeployment` → `providers/huawei/provider.go`) enumerates EIPs to purge **by name-prefix** (`listEIPsByNamePrefix`, keyed on the SovereignFQDN `bandwidth_name`), and `verifyZeroOrphans` checks the same prefix-matched set. An **UNBOUND, DOWN EIP `212.72.24.36`** — no `port_id`, no `catalyst-<stem>-` name match — was invisible to both. The wipe reported `verifiedZeroOrphans:true` while the EIP stayed in the project; an independent `GET /v1/{proj}/publicips` sweep caught it and it was released by hand.

Orphan EIPs accumulate per wipe → Huawei `publicIp` quota exhaustion → the same quota-leak class behind the #4614 production reap. The wipe's own "verified zero orphans" was therefore not trustworthy.

## Fix

New helper `listUnboundOrphanEIPs` (`wipe_unbound_eip.go`) does a project-wide `GET /v1/{proj}/publicips` — the same raw listing `SweepOrphanEIPs` / `allocateCleanEIP` already use — and returns every EIP that is **NON-bastion AND UNBOUND (`port_id == ""`)**. It reuses `doSignedRequest`, `endpointFor`, the existing `hwEIP` projection, and `snippet` — no new signing path.

- **Purge path** (`provider.go`, step 4, after the name-prefix delete loop): release every non-bastion unbound EIP project-wide. Conservative — only unbound EIPs are deleted, never a bound one.
- **`verifyZeroOrphans`**: record any surviving non-bastion unbound EIP in `ResidualOrphans["floating_ips"]` so the existing G103 wipe-contract gate **fails** instead of falsely passing. Name-prefix + unbound matches are deduped.

## Bastion protection (HARD)

The bastion EIP `212.72.24.20` is **bound** to the `bastion-*` ECS, so the unbound-only filter (`port_id == ""`) already excludes it. As defence-in-depth `isProtectedBastionEIP` ALSO skips the literal bastion IP and any `bandwidth_name` carrying the `bastion` substring — mirroring every other bastion guard in this package. The bastion is never released and never flagged.

**Bastion identification — honest note:** the guard is primarily **port-binding** (a bound EIP is never released, which is what actually protects the live bastion) plus a literal-IP and name-substring belt-and-braces. I did not resolve the `port_id` back to the owning ECS name — the unbound-only rule makes that unnecessary for the release path (we only ever release EIPs with no binding at all), and the literal-IP guard covers the bastion's specific address.

## Test

`wipe_unbound_eip_test.go` stands up a fake `/publicips` listing with (a) the bastion EIP (bound, sacred IP), (b) a bound catalyst EIP, (c) an unbound nameless orphan (`212.72.24.36` shape):
- `listUnboundOrphanEIPs` returns **only** the orphan.
- `verifyZeroOrphans` lands the orphan in `ResidualOrphans["floating_ips"]`, flips `VerifiedZeroOrphans=false`, and **never** records the bastion.
- a bastion-only project still verifies zero orphans (no false positive).

Verified the `verifyZeroOrphans` test **FAILS** without the sweep and **PASSES** with it (real regression guard, not token-passing).

`go build ./...`, `go vet ./internal/providers/huawei/`, and the full huawei package tests are clean.

## Audit (the issue's "same gap elsewhere?")

The same **name-prefix-only** assumption applies to the **keypair** and **security-group** residual scans in `verifyZeroOrphans` (filter on `catalyst-<stem>-` / `listSGsByName`). A genuinely nameless keypair/SG would be missed too. EIP is the P0 here — tight 10/project quota and the actual live leak — so per the issue's guidance those are noted for a follow-up rather than expanded in this change.

Refs #4636 #4614 #4454

### DIFF
--- a/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
+++ b/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
@@ -715,11 +715,11 @@ spec:
       # event-driven, not render-time-bound. NOT a Helm hook (converges async
       # behind disableWait:true). No step-count change; the Job carries
       # component=gitea-admin-secret-bridge, not the cutover-step label.
-      # 0.1.97 (#4637): step-04 registry-pivot gates its per-node v2 ack on a
+      # 0.1.98 (#4637): step-04 registry-pivot gates its per-node v2 ack on a
       # node-local `GET http://127.0.0.1:4001/v2/` serving probe so catalyst-api's
       # daemonset-wait cannot green (→ step-07 catalyst-api re-pull) while the
       # dfdaemon mirror has 0 Ready pods (#4649 condition → connection-refused).
-      version: 0.1.97
+      version: 0.1.98
       sourceRef:
         kind: HelmRepository
         name: bp-self-sovereign-cutover
--- a/platform/self-sovereign-cutover/blueprint.yaml
+++ b/platform/self-sovereign-cutover/blueprint.yaml
@@ -6,7 +6,7 @@ metadata:
     catalyst.openova.io/section: pts-2-3-per-sovereign-supporting-services
     catalyst.openova.io/tier: catalyst-cp
 spec:
-  version: 0.1.97  # lockstep with chart/Chart.yaml (#4637 step-07 dfdaemon-serving v2-ack gate)
+  version: 0.1.98  # lockstep with chart/Chart.yaml (#4637 step-07 dfdaemon-serving v2-ack gate)
   card:
     title: self-sovereignty-cutover
     summary: |
--- a/platform/self-sovereign-cutover/chart/Chart.yaml
+++ b/platform/self-sovereign-cutover/chart/Chart.yaml
@@ -1481,7 +1481,7 @@ name: bp-self-sovereign-cutover
 # it back to true. Live 11-step→cutoverComplete=true validation is DEFERRED to a
 # fresh prov whose cutover reaches the 600s egress hold (roll-gated; not run on
 # the permanent env).
-version: 0.1.97
+version: 0.1.98
 description: |
   Catalyst Self-Sovereignty Cutover Blueprint. Installs DORMANT — this
   chart ships eight step ConfigMaps (PodSpec ConfigMaps, one per step),
--- a/platform/self-sovereign-cutover/chart/templates/04-registry-pivot-daemonset.yaml
+++ b/platform/self-sovereign-cutover/chart/templates/04-registry-pivot-daemonset.yaml
@@ -194,6 +194,16 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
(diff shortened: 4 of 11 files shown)


emrahbaysal:
Heads-up: this PR's stated wipe fix (sweep UNBOUND/nameless orphan EIPs in `verifyZeroOrphans`) is **already merged via #4671** and live in `main` (`provider.go:542-547`). This branch's diff, however, touches **cutover** files (`06a-bp-self-sovereign-cutover.yaml`, `04-registry-pivot-daemonset.yaml`, `cutover_reconciler.go`), not the wipe path — the diff appears stale/mismatched (mergeable UNKNOWN). Recommend: close as superseded for the wipe fix; if the cutover-file changes are still wanted, re-file them as a clean, correctly-scoped PR. Not closing unilaterally since the cutover diff may carry intent I can't fully assess.

emrahbaysal:
Closing as superseded — the identical mechanism merged to main via PR #4671 (commit 0dc606d6e, `fix(wipe): sweep UNBOUND/nameless orphan EIPs project-wide — verifyZeroOrphans + purge missed them (#4636)`).

Verified against current origin/main (1846b0853, 2026-07-19):

- `products/catalyst/bootstrap/api/internal/providers/huawei/wipe_unbound_eip.go` and `wipe_unbound_eip_test.go` on main are **byte-identical** to this branch's files (`diff` of both blobs is empty).
- Purge half: `internal/providers/huawei/provider.go:1026` — `listUnboundOrphanEIPs` + `deleteEIP` loop releases every NON-bastion UNBOUND (`port_id == ""`) EIP project-wide after the name-prefix delete pass, recording each into `ProviderPurge["floating_ips"]`.
- Verify half: `internal/providers/huawei/provider.go:549` — survivors land in `ResidualOrphans["floating_ips"]` so the G103 zero-orphans verdict fails instead of falsely passing.
- Bastion guard (bound-only filter + literal `212.72.24.20` + `bastion` name substring) is in the merged `wipe_unbound_eip.go` exactly as authored here.

The other commits stacked on this branch also landed via their own PRs: the level-triggered cutover reconciler via #4669 (e76c7e9d1, plus the operatorRetry follow-up #4673 / ed3b9aaec) and the dfdaemon HOST_IP probe (the `HOST_IP` downward-API env is live in `platform/self-sovereign-cutover/chart/templates/04-registry-pivot-daemonset.yaml:167`; that template has since evolved far past this branch). The chart-pin bumps here (cutover 0.1.99) are long superseded.

Issue #4636 is already CLOSED with a verify-pass comment citing the same file:line evidence. Nothing on this branch is left to rebase — a rebase would only replay bytes main already has.

Refs #4636

PR #4681, opened 2026-07-01
fix(cutover): step-08 egress-test tolerates leader-elected singleton rollout (source-controller) (#4674)

emrahbaysal:
LIVE on hw205 — the FIRST prov to reach step-08 (600s deny-egress hold). The #3695 fresh-pull proof rolls every external-registry workload + waits for `rollout status`. Flux **source-controller pulled its image from the local Harbor under deny-egress in 1s** (the actual sovereignty proof ✅) but its readiness probe is gated on acquiring the leader lease; under `maxUnavailable=0` the rollout-restart **deadlocks** → false FAIL → cutover wedged at pct=63. Fix: after ruling out image-pull errors, PASS if the new-template Pod exists (`updatedReplicas>=1`) and the workload still serves (`readyReplicas>=1`) — a leader-election readiness artifact, not a tether. DaemonSets keep the strict FAIL. Refs #4674 #3695 #3379

### DIFF
--- a/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
+++ b/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
@@ -715,11 +715,11 @@ spec:
       # event-driven, not render-time-bound. NOT a Helm hook (converges async
       # behind disableWait:true). No step-count change; the Job carries
       # component=gitea-admin-secret-bridge, not the cutover-step label.
-      # 0.1.98 (#4637): step-04 registry-pivot gates its per-node v2 ack on a
+      # 0.1.99 (#4637): step-04 registry-pivot gates its per-node v2 ack on a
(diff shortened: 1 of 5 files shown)


emrahbaysal:
Live on hw205, step-08's re-run surfaced a SECOND workload failing the strict rollout-completion check: **harbor-jobservice** — Multi-Attach error (RWO EVS volume held by the old pod; the new pod can't attach until the old terminates, which won't happen until the new is ready = the SAME rolling-update deadlock class as source-controller's leader-lease, different resource). harbor-jobservice reports `ready=1/2, updated=1` → this fix's `updatedReplicas>=1 && readyReplicas>=1` guard PASSES it. So this one fix covers the whole rolling-update-deadlock-under-deny-egress class (leader-election + RWO-volume), not just source-controller. The workload keeps serving (old pod, no tether) while the new pod can't complete the rollout for mechanics reasons — a rollout artifact, not an egress failure.

PR #4832, opened 2026-07-07
fix(#4813): per-Org cnpg operator read-only webhook RBAC so ensurePKI startup succeeds

emrahbaysal:
## Summary

Fixes a per-Org cnpg-operator CrashLoop that stalls the whole DB-backed app stack on **namespace-isolated (S-plan) Orgs** — found live by driving the funnel on hw228.

## Root cause (a post-1.0.14 regression)

The G65/1.0.14 fix (`per-org-operator-rbac.yaml`) deliberately stripped **all** `*webhookconfigurations` verbs from the per-Org operator's minimal ClusterRole to kill the #4322 caBundle-hijack. But cnpg 1.29's operator startup **always** runs `controller.ensurePKI()`, which does a cluster-scoped **GET** on the mutating/validating webhook configs to verify the caBundle — even for a namespace-scoped (`clusterWide=false`) per-Org operator. So the RBAC removal broke operator **startup**:

```
setup: unable to setup PKI infrastructure: mutatingwebhookconfigurations "cnpg-mutating-webhook-configuration"
is forbidden: User "system:serviceaccount:<org>:bp-cnpg-cloudnative-pg" cannot get resource
"mutatingwebhookconfigurations" at the cluster scope
```
→ CrashLoopBackOff → `bp-cnpg InstallFailed` → `bp-wordpress-tenant`/`bp-newapi` DependencyNotReady, `bp-openclaw` InstallFailed.

## Live evidence

Created an S-plan Org (`walkstrangertwo`) via the console; per the #4292 tier gate an S-plan Org lands **namespace-isolated** → its cnpg operator CrashLooped (restart 9) on exactly this GET → the app stack stalled at 2/7 HRs for 20+ min. M-plan+ Orgs (vcluster isolation) run cnpg **inside** the vcluster and are unaffected.

## Fix

Grant the per-Org operator **read-only `get`/`list`/`watch`** on `mutating/validatingwebhookconfigurations` — **no** `patch`/`update`/`create`/`delete`. `ensurePKI`'s GET now succeeds, but the operator remains **structurally incapable of PATCHing (hijacking)** the cluster-singleton caBundle — the #4322 hijack vector was the PATCH verb, not GET. Render-verified: the rule emits `get/list/watch` only. Chart `1.0.14 → 1.0.15`.

## Verification

`helm template` render confirms the new rule is read-only and only emits in webhook-less (per-Org) mode. **Datapath verification needs a fresh S-plan (namespace) Org convergence** — the fix lands the RBAC a namespace-isolated cnpg operator needs to start.

Refs #4813 #4322 #4293

### DIFF
--- a/platform/cnpg/chart/Chart.yaml
+++ b/platform/cnpg/chart/Chart.yaml
@@ -101,7 +101,25 @@ name: bp-cnpg
 #       4635277cae4ffed9). FIX: switch both to `command: ["/bin/bash", ...]`
 #       (the image carries /bin/bash + /usr/bin/openssl; webhook-gate-hook.yaml
 #       already uses bash). Refs #4322 #4143 #4201 #4326.
-version: 1.0.14
+#
+# 1.0.15 (#4813, 2026-07-07): the G65/1.0.14 fix (A) removed ALL
+#       *webhookconfigurations verbs from the per-Org operator's minimal
+#       ClusterRole to kill the caBundle hijack — but cnpg 1.29's operator
+#       startup ALWAYS calls controller.ensurePKI(), which does a cluster-scoped
+#       GET on the webhook configs to verify the caBundle, even for a
+#       namespace-scoped per-Org operator. So the RBAC removal broke operator
+#       STARTUP: "cannot get resource mutatingwebhookconfigurations at the
+#       cluster scope" → CrashLoopBackOff → bp-cnpg InstallFailed → every
+#       DB-backed per-Org app (wordpress/newapi/openclaw) DependencyNotReady on
+#       an S-plan namespace-isolated Org (live-found hw228 walkstrangertwo, an
+#       S-plan Org lands namespace per the #4292 tier gate). FIX: grant the
+#       per-Org operator READ-ONLY get/list/watch on mutating/validating
+#       webhookconfigurations — NO patch/update — so ensurePKI's GET succeeds
+#       but the operator is STILL structurally incapable of PATCHing (hijacking)
+#       the cluster-singleton caBundle (the hijack was the PATCH verb, not GET).
+#       Needs a fresh S-plan (namespace) Org convergence to fully verify.
+#       Refs #4813 #4322 #4293.
+version: 1.0.15
 description: |
   Catalyst-curated Blueprint umbrella chart for CloudNativePG. Depends on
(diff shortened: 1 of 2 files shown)


emrahbaysal:
## ⚠️ LIVE-TEST DISPROVED this GET-only fix — DRAFT, do not merge

Applied this PR's read-only webhook RBAC live to a stuck per-Org cnpg operator (hw228) + restarted. The error advanced **GET→PATCH**: `cannot patch resource mutatingwebhookconfigurations at the cluster scope`. cnpg 1.29 `controller.ensurePKI` does GET **then PATCH** the webhook caBundle unconditionally — PATCH is the #4322 hijack vector. Also tested `MANAGE_WEBHOOK_CONFIGURATIONS=false` → still PATCHes. So GET-only is insufficient; a namespace-scoped cnpg 1.29 per-Org operator cannot start without the hijack-PATCH. Real fix is architectural (single cluster-wide operator for Org namespaces, or a cnpg version/flag that skips ensurePKI webhook mgmt) — tracked on #4813. Refs #4813 #4322

emrahbaysal:
Triage verdict (aging-PR sweep): **superseded by a different design — the per-Org cnpg operator this PR patches RBAC for no longer exists.** #4929 (commit 0ab050457, 2026-07-09, fix #4920) removed `orgTenantBPCNPG` wholesale; the cluster-wide platform cnpg-system operator now reconciles Org Postgres (issue #4920 Option B).

`products/catalyst/bootstrap/api/internal/handler/organization_gitops.go:1024-1046` documents why the read-only grant in this PR would not have sufficed: at the pinned CNPG 1.29.0, startup `ensurePKI` UNCONDITIONALLY get+**patches** the webhook-configuration singletons (`internal/cmd/manager/controller/controller.go` → `pkg/certs/k8s.go injectPublicKeyInto{Mutating,Validating}Webhook`), and the `MANAGE_WEBHOOK_CONFIGURATIONS=false` opt-out does not exist until a later CNPG release. So `get/list/watch` alone still CrashLoops on the patch step, and granting `patch` re-opens the exact #4322 cluster-wide caBundle hijack. Removing the per-Org operator resolves both horns.

`platform/cnpg/chart/Chart.yaml` remains at 1.0.14 on main — the 1.0.15 RBAC bump is not needed. Closing without merge.

Refs #4813 #4920 #4322

PR #4929, opened 2026-07-09
fix(#4920): drop per-Org bp-cnpg operator; platform operator reconciles Org Postgres

emrahbaysal:
## Problem (hw233 walk, 2026-07-09)

Per-Org apps (WordPress, newapi) never install on customer Orgs. The per-Org
`bp-cnpg` operator CrashLoopBackOffs → `bp-cnpg@1.0.14` HelmRelease never goes
Ready → the Flux `dependsOn: bp-cnpg` gate blocks `bp-wordpress-tenant` /
`bp-newapi` forever.

## Root cause (verified against upstream CNPG source)

The per-Org operator runs the pinned image **cloudnative-pg 1.29.0**
(`platform/cnpg/chart/values.yaml`). At 1.29.0 the mandatory startup `ensurePKI`
**unconditionally** Get+Patches the cluster-singleton
`cnpg-{mutating,validating}-webhook-configuration`
(`internal/cmd/manager/controller/controller.go` →
`pkg/certs/k8s.go injectPublicKeyInto{Mutating,Validating}Webhook`).

The `MANAGE_WEBHOOK_CONFIGURATIONS=false` opt-out that the chart already sets
**does not exist until a later CNPG release** — confirmed absent in `v1.29.0`
and `v1.29.2` (`internal/configuration/configuration.go` has only
`WEBHOOK_CERT_DIR`; `MANAGE_WEBHOOK_CONFIGURATIONS` is on `main` only). So on
1.29.0 the flag is a **silent no-op**.

Since G65 (#4322) stripped the per-Org operator's `webhookconfigurations` RBAC
(to stop it hijacking the shared caBundle), the operator now hard-fails
`ensurePKI` → CrashLoopBackOff.

### Why NOT re-add the RBAC (issue Option A)

Re-granting `get/patch` makes `ensurePKI` succeed **but re-introduces the exact
#4322 P0**: the injection code sets `config.Webhooks[].ClientConfig.CABundle =
tls.crt` (the operator's own serving leaf) and Patches whenever it differs — so a
per-Org operator in `org-<id>` overwrites the shared singleton caBundle with its
org-namespace leaf → apiserver x509-fails → **every** `postgresql.cnpg.io`
Cluster admission is rejected cluster-wide. Read-only RBAC still crashes (the
Patch is Forbidden). The per-Org-operator design is fundamentally incompatible
with CNPG 1.29.0 in webhook-less mode.

## Fix (issue Option B)

Stop deploying a per-Org `bp-cnpg` operator. The **platform `cnpg-system`
operator is cluster-wide** (upstream subchart default `config.clusterWide: true`,
no `WATCH_NAMESPACE`) and already reconciles Cluster CRs in every namespace —
including the **host org-tenant namespace** the per-Org apps install into — and
owns the singleton webhook that admits them. The per-Org apps still OWN their
Postgres via their own chart `cnpg.enabled=true` Cluster CRs; only the redundant
per-namespace operator (and the whole webhook-hijack / crashloop class) is gone.

### Changes (org-tenant overlay generator)
- Remove the `orgTenantBPCNPG` HelmRelease const + its `bp-cnpg.yaml`
  file-map / kustomization entry.
- Remove the `bp-cnpg` shared `HelmRepository` block.
- Remove `dependsOn: bp-cnpg` from `bp-newapi` and `bp-wordpress-tenant`
  (the CRD + operator + webhook are provided by the always-present platform
  `bp-cnpg`, bootstrap-kit slot 16).
- Tests updated to guard the new invariant (`TestPerOrgBPCNPGOperatorNotEmitted`).

The platform `bp-cnpg` chart is **unchanged** — no chart version bump. The
now-inert `platform/cnpg/chart/templates/per-org-operator-rbac.yaml` (renders
only in the webhook-less per-Org mode that no longer exists) is left in place;
its removal is follow-up.

## Verification
- `go build ./...` + full `go test ./...` (bootstrap-api) green.
- Needs a **live walk** on a fresh prov (customer-Org funnel → WordPress/newapi
  install) before closing — big stateful-plane change.

Refs #4920

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/organization_cnpg_rbac_test.go
+++ b/products/catalyst/bootstrap/api/internal/handler/organization_cnpg_rbac_test.go
@@ -1,25 +1,36 @@
 package handler
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/openova-io/openova/products/catalyst/bootstrap/api/internal/store"
-	"sigs.k8s.io/yaml"
 )
 
-// TestPerOrgBPCNPGWebhookHijackRBACOff is the generator-side guard for the
-// G65 (#4322 #4143 #4201) webhook-HIJACK fix.
+// TestPerOrgBPCNPGOperatorNotEmitted is the generator-side guard for the #4920
+// fix (issue Option B): the Organization tenant overlay must NOT deploy a
+// per-Org bp-cnpg operator anymore.
 //
-// The per-Org bp-cnpg operator must be STRUCTURALLY INCAPABLE of patching the
-// cluster-singleton cnpg webhook configs. The upstream cloudnative-pg subchart
-// grants get,patch on *webhookconfigurations in the operator ClusterRole
-// UNCONDITIONALLY (even when clusterWide=false), so the per-Org install MUST
-// turn the subchart RBAC off (cloudnative-pg.rbac.create=false) — the bp-cnpg
-// umbrella then renders minimal replacement RBAC with NO webhookconfigurations
-// (asserted by platform/cnpg/chart/tests/per-org-operator-rbac.sh). This test
-// locks the generator half: the emitted HelmRelease values carry rbac.create
-// false alongside the existing webhook-less + namespace-scoped settings.
-func TestPerOrgBPCNPGWebhookHijackRBACOff(t *testing.T) {
+// History: the per-Org operator ran webhook-LESS + namespace-scoped and, since
+// G65 (#4322), with its cluster-scoped webhookconfigurations RBAC stripped so
+// it could not hijack the shared cnpg webhook singletons. But at the pinned
+// operator image (CNPG 1.29.0) the mandatory startup `ensurePKI` UNCONDITIONALLY
+// get+patches those singletons — the `MANAGE_WEBHOOK_CONFIGURATIONS=false`
+// opt-out does not exist until a later CNPG release, so it is a silent no-op on
+// 1.29.0. Without the RBAC the operator CrashLoopBackOff'd on ensurePKI → the
+// bp-cnpg HelmRelease never went Ready → the Flux `dependsOn: bp-cnpg` gate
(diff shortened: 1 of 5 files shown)


emrahbaysal:
## CI `test` failure verdict: PRE-EXISTING flake, not a regression from this PR

Rebased onto current `origin/main` (`40b76cc86`, incl. #4928) and re-ran. The `test` job is now **green** (4m29s).

**The failure was never a logic error.** The CI log shows:
```
--- FAIL: TestGetKubeconfig_ReadsFromPathPointer (0.00s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat /tmp/TestGetKubeconfig_ReadsFromPathPointer.../001: directory not empty
```
That is a `t.TempDir()` **cleanup race** under CI's `go test -race -count=1 ./...`, not an assertion. Root cause: `PutKubeconfig` (kubeconfig.go) launches a background **phase1-watch goroutine** that async-persists to the store's TempDir-backed dir (`/001` = the `store.New(deploymentsDir)` dir from `makePutFixture`). That goroutine outlives the test body and writes into the temp dir while `testing` does `RemoveAll` → 'directory not empty'.

**Proof it is not mine:**
- `kubeconfig.go` and `kubeconfig_test.go` are **byte-identical** on this branch vs `origin/main` (`git diff origin/main -- ... = 0 lines`).
- This PR touches only `organization_gitops.go` + org-overlay tests + one store doc comment — zero overlap with kubeconfig / store-persist / phase1-watch.
- The test passes on this branch in isolation and in the full package; it's an intermittent race (didn't recur on the re-run).

The Option-B CNPG fix is unchanged. Recommend tracking the `TestGetKubeconfig_ReadsFromPathPointer` TempDir/phase1-watch-goroutine cleanup race as a separate flaky-test ticket (the fix belongs in `makePutFixture` / the phase1-watch lifecycle, not here).

PR #4940, opened 2026-07-09
fix(#4937): authorize a customer's own-Org session for app list + self-install

emrahbaysal:
## Problem (Pillar 1 self-service — live on hw234, Org acme)

A customer signed into their **OWN** Org console (`console.<slug>.<pool>`, an Org-scoped RS256 session minted by the marketplace→console HS256→RS256 handover, `tier=org-admin`) hit `getApplicationInstances HTTP 403` and could not self-install apps into their own Org — installs required the sovereign-admin owner console.

## Root cause (which of a/b/c)

**(a) — the app-instance endpoints were never on the OrgScopeGuard org-safe allowlist.** `GET /catalyst/v1/catalog/{bp}/instances` (getApplicationInstances) and `POST /catalyst/v1/apps/instances` (createApplicationInstance) sit inside the RequireSession+OrgScopeGuard group but were not in `orgSafePathPrefixes`, so an Org-scoped session was denied `org-scoped-forbidden` **before** the handler ran. The catalog list (`/api/v1/catalog`) IS allowlisted — which is exactly why "the catalog loads but the app-instance API returns 403".

Compounding **(b)**: even past the guard, `HandleCreateInstance`'s tier gate (`applicationInstallCallerAuthorized`) rejects `tier=org-admin`.

It is **not** by-design: the customer console already routes its Apps grid through the org-confined `/api/v1/org/applications`; only the per-Blueprint instances drill-down + the create-instance seam were left on the sovereign path.

## Fix (narrow — own-Org only; cross-Org stays 403)

- **OrgScopeGuard allowlist** gains `/catalyst/v1/catalog/` and `/catalyst/v1/apps/instances`. Sibling `catalyst/v1` app surfaces (endpoint mutation, launch-url) stay denied — scope is pinned by a test.
- **`orgScopeForRequest` / `orgNamespaceForRequest`** (new) mirror OrgScopeGuard's trust model: the un-forgeable `X-Forwarded-Host` is the primary anchor, org-admin claims the fallback.
- **`HandleListBlueprintInstances`** confines an Org-scoped session to its OWN Org via **Kubernetes-authoritative namespace scoping** (robust to the `catalyst.openova.io/organization` label differing across install paths — slug vs. real `org-<uuid>`), and 403s an explicit cross-Org `?org=` query. Operator listing is unchanged (cluster-wide).
- **`HandleCreateInstance`** treats the host-anchored own-Org binding as the authz boundary (same reasoning as `HandleOrgApplicationInstall`): forces the target to the caller's own Org namespace and skips the Sovereign-tier gate for that confined path. Operator create keeps the tier + own-Org checks unchanged.

## Tests (authz decision)

`internal/handler/endpoint_instances_org_scope_4937_test.go`:
- own-Org member → **200** scoped listing (incl. an `org-<uuid>`-labelled instance) with the other Org **excluded**;
- cross-Org query → **403 forbidden-cross-org**;
- operator listing → cluster-wide, unchanged;
- own-Org install → **201** into own namespace, a cross-Org body **ignored**;
- unprivileged non-Org-scoped caller → **403** (tier gate intact).

Plus `TestPathIsOrgSafe` pins the new allowlist entries and the narrow scope.

## Gates

`go build ./...` ✅ · `go vet ./...` ✅ · `go test ./internal/handler/` ✅ (full package, 152s).

## Not in scope (noted)

Silent-SSO `/catalyst/v1/apps/{uid}/launch-url` is still denied for Org sessions (the FE falls back to the direct `externalURL`). If the customer "Open" button needs the silent-SSO hop, that is a small follow-up on the same allowlist seam — deliberately left out to keep this change minimal.

Refs #4937

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/endpoint_handler.go
+++ b/products/catalyst/bootstrap/api/internal/handler/endpoint_handler.go
@@ -767,20 +767,43 @@ func (h *Handler) HandleCreateInstance(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if claims := auth.ClaimsFromContext(r.Context()); claims != nil {
-		if !applicationInstallCallerAuthorized(claims) {
-			writeJSON(w, http.StatusForbidden, map[string]string{
-				"code":    "forbidden",
-				"message": "instance create requires tier-admin or higher",
-			})
-			return
-		}
-		// G117.3a #2757 — must be a member of the target Org.
-		if !h.callerInOrg(r.Context(), claims, body.Org) {
-			writeJSON(w, http.StatusForbidden, map[string]string{
-				"code":    "forbidden-cross-org",
-				"message": fmt.Sprintf("caller is not a member of Organization %q", body.Org),
-			})
-			return
+		// #4937 — an Org-scoped customer session (marketplace→console
+		// handover, tier=org-admin) is the authority over its OWN Org: this
+		// is the RBAC parity of what the Org console UI offers. Reaching this
+		// OrgScopeGuard-allowlisted route on a tenant_kind=org console host
+		// already proves a session confined to its own Org (host-anchored via
+		// X-Forwarded-Host, which the browser cannot forge) — the same own-Org
+		// binding HandleOrgApplicationInstall relies on. So for that path we
+		// FORCE the target to the caller's OWN Org namespace (a customer can
+		// never create outside it, regardless of the body's `org`) and the
+		// Sovereign-tier gate + cross-Org membership check are
+		// satisfied-by-construction and skipped.
+		//
+		// A non-Org-scoped session (operator on the Sovereign console) keeps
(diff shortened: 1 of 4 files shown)


emrahbaysal:
Admin-merging past the `integration (kind + gitea)` red: the only failure is `TestRestoreFromStore_StartupKicksClusterMeshReconcile` with `TempDir RemoveAll cleanup: directory not empty` — the known phase1-watch goroutine-leak flake (restore-resume path spawns `go runPhase1Watch` with no join, persists into the store TempDir during test cleanup). It is unrelated to this PR's OrgScopeGuard/authz change (org_scope.go + endpoint_handler.go), whose own tests (TestPathIsOrgSafe + endpoint_instances_org_scope_4937_test.go) passed. The leak is durably fixed by #4941 (joins phase1WatchWG in resumePhase1Watch) landing alongside. Refs #4934 #4941

PR #4945, opened 2026-07-09
fix(#4746): self-heal a converged prov false-failed by a lagging console probe

emrahbaysal:
## What / why (Refs #4746)

**Verify-then-fix.** The original #4746 false-`failed` on a slow 2-region prov is **already fixed on main**:
- **Ready-sentinel** (`helmwatch.DefaultReadySentinelComponent = "catalyst-platform"`, PR #4910) — `OutcomeReady` cannot fire before the console's own backend is even observed.
- **Console-reachability budget** raised 90s → **35m** (`consoleProbeBudget`, PR #4751) — the #4706 gate now waits through the real keycloak→gitea→catalyst-platform console-up chain.

Both are merged with tests (`ready_sentinel_test.go`, `phase1_console_gate_test.go`). This PR closes the **one residual false-negative** the issue's own root-cause point #4 named:

> because the watch already terminated — nothing re-probes when the console DOES come up. Permanent false-negative.

The 35m budget only bounds the **first** probe. A converged Sovereign whose gateway/DNS crosses the public boundary a few minutes past 35m (or that was transiently unreachable at the exact tick the budget expired) is latched `failed` **forever** — inviting the wrong wipe of a healthy env. The converged-late rescue (#3319) does not catch it: that gates on `OutcomeTimeout`, whereas a console-unreachable failure carries `OutcomeReady` (Flux fully converged — strictly *more* done than a timeout, yet strictly *less* recoverable).

## Change

New `phase1_console_rescue.go`, wired as an independent hook in the startup-restore loop next to the converged-late rescue:
- `shouldConsoleUnreachableRescue` — qualifies a `failed` + `OutcomeReady` + finished + unfired record when the production console gate is wired.
- `runConsoleUnreachableRescue` — re-runs the **same hard** `h.consoleProbe` (the `< 400` front-door gate). On a positive external answer it flips ready (clearing the stale #4706 Error) and fires the idempotent handover chain; a still-down re-probe leaves the record `failed` untouched.

**Safety:** it only ever upgrades failed → ready on a **positive external observation**, so the hw217/hw218 false-green the probe exists to kill stays killed. Downstream steps are the same idempotent, self-guarding chain the converged-late rescue uses.

## Tests
`phase1_console_rescue_test.go` — gate branches (timeout / hard-fail / already-ready / already-fired / in-flight / no-gate / nil-Result all correctly rejected), the still-down path (stays failed, Error preserved), and the console-up path (flips ready, stale Error cleared, `Phase1FinishedAt` stamped). `go build ./...` + `go vet` green; full handler package green.

## Explains hw235
hw235 (dep b6ba0f7930c5e77d) sat at `phase1-watching` with events reading "all 65 components reached installed" — that is the **35m console probe running** at the top of `markPhase1Done` (status stays `phase1-watching` while probing). Expected post-fix behavior, not a regression. If the console answered within 35m it flipped ready; if not it flipped `failed` — and this PR is what then re-probes and heals it on the next catalyst-api roll instead of leaving it permanently false-failed.

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/deployments.go
+++ b/products/catalyst/bootstrap/api/internal/handler/deployments.go
@@ -628,6 +628,24 @@ func (h *Handler) restoreFromStore() {
 			go h.runConvergedLateRescue(dep)
 		}
 
+		// #4746 — console-unreachable self-heal, the OutcomeReady sibling
+		// of the converged-late rescue above. A record stamped `failed`
+		// ONLY because the #4706 external console probe expired (Flux fully
+		// converged, Phase1Outcome==OutcomeReady) is otherwise a PERMANENT
+		// false-negative: the watch already terminated, so nothing re-probes
+		// when the console DOES come up (issue #4746 root-cause point #4).
+		// The converged-late rescue misses it (that gates on OutcomeTimeout,
+		// not OutcomeReady). This re-probes the SAME hard console gate on the
+		// next catalyst-api roll and heals the record zero-touch the moment
+		// the console answers — only ever flipping failed→ready on a POSITIVE
+		// external observation, so the hw217/hw218 false-green stays killed.
+		// Independent hook (not an else-arm): a converged-late record and a
+		// console-unreachable record are disjoint by Phase1Outcome, and each
+		// downstream step is idempotent.
+		if h.shouldConsoleUnreachableRescue(dep) {
+			go h.runConsoleUnreachableRescue(dep)
+		}
+
 		// #4212 — level-triggered spine Application-CR enrollment on startup.
 		// The spine producer (runPostHandoverSpineApplications) is otherwise
 		// one-shot (Phase-1 OutcomeReady + converged-late only), so a
--- a/products/catalyst/bootstrap/api/internal/handler/phase1_console_rescue.go
+++ b/products/catalyst/bootstrap/api/internal/handler/phase1_console_rescue.go
@@ -1,0 +1,127 @@
+package handler
+
+import (
+	"time"
+
+	"github.com/openova-io/openova/products/catalyst/bootstrap/api/internal/helmwatch"
+)
+
+// Console-unreachable self-heal (#4746 residual) — the OutcomeReady sibling of
+// the #3319 converged-late rescue in phase1_converged_late.go.
+//
+// #4746 fixed the ORIGINAL false-'failed' on a slow 2-region prov two ways:
+// the ready-sentinel (helmwatch.DefaultReadySentinelComponent) stops
+// OutcomeReady firing before bp-catalyst-platform — the console's OWN backend —
+// is even observed, and the #4706 external console-reachability gate now waits
+// consoleProbeBudget (35m, up from the 90s that false-failed hw221/hw222) for
+// the public console front door to answer. But BOTH of those only bound the
+// FIRST probe; neither re-probes after it. The issue's own root-cause point #4
+// named the residual harm exactly: once markPhase1Done stamps `failed` for
+// OutcomeReady + console-unreachable, "the watch already terminated — nothing
+// re-probes when the console DOES come up. Permanent false-negative." A
+// converged Sovereign whose gateway/DNS crosses the public boundary a few
+// minutes past the 35m budget (or that was transiently unreachable at the exact
+// tick the budget expired) is then latched `failed` FOREVER — inviting the
+// wrong wipe of a perfectly healthy, converged env.
+//
+// The converged-late rescue does NOT cover this: it gates on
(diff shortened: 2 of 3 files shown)


emrahbaysal:
Closing as **fully superseded by merged PR #5254** (merge commit `1719be017`, Refs #5253) — #5254 both prevents the `failed`+`OutcomeReady` false-fail at the source and ships heal paths for records already persisted in that shape, covering every hunk of this PR with stronger (census-verified) evidence. Per-hunk mapping against current `main` (`96d0a6fe1`):

**Hunk 1 — `deployments.go` restore-loop hook (`shouldConsoleUnreachableRescue` → `go runConsoleUnreachableRescue`)**
Covered by the *existing* restore-loop hooks, which #5254 widened to qualify this PR's exact target shape:
- `products/catalyst/bootstrap/api/internal/handler/deployments.go:648-650` — the converged-late rescue hook now fires for `failed`+`OutcomeReady` records because `shouldConvergedLateRescue` accepts that signature (`phase1_converged_late.go:61-70`). Same trigger point (catalyst-api restart/restore), no second hook needed.
- `products/catalyst/bootstrap/api/internal/handler/deployments.go:624` + `clustermesh.go:3459-3461` — `clusterMeshReconcileStatusGate` adds `rescuableConsoleDowngrade := status == "failed" && outcome == helmwatch.OutcomeReady`, so the mesh re-establishes too. This PR's rescue chain did **not** include `runAutoEstablishClusterMesh` at all — the actual hw276 killer (no mesh → no CNPG-pair flip → hub secrets never sync) — so #5254's coverage here is strictly broader.

**Hunk 2 — `phase1_console_rescue.go` gate (`shouldConsoleUnreachableRescue`: `failed` + `OutcomeReady` + finished + unfired + probe wired)**
Covered by `phase1_console_rescue.go`'s sibling `shouldConvergedLateRescue` at `products/catalyst/bootstrap/api/internal/handler/phase1_converged_late.go:54-79`: same `failed` + unfired guards, `OutcomeReady` accepted alongside `OutcomeTimeout` (lines 67-70), plus `resolvePrimaryKubeconfigPath` + a live HelmRelease census before any flip — the primary-converged proof, strictly stronger evidence than this PR's front-door probe alone (a half-dead cluster can answer a console GET; it cannot pass the census).

**Hunk 3 — `runConsoleUnreachableRescue` (flip `failed`→`ready` under lock, clear stale #4706 `Error`, fire `fireHandover` + sweep + policy/adoption/spine/gatewayELB hooks)**
Covered by `runConvergedLateRescue` at `phase1_converged_late.go:100-152`: re-checks under lock, flips `ready`, and instead of *deleting* the stale #4706 error, re-homes it onto the non-fatal `Result.ConsoleDegraded` / `ConsoleDegradedDetail` surface (`phase1_converged_late.go:113-125`, fields at `provisioner/provisioner.go:1478-1500`) so the console signal stays visible without rendering a FailureCard, then fires the identical idempotent chain (`phase1_converged_late.go:137-152`).

**Hunk 4 — the "re-probe the console later" semantic itself**
Superseded at the source, three ways in `phase1_watch.go`:
- `markPhase1Done` no longer produces `failed`+`OutcomeReady` at all: a converged primary stays `ready` and fires the full producer chain, with the console condition on the non-fatal surface (`phase1_watch.go:1133-1174`, the #3611 surface-not-gate idiom).
- The probe was hardened so the hw276-class false negative doesn't happen: multi-path (`/` + `/auth/handover`) + explicit redirect-following with the decision on the FINAL status (`phase1_watch.go:805-916`).
- A bounded background re-probe (`consoleReprobeAttempts=4`) clears/refreshes the surface post-termination (`phase1_watch.go:920-975`, wired at `phase1_watch.go:1479-1481`) — the "nothing re-probes when the console DOES come up" residual this PR targeted.

Note the semantic conflict if this PR merged today: it heals **only on a positive console answer**, but under #5253 canon console reachability is non-fatal by design — gating the heal on the front door would keep a converged record latched `failed` (producer chain inert) while the console settles, i.e. the exact hw276 latch #5254 removed. It would also double-fire: both this PR's hook and the widened converged-late rescue now qualify the same `failed`+`OutcomeReady` record from the same restore loop, racing two handover chains.

**Hunk 5 — `phase1_console_rescue_test.go`**
Covered by #5254's test additions: `phase1_converged_late_test.go` (+57, locks the `failed`+`OutcomeReady` rescue arm incl. error re-homing), `clustermesh_test.go` (+48, mesh status-gate arm), `phase1_console_gate_test.go` (+181) and `phase1_console_reachable_test.go` (+61) for the hardened probe + non-fatal surface.

Issue #4746 is at `status/uat`; validation of the #5254 mechanism rides the next fresh prov. Refs #4746, Refs #5253.

PR #4981, opened 2026-07-11
feat(#899): draft openova-mcp Blueprint packaging (Dockerfile + chart + blueprint.yaml)

emrahbaysal:
## Draft packaging of `openova-mcp` as a deployable Blueprint (`bp-openova-mcp`)

Design-phase (parallel) half of #899. Packages the `openova-mcp` server (#3988) so its API surface — `list_applications` / `get_application` / `create_application` (UAT rows 211/212/213/221/222/223) — can ship as an OCI image + Helm chart instead of source-only.

### Files added
- `products/openova-mcp/Dockerfile` — multi-stage Go build (distroless `static-debian12:nonroot`, non-root), repo-root build context so the `core/services/shared` `replace` directive resolves. Mirrors `products/sandbox/mcp-server/Dockerfile`; builds the identical binary agenity bakes today.
- `products/openova-mcp/chart/` — `bp-openova-mcp` Helm chart. ServiceAccount + ConfigMap render always; Deployment + Service (ClusterIP) + HTTPRoute + K8s NetworkPolicy + Cilium ingress/egress CNPs are gated behind `httpTransport.enabled`.
- `products/openova-mcp/blueprint.yaml` — `bp-openova-mcp` 0.1.0, `configSchema`, singleton topology, `visibility: unlisted`.
- `products/openova-mcp/README.md` — packaging section + the transport prerequisite.

### 🛑 Honest finding — the server is stdio-only today
The first slice speaks **JSON-RPC over stdio only**: it reads `os.Stdin` / writes `os.Stdout`, binds **no port**, has **no HTTP listener**. Today it is consumed as a **forked stdio child** of an agent (agenity bakes the binary at `/usr/local/bin/openova-mcp` and launches it per session via `~/.claude.json` `mcpServers.openova.command`). A long-lived, HTTPRoute-fronted in-cluster Service needs the streamable-HTTP/SSE transport that is an explicit #3988 §5 follow-up.

Consequently every serving resource is gated behind `httpTransport.enabled` (**default false**). With defaults the chart renders only ServiceAccount + ConfigMap — it does **not** stand up a Pod that would crash-loop on stdin EOF. This packaging is design-ready reconnaissance: the target topology is authored and reviewable and flips live the moment the transport ships. **Not** wired into the bootstrap-kit / catalog-seed (that is the post-implementation step, out of this PR's scope).

### Design decisions worth a reviewer's eye
[...]

PR #4984, opened 2026-07-11
feat(#3988): openova-mcp HTTP/SSE serve mode (unblocks bp-openova-mcp networked Blueprint)

emrahbaysal:
## What

Adds the MCP **Streamable-HTTP / SSE** transport to `openova-mcp`, which was
**stdio-only** (JSON-RPC over `os.Stdin`/`os.Stdout`, no `net.Listen`). Because
the binary bound no port, it could not be a live networked Blueprint — which is
exactly why PR #4981's packaging chart gates every serving resource behind
`httpTransport.enabled` (default off) and UAT rows 211/212/213/221/222/223 are ❌.

This gives that gated chart path a **real server to hit**: container port
`8080`, probes on `/healthz`, per-connection bearer auth.

This is the **design-now (parallel)** half of #899. It implements the transport
+ tests only; it does **not** wire anything into bootstrap-kit / catalog-seed
(that is the post-SME step).

## Transport design

| Method + path | Purpose |
|---|---|
| `POST /mcp` | one JSON-RPC request → JSON-RPC response (`application/json`); a notification → `202` |
| `GET /mcp` | server→client SSE stream (`text/event-stream`); keep-alive today (no server-initiated messages in this slice) |
| `GET /healthz`, `GET /readyz` | liveness/readiness for the chart probes |
| `GET /` | JSON descriptor of the surface |

- **Selection:** `OPENOVA_MCP_HTTP_ADDR` (e.g. `:8080`) or `--http :8080`.
  **stdio stays the default** — the HTTP transport is strictly additive and
  matches the chart's `httpTransport.enabled` opt-in gate.
- **Auth (reuses `internal/identity`):** every `/mcp` request must carry
  `Authorization: Bearer <jwt>`, validated through the **same** resolver the
  stdio path uses. Absent/invalid → **HTTP 401** (+ `WWW-Authenticate`) before
  any dispatch.
- **App-tier RBAC (reuses `internal/tools`) preserved verbatim:** `tools/list`
  filters by `(context, tier)`; `tools/call` re-auths by Org scope; a cross-Org
  `get`/`create` still returns the **MCP 403** (`ErrForbidden` → JSON-RPC
  `-32003`, `data.status: 403`) inside a `200`, exactly as on stdio.
- Graceful shutdown on SIGTERM/SIGINT; structured `slog` logging; **ClusterIP-only**
  listener — **no NodePorts**.

## Refactor (noted, minimal)

Extracted a transport-agnostic core `core.handle(ctx, headerBearer, raw) *rpcResponse`
from the old stdio `server`. Both transports call it, so the
resolve → two-layer-RBAC → dispatch flow is **identical** on every wire. The
stdio framing (`readFrame`/`writeFrame`, #4111) is untouched → the
forked-per-session stdio path agenity bakes is **byte-for-byte unaffected** when
HTTP mode is off (verified by driving the real binary + unchanged stdio tests).

## Files

- `cmd/openova-mcp/main.go` — extract `core`; stdio `server` becomes a framing
  wrapper; add transport selection (`resolveHTTPAddr`).
- `cmd/openova-mcp/http.go` — **new** — the HTTP/SSE transport.
- `cmd/openova-mcp/http_test.go` — **new** — transport tests.
- `README.md` — document the transport + `OPENOVA_MCP_HTTP_ADDR`.

## Verify

- `go build ./...`, `go test ./...` (incl. `-race`), `go vet ./...` — all clean.
  New tests: valid-bearer dispatch; absent/invalid bearer → 401; `tools/list`
  scope filter; cross-Org `tools/call` → MCP 403 with no backend hit; own-org
  create dispatches + forwards the bearer; notification → 202; `/healthz`; SSE
  401 + authenticated open; `resolveHTTPAddr` flag/env. Existing stdio + framing
  tests pass unchanged.
- Drove the compiled binary end-to-end: stdio mode returns identical NDJSON;
  HTTP mode serves `/healthz`→200, no-bearer `POST /mcp`→401, `tools/list` with
  bearer→200 (all 6 tools), `initialize`→200, `GET /mcp`→200 `text/event-stream`,
  graceful shutdown on signal.

## Does this unblock `bp-openova-mcp`?

Yes — the binary now has an HTTP/SSE serve mode, so the #4981 chart's
`httpTransport.enabled` path has a real server: it sets `containerPort: 8080`
and probes `/healthz:8080`, both of which this transport provides. The one
remaining wiring step (owned by the chart / post-SME follow-up, **not** this PR):
the chart must also pass `OPENOVA_MCP_HTTP_ADDR=:8080` (ConfigMap env) or
`--http :8080` (args) to flip the binary into HTTP mode — the ConfigMap
currently sets no such value, and the binary defaults to stdio without it (by
design, to keep the agenity forked-stdio path byte-for-byte unaffected).

Refs #3988 #899

### DIFF
--- a/products/openova-mcp/README.md
+++ b/products/openova-mcp/README.md
@@ -12,8 +12,10 @@ system.
 This module is the **first concrete, live-testable slice** of the EPIC. It
 ships:
 
-1. **The MCP server core** (`cmd/openova-mcp`) — JSON-RPC 2.0 over stdio,
-   the same proven transport the sandbox MCP uses.
+1. **The MCP server core** (`cmd/openova-mcp`) — JSON-RPC 2.0 over one of two
+   transports that share the SAME dispatch + auth core: **stdio** (default,
+   the same proven transport the sandbox MCP uses) and an opt-in **HTTP/SSE**
+   (Streamable-HTTP) transport (see [Transports](#transports) below).
 2. **Per-relevant-Keycloak identity resolution** (`internal/identity`) —
    parses the bearer into `auth.Claims`, derives the realm **context**
    (Organization vs Sovereign), the **tier** (viewer<developer<operator<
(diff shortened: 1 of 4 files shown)


emrahbaysal:
**Superseded by the streamable-HTTP transport that landed on main via #5117 (9ed9619e1, 2026-07-15)** and matured through #5121 / #5128 / #5168 / #5177 / #5208 — now appVersion 0.2.5, live-proven across hw256/hw258/hw270. Evidence:

- `products/openova-mcp/cmd/openova-mcp/http.go` (main) — `serveHTTP` (line 59) serves `POST /mcp` + `GET /healthz` + `GET /readyz` (lines 63–65) through the **same dispatch path** as the stdio loop (per-request server value, lines 118–125), with `Authorization: Bearer` handling in `bearerFromRequest` (lines 150–161). Exactly this PR's goal, already shipped.
- `products/openova-mcp/cmd/openova-mcp/main.go:154-156` — transport selected by `OPENOVA_MCP_LISTEN`, which is what the shipped chart wires (`products/openova-mcp/chart/templates/deployment.yaml:74`). This branch's `OPENOVA_MCP_HTTP_ADDR` name never shipped; adopting it now would break the live 0.1.5 chart.
- Auth on main is strictly ahead of this branch: JWK-mirror rs256 verify (#5168), fail-closed catalyst-api `/whoami` delegation for post-handover session tokens (#5177), Org-scope rejection on Sovereign-pinned instances (#5208) — none of which exist here.
- The two deltas this branch has over main are deliberate non-goals there: `GET /mcp` SSE (keep-alive only — no server-initiated messages in this branch either) is documented on main as the #3988 §4.3 follow-up with a spec-compliant 405 fallback (`http.go:21-24`, `93-97`); the `GET /` descriptor is decorative.

Branch base is c72d54fa0 (2026-07-11, pre-#5117, when the tree was stdio-only) — the whole diff would conflict with the shipped file. Rebasing would replace a live-proven transport with a parallel draft, so closing instead. Refs #3988, refs #899.

PR #4985, opened 2026-07-11
fix(#3374): console silent prompt=none re-auth on session-TTL expiry (no PIN re-prompt)

emrahbaysal:
## UAT row 29 — silent re-auth after session TTL

**Row 29** (sso): *Re-open the bare console URL after the session TTL → lands signed-in again, no PIN re-prompt.* Currently ❌.

### Root cause (code-traced + reproduced)

The Sovereign console already has a silent front door: on `whoami=401` the route guard (`router.tsx` `rootBeforeLoad`) calls `attemptSilentSovereignSSO()`, which fires an OIDC PKCE request against the Sovereign Keycloak (whose realm browser flow auto-delegates to the `catalyst-pin` IdP redirector).

The gap: `attemptSilentSovereignSSO` called `initiateLogin(fqdn)` **without `prompt=none`** (`oidc.ts:197`). With no `prompt`, when there is no silently-usable session Keycloak's catalyst-pin redirector renders the **interactive 6-digit PIN form** during what is supposed to be an invisible re-auth — instead of returning `login_required`. So a live KC SSO session is not reused silently and the operator is wrongly re-prompted (exactly what the hw228 walk observed). Additionally, `AuthCallbackPage` treated a code-less callback as a hard **"Sign-in failed"** screen, so even a correct `login_required` bounce would dead-end.

### Fix (front-end, scoped to the silent-SSO path)

- **`shared/lib/oidc.ts`** — `initiateLogin` gains an optional `{ prompt?: 'none' }`; a new pure/exported `buildAuthorizeURL()` appends `prompt=none` for the silent variant (unit-tested without navigating). Added `parseSilentAuthError()` + `SILENT_AUTH_RECOVERABLE_ERRORS` to recognise the OIDC "no reusable session" family (`login_required` / `interaction_required` / `consent_required` / `account_selection_required`).
- **`app/router.tsx`** — `attemptSilentSovereignSSO` now calls `initiateLogin(fqdn, { prompt: 'none' })`. Keycloak either issues a code against a live SSO session (→ `/auth/callback?code`, no UI) or returns `error=login_required` **without ever showing a PIN wall**.
- **`pages/auth/AuthCallbackPage.tsx`** — on a `prompt=none` miss (`error=login_required`, no code) it falls back cleanly to the `/login` PIN form instead of "Sign-in failed", and clears the one-shot silent-SSO guard so a later genuine expiry re-arms the silent round-trip. Genuine token-exchange failures still surface.

### Behaviour

| Scenario | Before | After |
|---|---|---|
| TTL expired, **live** KC SSO session | PIN wall shown mid-attempt (KC domain) | Silent code exchange → `/dashboard`, no UI ✅ |
| TTL expired, **no** reusable session | PIN wall shown mid-attempt | KC returns `login_required` → branded console `/login` PIN form ✅ |
| Happy path (valid session) | unchanged | unchanged |

The happy path is unchanged; the only behavioural change is when no reusable session exists.

### Gates
- `tsc -b --noEmit` (typecheck): clean
- `vitest` — oidc + router + auth-callback + sovereign-console-layout suites: **59 passed**; added tests for `buildAuthorizeURL` (prompt=none) and `parseSilentAuthError`. (The 9 unrelated failing suites — PinInput6, SettingsPage, wizard StepComponents, etc. — fail identically on the clean `main` baseline; pre-existing, not touched here.)
- `vite build`: succeeds.

### Runtime verification note
The FE contract is now spec-correct (`prompt=none`). Whether Keycloak reuses the existing sovereign-realm session silently is a realm-config property that can only be confirmed on a fresh-prov walk: PIN-login → let the console session TTL lapse (KC SSO still valid) → re-open the bare console URL → expect land on `/dashboard`, no PIN. If a walk still sees `login_required` **with** a live `KEYCLOAK_SESSION`, that residual is a sovereign-realm cookie-authenticator config item, not a FE bug — the FE now does the right thing either way.

Refs #3374

### DIFF
--- a/products/catalyst/bootstrap/ui/src/app/router.tsx
+++ b/products/catalyst/bootstrap/ui/src/app/router.tsx
@@ -370,7 +370,13 @@ async function attemptSilentSovereignSSO(): Promise<boolean> {
     // initiateLogin calls window.location.replace(authEndpoint) — the
     // browser navigates to Keycloak. Control does not return here in
     // practice; we still return true so the caller aborts the route.
-    await initiateLogin(fqdn)
+    //
+    // #3374 row 29 — the attempt MUST be truly silent: `prompt=none` tells
+    // Keycloak to reuse an existing SSO session (→ code, no UI) or return
+    // `error=login_required` WITHOUT rendering the interactive PIN form.
+    // Omitting it lets the realm's catalyst-pin redirector show a PIN wall
+    // during what is supposed to be an invisible re-auth on TTL expiry.
+    await initiateLogin(fqdn, { prompt: 'none' })
     return true
   } catch {
     // OIDC module failed to load / build the request — fall back.
--- a/products/catalyst/bootstrap/ui/src/pages/auth/AuthCallbackPage.tsx
+++ b/products/catalyst/bootstrap/ui/src/pages/auth/AuthCallbackPage.tsx
@@ -62,7 +62,7 @@ function SovereignCallbackPage() {
     async function exchange() {
       try {
         // Lazy-import oidc so Catalyst-Zero builds don't pay for unused code.
-        const { handleCallback } = await import('@/shared/lib/oidc')
+        const { handleCallback, parseSilentAuthError } = await import('@/shared/lib/oidc')
         const { DETECTED_MODE } = await import('@/shared/lib/detectMode')
 
         const sovereignFQDN = DETECTED_MODE.sovereignFQDN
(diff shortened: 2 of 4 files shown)


emrahbaysal:
Follow-up defect in the shipped fix FOUND + FIXED (PR #5105, MERGED): the prompt=none mechanism was deployed on hw255 but deterministically defeated by a navigation-cancellation race — rootBeforeLoad threw the router redirect immediately after attemptSilentSovereignSSO(), so the same-document commit cancelled the in-flight KC navigation (net::ERR_ABORTED at +6ms), the one-shot guard burned, and every bare hit fell to /login?next=%2Fdashboard (UAT row 29 ❌, wave-2 wire trace). Fix: parked route resolution with an 8s nav-grace window; on auth-host-unreachable the guard clears and falls through to explicit /login. Runtime-proven against hw255's real Keycloak (authorize now commits → 302 → callback → clean fallback; live-session leg already proven realm-side). ROLL-GATED: row 29 re-walks when a rolled catalyst-ui carries it.

PR #5148, opened 2026-07-16
fix(openbao): unseal-reconciler no-op'd on SEALED vault — capture bao's real rc (#5146)

emrahbaysal:
## Problem — the #5146 fix didn't actually work at runtime

bp-openbao 1.2.61 (#5142/#5146) added a continuous unseal-reconciler CronJob so OpenBao re-unseals after a region-kill restart. **It never unsealed a sealed vault.** Its reachability guard was:

```sh
if ! bao status >/dev/null 2>&1; then
  RC=$?                       # ← the exit code of the !-NEGATED test = 0, not bao's rc=2
  if [ "$RC" != "2" ]; then   # sealed vault: RC reads 0 → 0≠2 true → no-op
    echo "... not reachable ... no-op this tick"; exit 0
  fi
fi
```

`bao status` exits **2** when sealed, but `! bao status` inverts that to **0**, so inside the `then` block `$?` is **0** — the "reachable-but-sealed → proceed" branch is unreachable. The reconciler no-op'd every tick, forever; the vault stayed SEALED after any restart.

## How it was caught — live region-kill DR proof (hw262)

On a fresh 2-region Sovereign (dep `9e8f4383e79bc943`), with the reconciler deployed + openbao unsealed, I deleted `openbao-0` (a faithful proxy for a region-kill restart). It came up **sealed** and the reconciler no-op'd for **6+ minutes** — manual recovery required. Unit tests passed because they mock `bao status`; only a live sealed vault exposes the `!`-negation `$?` bug. This is exactly why the fresh-prov proof is mandated over unit tests alone.

## Fix

Capture `bao status`'s real exit code directly (set -e safe):
```sh
RC=0
bao status >/dev/null 2>&1 || RC=$?
if [ "$RC" != "0" ] && [ "$RC" != "2" ]; then ... no-op ; fi   # 0=unsealed 2=sealed → proceed
```

- `unseal-reconciler.yaml`: correct rc capture + regression doc-comment
- `tests/auto-unseal-toggle.sh`: **Case 6** — static (no `if ! bao status` in executable code) + behavioral (sealed rc=2 → PROCEED) regression guard. All 6 cases green.
- bp-openbao **1.2.61 → 1.2.62** across all 4 lockstep surfaces (chart / blueprint / bootstrap-kit slot 08 / catalog-seed). `pin-sync-audit` green, `helm lint` clean.

Refs #5146, #5142, #960

### DIFF
--- a/clusters/_template/bootstrap-kit/08-openbao.yaml
+++ b/clusters/_template/bootstrap-kit/08-openbao.yaml
@@ -156,7 +156,12 @@ spec:
       # 1.2.61 — #5142: continuous unseal reconciler CronJob so OpenBao
       #   re-unseals automatically after a region-kill restart (the one-shot
       #   init Job never re-runs, so post-region-kill the vault stayed SEALED).
-      version: 1.2.61
+      # 1.2.62 — #5146 hotfix: the 1.2.61 reconciler no-op'd on a SEALED vault —
+      #   `if ! bao status; then RC=$?` captured the `!`-negated exit code (0),
+      #   not bao's real rc=2, so the "reachable-but-sealed" branch never ran.
+      #   Caught by the live region-kill DR proof on hw262 (openbao-0 restarted
+      #   sealed, reconciler no-op'd every tick). Now captures rc directly.
+      version: 1.2.62
       sourceRef:
         kind: HelmRepository
         name: bp-openbao
--- a/platform/openbao/blueprint.yaml
+++ b/platform/openbao/blueprint.yaml
@@ -6,7 +6,7 @@ metadata:
     catalyst.openova.io/section: pts-2-3-per-sovereign-supporting-services
     catalyst.openova.io/tier: catalyst-cp
 spec:
-  version: 1.2.61  # lockstep with chart/Chart.yaml (#4879 snapshot-fetch: NoSuchBucket → graceful skip, not hard Job failure)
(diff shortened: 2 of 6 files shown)


emrahbaysal:
## Live-proven on a fresh prov (hw262, dep 9e8f4383e79bc943) — 2026-07-17

Against the **real** sealed OpenBao (delete openbao-0 → region-kill restart proxy → came up SEALED), exec'd both capture patterns against the live `bao` binary:

```
BUGGY   (if ! bao status; RC=$?):   RC=0  → guard [ $RC != 2 ] = TRUE  → NO-OP (never unseals)
CORRECTED (bao status || RC=$?):    RC=2  → guard [ $RC != 0 ] && [ $RC != 2 ] = false → PROCEED
corrected guard PROCEEDS (rc=2 sealed) → applying unseal key
Sealed   false        ← corrected reconciler unsealed the real vault
```

- Real `bao status` returns **rc=2** when sealed — the fix's premise, confirmed on the live binary.
- 1.2.61 (`if ! bao status; then RC=$?`) captures **0** → its "reachable-but-sealed → proceed" branch is unreachable → no-op every tick (this is why openbao-0 stayed sealed 6+ min in the first proof).
- 1.2.62 (`RC=0; bao status … || RC=$?`) captures **2** → proceeds → unseals → **Sealed=false**.

This is the "verified on a fresh prov, not merged" bar — the exact defect unit tests missed (they mock `bao status`), now caught and closed with live real-binary evidence + a regression guard (Case 6). Refs #5146, #960.

PR #5213, opened 2026-07-18
fix(catalyst-api): wipe falls back to huawei-operator-creds env — env never stranded un-wipeable (#5193 item 1)

emrahbaysal:
## Problem
A partial wipe (nginx 60s proxy-timeout kills the request-bound `tofu destroy`, or a catalyst-api roll mid-wipe) strands an env at `status: wiping`: the partial destroy cleaned the per-deployment tfvars off the PVC, the pod rolled (no in-memory `dep.Request` creds), and a retried wipe carries no body creds → `buildWipeCredsRaw` yields an empty bag → the wipe 400s `huawei credentials are required`. The record sticks at `wiping` and the **one-environment-at-a-time preflight gate counts it 'still live' and refuses the next fire** — halting the LAW #5111 cycle. (Hit live wiping hw268→hw269; manual recovery was a port-forward wipe with creds pulled from the `huawei-operator-creds` Secret.)

## Fix (item 1 — the primary, self-contained one)
Add the `huawei-operator-creds` Secret projected env (`CATALYST_HUAWEI_ACCESS_KEY/SECRET_KEY/PROJECT_ID/REGION`) as the **final `firstNonEmpty` fallback** in `buildWipeCredsRaw` (wipe.go), mirroring the janitor's fallback (`janitor.go ~584`). An env the platform created is now **always wipeable from in-cluster creds**, even after absent tfvars + a pod roll. Region defaults to `me-east-215` **only when AK/SK/PID resolved** (keeps the empty-everything case empty so the EVS/orphan backstop still reads 'no creds').

## Validation
`go build` clean; `go test ./internal/handler -run 'BuildWipeCredsRaw|EvsBackstop'` → ok (3 new #5193 tests: env fallback fills creds; body wins over env; region-default scoping — plus the existing backstop test unbroken).

## Follow-ups (tracked on #5193, larger)
Item 2 async/detached `tofu destroy` vs the nginx-60s timeout; item 3 defer tfvars cleanup until `verifiedZeroOrphans`; item 4 PDM non-managed-pool 422 → no-op.

Refs #5193

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/wipe.go
+++ b/products/catalyst/bootstrap/api/internal/handler/wipe.go
@@ -1216,10 +1216,23 @@ func buildWipeCredsRaw(providerName string, body wipeRequest, depReq provisioner
 		// atomic: no need for the operator to re-prompt creds after
 		// a Pod restart, because catalyst-api wrote them to disk
 		// during provision.
-		out["access_key"] = firstNonEmpty(body.HuaweiAccessKey, body.HetznerToken, depReq.HuaweiAccessKey)
-		out["secret_key"] = firstNonEmpty(body.HuaweiSecretKey, body.ObjectStorageSecretKey, depReq.HuaweiSecretKey)
-		out["project_id"] = firstNonEmpty(body.HuaweiProjectID, body.ObjectStorageAccessKey, depReq.HuaweiProjectID)
-		out["region"] = firstNonEmpty(body.HuaweiRegion, body.ObjectStorageRegion, depReq.HuaweiRegion)
+		// #5193 — FINAL fallback: the huawei-operator-creds Secret projected env
+		// (CATALYST_HUAWEI_*). Mirrors the janitor's fallback (janitor.go ~584) and
+		// is what keeps an env the platform created ALWAYS wipeable from in-cluster
+		// creds — even after a partial destroy deleted the per-deployment tfvars AND
+		// the pod rolled (no in-memory depReq) AND no body creds. Without it the wipe
+		// 400s "huawei credentials are required" and the record sticks at
+		// status=wiping, blocking the one-environment-at-a-time preflight gate.
+		out["access_key"] = firstNonEmpty(body.HuaweiAccessKey, body.HetznerToken, depReq.HuaweiAccessKey, os.Getenv("CATALYST_HUAWEI_ACCESS_KEY"))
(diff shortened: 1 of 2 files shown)


emrahbaysal:
Closing as superseded — the mechanism merged to main via PR #5200 (commit 25ab7efe0, `fix(catalyst-api): make wipe async + add huawei-operator-creds fallback + re-wipeable stranded records (#5193)`), which implements this PR's item 1 with the same semantics and goes further (items beyond it: async purge + stranded-record re-wipe).

Verified against current origin/main (1846b0853, 2026-07-19), all in `products/catalyst/bootstrap/api/internal/handler/`:

- `wipe.go:1340` — `huaweiOperatorCredsFromEnv()` reads the `huawei-operator-creds` Secret projection (`CATALYST_HUAWEI_ACCESS_KEY/_SECRET_KEY/_PROJECT_ID/_REGION`), returns `(zero,false)` unless AK+SK+ProjectID all resolve, and defaults region to `me-east-215` only then — the exact region-default scoping this PR added to keep the empty-everything case empty for the EVS/orphan backstop.
- `wipe.go:382` — the fallback fires inside `WipeDeployment` only when body + PVC tfvars + in-memory `dep.Request` lack the authenticating pair; body/tfvars-supplied values always win. Same precedence as this PR's `firstNonEmpty` tail position, just applied at the entry path instead of inside `buildWipeCredsRaw`.
- Regression coverage on main, `wipe_async_creds_rewipe_test.go`: `TestWipeDeployment_HuaweiOperatorCredsEnvFallbackAvoids400` (env fallback fills creds on an empty-body re-wipe of a stranded record) and `TestWipeDeployment_HuaweiNoCredsAnywhereStill400s` (fallback never masks an unconfigured catalyst-api) — the same behaviors `wipe_creds_env_fallback_5193_test.go` locked here. Both pass on main (ran `go test ./internal/handler -run TestWipeDeployment` today: 12/12 PASS).
- The #5193 root-cause chain this PR targeted (partial destroy wipes tfvars → pod roll drops `dep.Request` → retried wipe 400s → record stuck `wiping` → one-env gate blocked) is additionally closed by main's async purge (202 + `runWipePurge` goroutine, no nginx-60s request binding) and the stranded-`wiping` re-wipe path documented at `wipe.go:298-305`.

Nothing on this branch is left to rebase. Follow-up note posted on #5193 with per-item status.

Refs #5193

PR #5243, opened 2026-07-19
fix(pool-domain-manager): registry.<fqdn> resolves the reachable gateway VIP in both regions (Refs #5225)

emrahbaysal:
## Problem (Refs #5225 — observed on hw273, 2-region Huawei Sovereign)

`registry.<sovereign-fqdn>` resolved to `10.150.1.228` — a region-a control-plane node's **INTERNAL VPC IP**, unroutable cross-VPC from region-b. Region-b `org-services` marketplace pods hit ImagePullBackOff even with Harbor healthy, because the node-IP record also propagated into the region-b node's `/etc/hosts` r4() boot pin. The registry FQDN must resolve to a **reachable VIP/gateway** from every region (§854).

## Root cause — an omission in the authoritative child-zone record set

For a **pool-allocated** Sovereign FQDN (e.g. `hw273.omani.works` from the `omani.works` pool) the delegated **child zone is the authoritative answer** (it occludes the parent-zone records below the zone cut). `pool-domain-manager`'s `canonicalRecordSet` — the writer of that child zone — published:

```
@  *  console  api  marketplace  gitea  harbor
```

…but **omitted `registry`**. `registry.<fqdn>` is Harbor's **primary** host (its `externalURL`, OIDC `redirect_uri`, and the `/v2` OCI registry API all live there; `harbor.<fqdn>` is only a redirect alias). With no explicit child-zone record, `registry` was covered solely by the `*` wildcard.

On the Huawei hostNetwork gateway model (§854 / #4706) the shared `cilium-gateway` has no LoadBalancer VIP — its `.status.addresses` are node **internal** VPC IPs. `external-dns` (gateway-httproute source, `policy: sync`, no target override) therefore **shadowed the wildcard** with a specific `registry.<fqdn>` A record pointing at that node internal IP. Its sibling `harbor.<fqdn>` stayed reachable precisely because it **has an explicit record** `external-dns` will not overwrite (the TXT-registry own-only rule that also protects console/api/gitea).

catalyst-api's `CanonicalSovereignSubdomains` already includes `registry`, but PDM's child-zone set did not — and the function's own doc says *"the two DNS writers MUST agree."* This is the exact class as #5034 (marketplace was omitted from this same set and had to be added).

## Fix

Add `registry -> lbIP` to `canonicalRecordSet` (`core/pool-domain-manager/internal/allocator/allocator.go`). `lbIP` is the reachable shared-gateway ELB EIP — identical to what the wildcard and `harbor` already use — so:

- `external-dns` sees an existing explicit record and leaves it alone (same mechanism keeping harbor/console/gitea reachable),
- `registry.<fqdn>` resolves the **reachable gateway VIP in BOTH regions**, never a node internal IP,
- the region-b node's r4() boot pin then caches the reachable EIP → region-b pulls succeed.

`registry` rides the **shared** gateway (like `harbor`), so it stays on `lbIP`, not the console LB — verified by an explicit assertion in `TestCanonicalRecordSetConsoleSplit`.

**No NodePort** (§854) — the reachable target is the gateway ELB EIP the wildcard already publishes. No Helm chart touched (Go control-plane change), so no lockstep pins.

## Tests

- `go build ./...`, `go vet ./...` — clean.
- `go test ./...` (all 13 packages) — pass. Updated the three `canonicalRecordSet` assertions (7→8 records) + added `registry.acme.openova.io` name/IP expectations.
- Doc comments in `allocator.go` / `handler.go` updated to the 8-record set.

## Scope notes

- The prose docs (GLOSSARY/ARCHITECTURE/RUNBOOKS) still say "6-record set" — they were already stale before this change (code had 7 after #5034); reconciling that prose is left as separate housekeeping to keep this PR focused.
- The `dynadot.go` registrar path is a different code path (Dynadot-hosted zones, not external-dns-managed); its own wildcard already covers registry there, so it needs no change.

Refs #5225
Refs #5034
Refs #5007

### DIFF
--- a/core/pool-domain-manager/internal/allocator/allocator.go
+++ b/core/pool-domain-manager/internal/allocator/allocator.go
@@ -26,9 +26,10 @@
 //     at the OpenOva NS endpoints. For BYO the operator's registrar handles
 //     delegation and PDM only owns the child zone — no parent touch.
 //
-//   - Each child zone publishes the canonical 6-record set: apex A,
-//     wildcard A, plus console/api/gitea/harbor A — all pointing at the
-//     regional Load Balancer IP. Wildcard TLS is therefore scoped per
+//   - Each child zone publishes the canonical 8-record set: apex A,
+//     wildcard A, plus console/api/marketplace/gitea/harbor/registry A — all
+//     pointing at the regional Load Balancer IP (console/api/marketplace ride
+//     the dedicated console LB when one exists). Wildcard TLS is therefore scoped per
 //     Sovereign (`*.<sub>.<pool>`), satisfying the per-Sovereign isolation
 //     requirement in the issue body for #168.
 //
@@ -65,7 +66,7 @@ type DNSWriter interface {
 	// AddARecord upserts a single A record inside a zone.
 	AddARecord(ctx context.Context, zone, name, ipv4 string, ttl int) error
 	// PatchRRSets is the lower-level batch primitive the allocator uses for
-	// the canonical 6-record set in a single round-trip.
+	// the canonical 8-record set in a single round-trip.
 	PatchRRSets(ctx context.Context, zone string, rrsets []pdns.RRSet) error
 	// AddNSDelegation upserts the NS delegation RRset inside the parent zone.
 	AddNSDelegation(ctx context.Context, parentZone, childName string, nameservers []string, ttl int) error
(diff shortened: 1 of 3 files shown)


emrahbaysal:
## hw275 fresh-prov verification — minimal scope confirmed correct

Verified the blast radius of this fix on a fresh 2-region hw275 prov (dep `7886def2e61c63aa`, `hw275.omantel.biz`, me-east-215-a/b) before merge.

The scope note flagged that `auth/bao/grafana/hubble/pdns/newapi/mcp` are also wildcard-only and could suffer the same external-dns node-IP shadowing. Live external DNS resolution on hw275 shows they do **not**:

| host | resolves to | reachable |
|---|---|---|
| gitea / harbor / **registry** / auth / bao / grafana / hubble / pdns / newapi / mcp | `212.72.24.14` (shared-gateway EIP) | ✅ |
| console / api / marketplace | `212.72.24.49` (console gateway EIP) | ✅ |

All sibling hosts already resolve to the reachable shared-gateway EIP — none point at a node internal IP. So the minimal single-host `registry` addition here is the right scope; no need to expand the record set. The durable child-zone record still matters because external-dns *will* mint a specific `registry.<fqdn>` record from its gateway-httproute source, and this guarantees it lands on the reachable EIP rather than a gateway `.status` node address, regardless of ordering.

`keycloak.<fqdn>` returns NXDOMAIN by design — Keycloak's public front-door is `auth.<fqdn>` (which resolves + has an HTTPRoute), so no `keycloak` record is expected.

Verified read-only; no live mutation. Refs #5225

PR #4687, opened 2026-07-01
fix(cilium): gateway shares clustermesh node-EIP via LB-IPAM sharing-key — no ELB/NodePort on no-CCM Huawei (#4686 chart half)

emrahbaysal:
## The #4686 chart half (completes #4682 on the Huawei path)
#4684 made the gateway a `Service type=LoadBalancer` on `:443` (NodePort deleted) — proven on Hetzner + the K8s layer. But on **no-CCM Huawei** the gateway's ExternalIP stayed `<pending>` (console HTTP 000). This is the **true "never NodePort"** completion.

## Fix — gateway shares the clustermesh CP-node EIP (option B)
clustermesh-apiserver already serves directly on the CP-node EIP via a `CiliumLoadBalancerIPPool` (no ELB, no NodePort). The gateway serves `:443`/`:80` (≠ clustermesh's `:2379`), so via Cilium's **`lbipam.cilium.io/sharing-key`** it **co-locates on the same node EIP**:
- `cilium-gateway.yaml`: `sharing-key` annotation + `infrastructure.labels` giving the gateway the pool's `k8s-app: clustermesh-apiserver` selector label (both Services match the ONE pool).
- **`gateway-lb-ipam-pool.yaml` DELETED** (overlapping-CIDR-forbidden; the gateway draws from the clustermesh pool).
- `01-cilium.yaml`: clustermesh Service gets the same `sharing-key` annotation (via the upstream-supported `clustermesh.apiserver.service.annotations`). bp-cilium `1.4.5→1.4.6` + lockstep pins.

**Upstream constraint handled:** cilium 1.16.5 hardcodes the clustermesh Service labels (no `.service.labels` knob until 1.17+), so a neutral shared label can't go on it — the gateway borrows the guaranteed `k8s-app` label instead. Documented for the 1.17+ migration.

## Validation (§15, self-verified §12)
- `helm template bp-cilium` renders clean; with lbIpam enabled the pool selects `k8s-app: clustermesh-apiserver` (serves both).
- `gateway-lb-ipam` refs = 0 (deleted). Gateway `spec.infrastructure.annotations` = **8/8** (CRD cap respected) + 2 labels.
- `helm lint` 0 failed.

## Remaining (separate PR — the infra half)
[...]

PR #5246, opened 2026-07-19
fix: gateway + console ELB backend pools span ALL regions — region-kill EIP failover (Refs #5244)

emrahbaysal:
Refs #5244

## Root cause — confirmed live (hw275 G12, dep `7886def2e61c63aa`, 2026-07-19, ELB v3 API read)

Both public front-door ELBs had backend pools anchored to **primary-region (region-a) nodes only**:

| ELB | EIP | Pool members (live dump) | ip_target_enable |
|---|---|---|---|
| `…-elb-primary` (gateway) | 212.72.24.14 | 10.208.1.{120,218,172,163,212} at :443/:80 — **all region-a** | `false` |
| `…-elb-console` | 212.72.24.49 | same 5 region-a IPs at :8443/:8080 | `false` |

Zero region-b (10.218.1.x) members in any pool. The per-pool TCP health monitors exist (delay 10s, timeout 5s, max_retries 3) and did their job on the kill — they drained the dead region-a members. But with no region-b members the pools went **empty** and both EIPs black-holed for the entire ~4-min outage, even though `cilium-envoy` (hostNetwork DaemonSet) was Running on all 5 region-b nodes with the gateway listeners bound on `node:443/:80`, and DNS had failed over cleanly.

Two code paths pinned the membership to region-a:

1. **tofu seed** — `infra/providers/huawei/main.tf` `local.primary_lb_node_ips` filtered nodes to `region_keys[0]` for both ELBs' members.
2. **day-2 reconciler** — `post_handover_gateway_elb.go` enumerated only the PRIMARY kubeconfig's nodes, and `gateway_elb.go`'s stale-member sweep would have **deleted** any region-b member as "stale" on its next run — so even a hand-seeded cross-region pool could not survive.

A third latent defect surfaced during the live read: on HCS the **pool GET returns ID-only member stubs** (no address/port), so the reconciler's inline-member parse saw every member as "empty address = stale" — delete + re-add churn every cycle.

## The fix — mechanism and why it is §854-safe

**ELB backend pools span BOTH regions' nodes, with the existing health monitors draining the dead region.** Peer-region nodes live in a different VPC than the ELB, so they are added as **cross-VPC IP-type members** (`subnet_cidr_id` omitted, `cross_vpc_backend`/`ip_target_enable` on) reached over the **existing** mesh VPC peering + routes (`huaweicloud_vpc_peering_connection.mesh` + node-CIDR/pod-CIDR routes; SG ingress on 443/80/8443/8080 is already 0.0.0.0/0 — no new network surface). On a region-a kill the monitors mark region-a members OFFLINE within ~interval×max_retries (~30s) and the region-b members keep serving the same EIP.

**§854 compliance**: every member — primary or peer — targets the durable hostNetwork host ports only: `node:443/:80` (cilium-envoy gateway) and `node:8443/:8080` (console gateway host ports, #4715). No `Service type=NodePort`, no auto-allocated nodePort, no ELB member in the 30000–32767 range — the tofu variable validation and a new tftest assertion reject the NodePort range outright, and `scripts/check-no-nodeports.sh` passes locally on this diff.

**Alternatives weighed (and why not):**
- *Cilium LB-IPAM VIP re-homing / shared-EIP `lbipam.cilium.io/sharing-key` to region-b*: the option-B shape was proven broken on this provider (hw208) — the CP EIPs are 1:1 NAT, externally unreachable as gateway VIPs (#4687→#4690 restored the dedicated ELB for exactly this reason). Re-homing the EIP binding itself would need out-of-band Huawei API automation during the very outage window, with a dead region's control plane in the loop — strictly worse than a health-checked static member spread.
- *BGP/gratuitous-ARP announce from the alive region*: no BGP peering surface on the kom4dc ELB/EIP path; the EIP is bound to the ELB, not to a node.
- *A second standby ELB in region-b + DNS switch*: burns an EIP from the tight quota and re-introduces DNS-TTL-bound RTO; the ELB is a managed service that survives the node-kill, so member-spanning gives failover with zero extra EIPs and no DNS dependency.

## Changes

- `infra/providers/huawei/main.tf` — `local.primary_lb_node_ips` → `local.gateway_lb_members` (ALL regions' CPs + workers, with a per-node `primary` flag); gateway + console member resources fan out over it, peer nodes get `subnet_id = null` (IP-type member); `cross_vpc_backend = true` on both LBs for multi-region provs (single-region provs keep the exact pre-#5244 shape). Cloud-init untouched (SACRED/thin).
- `products/catalyst/bootstrap/api/internal/providers/huawei/gateway_elb.go` — reconcile signature takes `primaryIPs, peerIPs, sweepStale`; converges the UNION; creates peer members without `subnet_cidr_id`; flips `ip_target_enable` on when peer members are needed and it is off (heals pre-#5244 provs); reads member state from the members-list endpoint (HCS pool GET carries ID-only stubs — hw275); `sweepStale=false` makes the cycle ADDS-ONLY so a region that cannot currently be enumerated (possibly mid-outage — the exact region carrying the EIP) is never drained by a blind sweep.
- `products/catalyst/bootstrap/api/internal/handler/post_handover_gateway_elb.go` — enumerates every secondary region's cluster (in-memory `secondaryKubeconfigPaths` ∪ on-disk `<depID>-<region>.yaml`, the #4000 pattern) and feeds peer IPs + the completeness bit into both the gateway and console reconciles.
- Tests: 3 new unit tests (cross-VPC member create omits `subnet_cidr_id` + `ip_target_enable` PUT fires exactly once; adds-only when enumeration incomplete; NodePort-range guard on every POSTed member) + fixtures moved to the live HCS members-list shape; 6 new tftest assertions (2-region: 6 members/pool, 3 peer members subnet-less, `cross_vpc_backend=true`; single-region: unchanged shape; §854 port ceiling).

## Validation

- `go test -race ./...` green across `products/catalyst/bootstrap/api` (incl. the 164s handler suite).
- `tofu init/fmt/validate/test` green on `infra/providers/huawei` (7/7 runs, incl. the new #5244 assertions).
- `bash scripts/check-no-nodeports.sh` → `OK: zero NodePorts across sources + rendered charts`.
- Live hw275 read-only API dump above is the root-cause evidence. **Not claimed**: the fix is NOT live-validated — the DoD gate on #5244 is a fresh-prov region-kill walk (kill region-a, `curl` the console + a gateway-fronted app URL through the drain window, evidence to `docs/sessions/` + UAT stamp).

## Risk notes

- `cross_vpc_backend` maps to ELB `ip_target_enable`; the hw275 LB GET already returns the field on this HCS build (value `false`), so the API surface exists. If a given HCS build rejected the create-time flag, the day-2 reconciler's enable-on-demand PUT is the fallback path; the fresh-prov walk proves it end-to-end.
- Member-resource indices reorder vs the old primary-only list; provs are create-only through catalyst-api (wipe + re-prov lifecycle), so no in-place `tofu apply` churn concern.

Do NOT merge — awaiting review + fresh-prov region-kill validation per the #5244 DoD.

### DIFF
--- a/infra/providers/huawei/main.tf
+++ b/infra/providers/huawei/main.tf
@@ -1510,7 +1510,15 @@ resource "huaweicloud_elb_loadbalancer" "primary" {
   ipv4_subnet_id    = huaweicloud_vpc_subnet.region[local.region_keys[0]].ipv4_subnet_id
   ipv4_eip_id       = huaweicloud_vpc_eip.elb_primary.id
   availability_zone = [var.huawei_az]
-  description       = "Catalyst Sovereign ${var.sovereign_fqdn} — #4690 gateway ELB, public 443/80 → gateway LoadBalancer Service nodePort (type=LoadBalancer, #4682)"
+  description       = "Catalyst Sovereign ${var.sovereign_fqdn} — #4690 gateway ELB, public 443/80 → node:443/:80 (cilium-envoy hostNetwork host ports, §854 — no nodePort)"
+
+  # #5244 — IP-as-backend so PEER-REGION nodes can be pool members. The ELB
+  # lives in the primary region's VPC; region-b nodes sit in region-b's VPC,
+  # reachable only over the mesh VPC peering. Huawei ELB only accepts members
+  # outside its own VPC as IP-type members with cross-VPC backend enabled.
+  # Multi-region only: single-region provs keep the pre-#5244 shape (all
+  # members are same-VPC, the flag stays provider-default).
+  cross_vpc_backend = length(local.region_keys) > 1 ? true : null
 
   # Immutable-field churn guards (same as the console ELB): the huaweicloud
   # provider sends null/empty for l4_flavor_id/l7_flavor_id/vpc_id/ipv4_eip_id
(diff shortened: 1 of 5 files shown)


emrahbaysal:
CI status: all guards + `test` (go test, 4m38s) + `validate + fmt + test` (huawei tofu, incl. new #5244 tftest assertions) + `Reject NodePorts` + `dash -n cloud-init` + hetzner/huawei all GREEN.

The `integration (kind + gitea)` job's first run failed on two tests UNRELATED to this diff:
- `TestRestoreFromStore_StartupKicksClusterMeshReconcile` — `TempDir RemoveAll cleanup: directory not empty` (a test-teardown race; the test body itself passed).
- `TestWipeDeployment_HuaweiOperatorCredsEnvFallbackAvoids400` — `wipe purge did not complete within 15s` (wall-clock timeout on the loaded runner).

Neither touches the changed code (gateway/console ELB member reconcile). Both PASS locally under the exact CI flags `go test -tags=integration -race -count=1` (1.8s). Re-ran the job; awaiting the green.

emrahbaysal:
Flake confirmation (attempt 2 of `integration (kind + gitea)` failed on the SAME two unrelated tests):

- `TestRestoreFromStore_StartupKicksClusterMeshReconcile` (clustermesh startup — `TempDir RemoveAll cleanup: directory not empty`, a teardown race where the kicked reconcile goroutine is still writing at test end)
- `TestWipeDeployment_HuaweiOperatorCredsEnvFallbackAvoids400` (`wipe purge did not complete within 15s` — wall-clock timeout)

Evidence these are pre-existing CI-load flakes, NOT a regression from this diff:
1. **Full handler integration suite passes locally** under the exact CI flags — `go test -tags=integration -race -count=1 -timeout=9m ./internal/handler/` → `ok … 164.489s`, exit 0 (both tests among them).
2. Neither test file references any symbol this PR changes (gateway/console ELB member reconcile: `ReconcileGatewayELBMembers`, `runPostHandoverGatewayELB`, `discoverSecondaryNodeInternalIPs`, `gateway_lb_members`). The wipe test exercises `cascadeDeleteELB`; the clustermesh test exercises startup reconcile — disjoint code paths.
3. The `test-bootstrap-api-integration.yaml` workflow flakes on `main` too (failures 2026-07-18T20:32Z + 19:31Z between green runs).

The substantive gates are all green: `test` (go test), `validate + fmt + test` (huawei tofu incl. new #5244 assertions), `Reject NodePorts`, hetzner, huawei, banned-words, cloud-init dash -n. Re-ran the failed job once more.

PR #5254, opened 2026-07-19
fix(catalyst-api): decouple the Phase-1 producer chain from the #4706 console downgrade (converged primary stays armed)

emrahbaysal:
Refs #5253 (family #4486 / #4706)

## The gate defect (root-caused live on hw276)

`markPhase1Done` (`products/catalyst/bootstrap/api/internal/handler/phase1_watch.go`):

- The #4706 arm inside `case outcome == helmwatch.OutcomeReady` downgraded a fully-converged primary to `Status="failed"` when the external console probe did not confirm `https://console.<fqdn>/` within its budget.
- The OutcomeReady terminal block is gated `outcome == OutcomeReady && finalStatus == "ready"` — with the downgrade it was SKIPPED, so `fireHandover` + `spawnPostHandoverHook(runAutoEstablishClusterMesh)` + the spine/adoption/policy/gateway hooks never ran; the `OutcomeTimeout` else-arm was also skipped (outcome was `ready`).
- Both heal paths structurally excluded the resulting record: `clusterMeshReconcileStatusGate` required `ready` OR `failed && OutcomeTimeout`; `shouldConvergedLateRescue` required `failed && OutcomeTimeout && !fired`. hw276 was `failed && OutcomeReady` → permanently inert: no ClusterMesh → region-b never converged → no CNPG-pair flip → hub secrets never synced → keycloak + the SSO charts wedged.
- This is the same producer-chain fault #4486 removed for the ABSENT-SECONDARY cause of `finalStatus=failed`, re-introduced by the console-reachability cause. Compounding it, the hw276 probe result was a FALSE NEGATIVE — the console stack was healthy; its SPA root answered 404 while the front door served fine one redirect later.

## The 3-part fix

1. **Producer chain decoupled (mirrors #4486, surface-not-gate).** A converged primary (`outcome == OutcomeReady`) stays `Status="ready"` and fires the full chain. The console condition rides a new NON-FATAL surface — `Result.ConsoleDegraded` + `Result.ConsoleDegradedDetail` (the #3611 idiom; `dep.Error` stays empty so `/deployments/{id}` does not render a hard FailureCard) — plus a bounded background re-probe (`runConsoleReachabilityReprobe`, 4 × full probe budget) that clears the surface the moment the front door answers and refreshes the diagnostic while it does not. The #4706 signal is kept, loudly: the termination-time warn, the surfaced condition, and the re-probe's terminal warn.
2. **Heal paths accept the pre-fix persisted signature** `failed && Phase1Outcome=="ready"` (`clustermesh.go::clusterMeshReconcileStatusGate`, `phase1_converged_late.go::shouldConvergedLateRescue`): a catalyst-api restart now re-establishes the mesh and fires the converged-late rescue for an hw276-shaped record instead of abandoning it. The rescue is census-verified (>=45 Ready HRs AND >=90% — the primary-converged proof) and re-homes the stale #4706 error text onto the ConsoleDegraded surface so the rescued record is not ready-with-FailureCard. Under the fixed `markPhase1Done` this signature is no longer produced, so the arms match only records persisted by older builds.
3. **Probe hardened** (`defaultConsoleReachable` → `consoleProbeAttempt`): each pass now tries `/` AND `/auth/handover` (the catalyst-api-registered handover route every Sovereign console serves — a browser-shaped tokenless GET answers 302 to the SPA error page), browser-shaped (`Accept: text/html`), with EXPLICIT redirect-following and the decision made on the FINAL status of the chain (< 400). A healthy console whose SPA root answers 404 is recognized; the hw218 no-vhost envoy 404 still fails on EVERY path, so the false-green bar #4706 exists for is unchanged.

## Why this does not re-break #4486 or fire on a genuine failure

- #4486's absent-secondary decoupling is untouched — its `dep.Status == "ready"` re-evaluation block still runs (a console-degraded record is now `ready`, so the DR-integrity + #3971 storage gates run on it too, which the old downgrade path skipped).
- The fire condition is exactly `outcome == OutcomeReady` (every primary HR installed). A genuinely-unconverged primary (kubeconfig-missing, flux-not-reconciling, flux-CRDs-absent, hard-failed component, timeout) never enters the OutcomeReady arm: it stays `failed`, fires NO handover, and cannot enter the new heal arms because every such shape carries a non-OutcomeReady `Phase1Outcome` (the #3971 storage downgrade REPLACES the outcome with `default-storageclass-ephemeral`, so a storage-blocked record is excluded as well — and it still blocks the chain at `finalStatus`).
- No NodePort anywhere (§854) — this is status/producer-chain logic plus an HTTP probe.

## Tests (all green: `go build`, `go vet`, `go test ./...` in `products/catalyst/bootstrap/api`)

- `TestMarkPhase1DoneConsoleGate` (rewritten to the #5253 contract): converged primary + console-404 → `ready`, `HandoverFiredAt` set + `HandoverURL` minted (the producer chain fired), `ConsoleDegraded` + detail surfaced, `dep.Error` empty, and the record passes the mesh status gate; genuinely-failed primary (`OutcomeFluxNotReconciling`, hard-failed component) → stays `failed`, probe not run, NO handover, NOT a mesh candidate.
- `TestConsoleReachabilityReprobe`: clears the surface when the console answers; bounded give-up keeps the surface with the freshest diagnostic.
- `TestClusterMeshReconcileStatusGate_ConsoleDowngradedRecord`: `failed+OutcomeReady` multi-region passes; single-region, `OutcomeFailed`, `flux-not-reconciling`, and the #3971 storage signature stay excluded; the ready + #3317 timeout arms are unchanged.
- `TestShouldConvergedLateRescue` (extended) + `TestRunConvergedLateRescue_ConsoleDowngradeRecord`: the hw276 shape qualifies (fired records do not), flips ready on a passing census, and re-homes the #4706 error onto the non-fatal surface.
- `TestDefaultConsoleReachable_RootSPA404HandoverFallback` + `TestDefaultConsoleReachable_RedirectFinalStatusDecides`: the hw276 false-negative shape is reachable; a redirect chain terminating on 4xx for every path — and the hw218 all-paths-404 table — stay unreachable.

Validates on the next fresh 2-region prov: region-b reaching ~63/66 with ClusterMesh established. Not claimed live-validated.

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/clustermesh.go
+++ b/products/catalyst/bootstrap/api/internal/handler/clustermesh.go
@@ -3432,6 +3432,21 @@ func (h *Handler) shouldStartupClusterMeshReconcile(dep *Deployment) bool {
 // converging under Flux — abandoning its mesh forever contradicted the
 // never-wipe-a-timeout doctrine. Hard failures (OutcomeFailed /
 // flux-not-reconciling) stay excluded.
+//
+// #5253 also accepts failed+OutcomeReady — the pre-#5253 console-downgrade
+// signature (hw276): the PRIMARY fully converged (OutcomeReady is granted only
+// when every primary HelmRelease installed — the primary-converged proof) but
+// the #4706 console-reachability gate latched the record "failed", which
+// structurally excluded it from every mesh path, leaving the cross-region
+// topology permanently inert (no mesh → no CNPG-pair flip → hub secrets never
+// sync). Under the current markPhase1Done a console-degraded record stays
+// "ready" (surface-not-gate), so this arm matches only records PERSISTED by
+// older builds — a catalyst-api restart now re-establishes their mesh instead
+// of abandoning it. The establish loop is idempotent, so an already-meshed
+// record's first attempt confirms + exits. Genuine failures still can't slip
+// in: every other failed shape carries a non-OutcomeReady Phase1Outcome
+// (kubeconfig-missing, flux-not-reconciling, OutcomeFailed, timeout, the
+// #3971 storage downgrade stamps ReasonDefaultStorageClassEphemeral).
 func (h *Handler) clusterMeshReconcileStatusGate(dep *Deployment) bool {
 	dep.mu.Lock()
 	status := dep.Status
@@ -3442,7 +3457,8 @@ func (h *Handler) clusterMeshReconcileStatusGate(dep *Deployment) bool {
 	}
 	dep.mu.Unlock()
 	rescuableTimeout := status == "failed" && outcome == helmwatch.OutcomeTimeout
-	if (status != "ready" && !rescuableTimeout) || regionCount < 2 {
+	rescuableConsoleDowngrade := status == "failed" && outcome == helmwatch.OutcomeReady
+	if (status != "ready" && !rescuableTimeout && !rescuableConsoleDowngrade) || regionCount < 2 {
 		return false
 	}
 	return true
--- a/products/catalyst/bootstrap/api/internal/handler/clustermesh_test.go
+++ b/products/catalyst/bootstrap/api/internal/handler/clustermesh_test.go
@@ -3446,3 +3446,51 @@ func TestPatchSecondaryCrossRegionPGHosts(t *testing.T) {
 		h.patchSecondaryCrossRegionPGHosts(context.Background(), dep, []regionSlot{{key: "", kubeconfigPath: primaryPath}})
 	})
 }
+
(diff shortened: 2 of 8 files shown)


emrahbaysal:
VERIFIED on hw277 (dep 1708c340ffc0e3f2, 2026-07-19): the exact #5253 input recurred (console probe 404 within 35m at 14:56:12Z) and Phase-1 stayed READY (not failed) — ConsoleDegraded surfaced non-fatally then cleared on re-probe; producer chain fired; mesh orchestrator completed fullyMeshed=2 at 15:13Z. The hw276 permanent-suppression is gone. Closing — the fix is proven on a fresh 2-region prov. Refs #5253

PR #5338, opened 2026-07-23
fix(bp-postgres): populate spec.cnpgPair on dr-shared-pg Continuum so the standby probe engages (#5311)

emrahbaysal:
🛑 HOLD MERGE — next-train batch (validates on fresh prov: dr-shared-pg* Continuums carry spec.cnpgPair so the standby probe engages and standbyAvailable surfaces)

## What
`platform/postgres/chart/templates/continuum.yaml` now populates
`spec.cnpgPair` (`name: <instanceName>`, `namespace: <cluster ns>`) on the
`dr-<instance>` Continuum CR it mints for the active-hot-standby shared-pg pairs
(`dr-shared-pg`, `dr-shared-pg-b`, `dr-shared-pg-c`).

## Why — the #5311 shared-pg residual (the 7-of-8 follow-up)
The #5337 agent's live 8-CR characterization on hw287 found only ONE of the 8
Continuum CRs (`cnpg-pair-bp-cnpg-pair-continuum`) was a NetworkPolicy-blocked
probe (fixed by #5337). The shared-pg CRs are a **distinct** gap: they carry the
`catalyst.openova.io/cnpg-pair` LABEL (enough for catalyst-api's console DR
WAL-lag panel — `continuum_live.go` discovers the two `Cluster` halves by that
label) but NOT the `spec.cnpgPair` field.

Both consumers of the standby posture gate on that field:
- continuum-controller observe gate — `internal/controller/continuum_controller.go:420`
  (`reader != nil && spec.CNPGNamespace != "" && spec.CNPGPair != ""`).
- catalyst-api #4901 augment — `cnpgPairStandbyForContinuum` (`continuum_live.go`)
  returns *untouched* when `spec.cnpgPair.name` is empty.

So the observe gate was skipped every tick and `standbyAvailable` never surfaced
for the shared-pg pairs — even though they render a genuine 2-region streaming
pair (`cluster.yaml` primary + `replica-cluster.yaml` follower).

## The fix (chart-render only, small + safe)
Set `spec.cnpgPair.{name,namespace}` to the SAME `instanceName`/`namespace` the
label and both `Cluster` halves already stamp, so `cnpg.FindPair` (label
`catalyst.openova.io/cnpg-pair=<name>` + `openova.io/cnpg-role`) resolves the
local primary and the #5311 `pg_stat_replication` probe engages on the true
2-region topology (the region-b replica CR is not in the primary API). Mirrors
`platform/cnpg-pair/chart/templates/continuum.yaml`.

Why it needs nothing else:
- **No new credential** — the CNPG-provisioned `<primary>-replication` Secret
  (`tls.crt`/`tls.key`) exists for every `Cluster`; that Secret `get` is the
  probe's only credential access (`pgxprober.go`).
- **No NetworkPolicy carve-out** — unlike bp-cnpg-pair (which needed #5337),
  this chart's `-allow-replication-to-primary` policy already admits **every
  namespace** on 5432 (`namespaceSelector: {}` — shared-pg is a shared data tier
  whose consumers live in many namespaces), so the controller in
  `catalyst-system` already reaches `<primary>-rw:5432`.
- **No false-alarm path** — a probe error (wrong secret/host) is "no
  determination" (`standbyAvailable` stays UNSET, same as today, per
  `observeCNPGStandby` + `resolveStandbyPosture`); only a *successful* connect
  with zero `pg_stat_replication` rows degrades after the 30s grace window,
  which is the honest standby-absent signal #4901 exists to surface. A genuinely
  streaming CNPG replica always produces a walsender row (present/absent is
  robust to column-masking), so there is no false-negative for a connected
  standby.

## Per-family verdict (the other 7 of the 8 — deliberately NOT bundled)
| Continuum CR | mechanism | verdict |
|---|---|---|
| `cnpg-pair-bp-cnpg-pair-continuum` | cnpg-pair | already has `spec.cnpgPair`; NetworkPolicy-blocked → **fixed by #5337** |
| `dr-shared-pg` / `-b` / `-c` | cnpg-pair | **genuine wiring gap → THIS PR** |
| `dr-spine-gitea` / `-harbor` / `-keycloak` | cnpg-pair | **by design** — minted by the *generic* application-controller producer (`core/controllers/application/internal/controller/continuum.go` `renderContinuumCR`, zero app-name literal, keyed only off topology); these apps have no dedicated pair (gitea/harbor/keycloak are shared-pg **consumers** — they mint no Continuum of their own), so their DB-tier standby is surfaced by `dr-shared-pg`. Teaching the generic producer a per-app pair ref would duplicate that; it stays with the generic producer, not this PR. |
| `dr-spine-openbao` | raft-transition | **by design** — openbao raft, no `pg_stat_replication`; a cnpg standby-probe is genuinely N/A (UNSET is correct). |

## Tests / gates (all green locally)
- `platform/postgres/chart/tests/postgres-render.sh` — Case #4986 extended to
  assert the `cnpgPair` block (name/namespace); full gate PASS.
- `scripts/check-bootstrap-kit-pin-sync.sh --changed-only` — PASS (16a/16c/16d).
- `scripts/check-catalog-seed-lockstep.sh` — PASS (68 checked, 0 fail).
- `core/controllers/continuum`: `go build ./...` clean, `go test ./...` all ok.

Chart `0.2.13 → 0.2.14`, single-bump + lockstep (Chart.yaml + blueprint.yaml +
16a/16c/16d bootstrap-kit HR pins + catalog-seed) — same file set the prior
bp-postgres bump (#5228) touched; no umbrella bump required.

**Validates on the next fresh 2-region prov:** `standbyAvailable=True` surfaces
on the `dr-shared-pg*` Continuums with a live streaming standby (or an honest
Degraded if the shared-pg standby is genuinely absent). Refs #5311 #4901 #4987 #5337.


emrahbaysal:
## Pre-merge triage 2026-08-01 — CONFLICTING

`gh pr list --json mergeable` reports this PR **CONFLICTING** against current `main`.
The conflict is in `clusters/_template/bootstrap-kit/` slot files — the chart-pin
manifests, which the deploy-bot advances continuously (859 such commits in the last 60
days alone).

A pin-file conflict on a PR this age usually means the pinned versions in the branch are
many releases behind, so a mechanical rebase can silently *downgrade* a slot. It needs a
deliberate re-pin against current `main` values, not `git rebase --continue`.

I have **not** rebased it: doing so unattended is how a stale pin lands, which is the
`chart_image_pin_drift` class. Flagging it so the merge window is not spent discovering
this.

**Recommendation:** re-pin against current slot values and re-run
`scripts/check-bootstrap-kit-pin-sync.sh`, or close if the fix has been superseded —
I found related work on `main` but nothing that conclusively supersedes this branch, so
I am not asserting it either way.

PR #5386, opened 2026-07-26
fix(ci): make the §854 guard's negative-test coverage tamper-evident

emrahbaysal:
## Gap

`scripts/check-no-nodeports.sh` correctly **excludes** `platform/kyverno-policies/chart/tests/` from its scan — those Services carry `type: NodePort` deliberately, as the "bad" inputs proving the §854 ClusterPolicy actually **denies** a NodePort (#5088/#5089). `kyverno-test.yaml` asserts `result: fail` against exactly those resources.

But the exclusion has nothing behind it. If the fixtures are ever converted to ClusterIP — a "cleanup" that gets proposed on the reasoning *"the repo still greps NodePort, convert them"* — then:

1. this scan still passes,
2. `kyverno-test`'s deny-assertions no longer exercise anything,
3. the §854 **Enforce** policy silently loses its only negative coverage,

with zero signal that protection was removed.

## Fix

Assert the fixtures are still NodePort. If none remain, fail loudly and name the consequence.

## Verified both directions

```
intact    → OK — §854 negative-test fixtures intact (still NodePort): …/resource-services.yaml
converted → FAIL: §854 NEGATIVE-TEST COVERAGE LOST: no 'type: NodePort' fixture remains …
            The forbid-nodeport-service policy's deny-assertions (kyverno-test.yaml,
            result: fail) now exercise nothing — the Enforce policy is unproven.
```

(The converted case was simulated by `sed`-ing the fixtures to ClusterIP, then restored — working tree clean.)

## Live audit that motivated this (2026-07-26, full inventory on #5348)

5 NodePorts on mothership, **all non-OpenOva**: `cinova/catalog-svc` (⛔ never-touch), `iogrid/proxy-gateway-socks5` (→ iogrid#844), and 3× ephemeral `cm-acme-http-solver-*` — which cert-manager creates as NodePort by default and **no chart produces**. Zero chart NodePorts repo-wide, and both OpenOva ClusterIssuers are dns01-webhook-only, so a Sovereign cannot generate a solver NodePort.

Refs #5348 #5088 #4765


emrahbaysal:
Residual-delta analysis against main as of 2026-08-03 (Phase B step 1 of the PATH TO 100 plan), so this is a one-word decision:

**Your gap is still real and still unguarded.** `scripts/check-no-nodeports.sh:132` carries the `platform/kyverno-policies/chart/tests/*` exclusion with nothing behind it — exactly as this PR describes. None of the merged successors covers it: #5518 hardened the scan against vacuous passes, #5578 split the advisory for cache-masked charts, and `check-kyverno-nodeport-exclusions.sh` (the #5564 work) asserts the POLICY text carries no name-glob bypass with a self-tested detector — the adjacent axis. Nothing asserts the deny-fixtures themselves remain `type: NodePort`. Converting them would still silently disarm the §854 negative coverage with zero signal, today.

**The only thing stale is the patch**, not the idea: your +33 was written against the 2026-07-26 version of a script that #5518/#5578 have since restructured, hence the conflict.

Reply with one word:
- **"re-land"** — we port your assertion onto today's script structure in a fresh PR (your design, credited), this PR stays open as the design record until that merges, then closes referencing it.
- **"rebase"** — you rebase this branch yourself and we merge it on green.

What we will NOT do without your word is close this as superseded — nothing supersedes it, and closing it would discard the last guard between a well-meaning cleanup and a disarmed §854.

Refs #5088 #5564 #5518 #960

PR #5455, opened 2026-07-27
fix(uat): count UAT rows by status column — ad-hoc tally overstated green by 23 rows (Refs #960)

emrahbaysal:
## The headline number was wrong, and it was wrong in the flattering direction

The UAT green percentage has been counted ad hoc — search each line for a verdict glyph in priority order, take the first hit. Evidence prose routinely mentions other verdicts, so a row like

```
| 191 | ... | ❌ | ...re-walk shows the earlier ✅ no longer holds... |
```

scores **GREEN** off the glyph in its *evidence* while its status cell says ❌.

Measured against the live ledger: **54 rows** carry a glyph in evidence that differs from their status cell.

| method | green | total | pct |
|---|---|---|---|
| whole-line scan (what was used) | 204 | 280 | 72.9% |
| **status column (correct)** | **181** | **281** | **64.4%** |

Overstated by **23 rows / 8.5 points**. The full corrected breakdown: ✅181 · ⚠️30 · ◑8 · ❌24 · ☐17 · ⛔19 · N/A 2, over 281 data rows (5 headers + 5 separators excluded — the old count silently folded some of those in, which is where 280 vs 281 came from).

This also corrects the open-row counts: **☐ is 17 and ◑ is 8**, not the 6 and 9 I had been reporting.

## Why a script rather than a careful one-off

The failure mode only ever inflates. A miscount that could err either way gets caught when someone notices the number went down; one that only ever rounds up is invisible and compounds. So the count belongs somewhere with guards, not in whichever shell pipeline a walker writes next.

`scripts/uat-tally.py` reads field 6 and nothing else, classifies headers and separators explicitly rather than by heuristic, and reports a glyphless status (`N/A`) as N/A — never as green. `--json` for tooling, `--check N` for CI, `--list <glyph>` to enumerate.

## Guards

`scripts/test_uat_tally.py` pins the exact inflating row shape and asserts the naive method still gets it wrong, so the regression cannot come back quietly:

```
naive whole-line scan -> ✅   (WRONG)
status-column scan    -> ❌   (correct)
```

Also covered: green rows citing a past ❌ stay green; headers and separators are never data; `N/A` is never green; and row-id families `G#`/`R#`/`M#` classify as data rows — numeric-only filters have dropped those before.

## Also in this branch

Two rows stamped with hw290 live evidence, **both staying non-green**:

- **G3** — the standby-absent precondition is confirmed (4 Continuum CRs, all Healthy, primary == standby `hw-me-east-215-rtz-prod`). That shows the branch is *tolerated*, not that a self-fire was *exercised*, so it stays partial. Recorded alongside it: `kubectl get continuum -A` returned empty while the fully-qualified `continuums.dr.openova.io` returned all 4 — an empty short-name result is not evidence of absence.
- **232** — openclaw is not deployed on hw290 at all (0 pods, 0 HelmReleases, 0 Services), so `/readyz` cannot be probed there. Environment gap, not a verdict.

Refs #960


emrahbaysal:
**On the red check.** `UAT predecessor-env drift guard` fails here, and it fails identically on `main` — verified by running the guard against `origin/main`'s `docs/ledger/UAT.md` with none of this branch's changes. It is not a regression from this PR.

Root cause is filed as #5456: the ledger H1 reads `(RESET 2026-07-26 — pending hw289)`, so the guard resolves no current env and flags all 181 green rows. The live env is actually hw290 (dep `31106b6aa4a7e8e7`, verified 2026-07-27 — nodes Ready, 7 Organizations, console/marketplace 200). Correcting the header drops the count from 181 violations to 20 genuine ones.

I drafted that header fix and then **reverted it out of this branch on purpose**. It pairs with a decision to flush 20 rows of predecessor-env evidence — a 7-point move in the headline and a judgement about which walks still count. That does not belong bundled into a PR about the counting method.

This PR changes how rows are counted. #5456 covers which rows should count.

PR #5483, opened 2026-07-29
fix(catalyst-api): read primaryRegion from status.placement, not a key the CR lacks (#5482)

emrahbaysal:
## The defect

App detail **Overview** rendered a *host cluster label* where a region belongs. One Application reported three different values on hw291:

| surface | value |
|---|---|
| Overview → PRIMARY REGION | `platform-bootstrap-owned-host` ← a host label, not a region |
| Topology tab | `hw-me-east-215-a-rtz-prod` |
| the Application CR | `me-east-215-a` |

The handler read `status.primaryRegion`. That key does not exist — the CR's status keys are `conditions`, `lastReconciledAt`, `observedGeneration`, `phase`, `placement`, and the region lives at **`status.placement.primaryRegion`**.

`NestedString` returns `ok=false` rather than erroring on a missing path, so the miss was **silent**: `resp.PrimaryRegion` stayed empty and `AppDetail.tsx:256` fell back to `appRegions[0]`.

So the Overview tab was not showing the wrong region — it was showing a different *kind of identifier* entirely, because the field it wanted was never populated. That is why this read as a data bug rather than a missing field.

## The fix

The read becomes a small helper so it can be tested. It prefers the nested path and keeps the top-level one as a fallback, so a CR that *does* set it is still honoured.

## Tests — proven in both directions

Four cases. Removing the nested read makes `TestApplicationPrimaryRegion_ReadsNestedPlacementPath` fail with exactly the empty value that caused the fallback.

The absent-region case is the one worth keeping honest: it asserts **empty**, not any value. Substituting a different kind of identifier for a missing one is the actual failure mode here, so the helper must return nothing and let the UI decide what to render.

Full `internal/handler` package green (167s).

## Related, not fixed here

`hw-me-east-215-a-rtz-prod` (Topology) vs `me-east-215-a` (CR) is a separate, milder inconsistency — a fully-qualified cluster name against a bare region id. Both are *correct* identifiers for the same region, unlike the Overview value, so it is on the issue checklist rather than in this change.

Refs #5482 #5422 #3969

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/application_primary_region_5482_test.go
+++ b/products/catalyst/bootstrap/api/internal/handler/application_primary_region_5482_test.go
@@ -1,0 +1,106 @@
+// Tests for #5482 — the App detail Overview rendered a HOST CLUSTER LABEL
+// where a region belongs, because the handler read a status key the
+// Application CR does not have.
+//
+// Live on hw291, one Application reported three different values:
+//
+//	Overview PRIMARY REGION   platform-bootstrap-owned-host   <- a host label
+//	Topology tab              hw-me-east-215-a-rtz-prod
+//	the CR itself             me-east-215-a
+//
+// The read was `status.primaryRegion`; the CR carries
+// `status.placement.primaryRegion`. NestedString returns ok=false rather than
+// erroring on a missing path, so the miss was silent and AppDetail.tsx:256
+// fell back to appRegions[0].
+//
+// Anti-theater: TestApplicationPrimaryRegion_ReadsNestedPlacementPath fails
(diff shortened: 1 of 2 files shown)


emrahbaysal:
Holding this merge deliberately, not stalling. It touches applications.go, and a walker is right now producing UAT evidence against the per-Organization / placement family that this handler serves. Merging mid-walk is what invalidated the SSO rows earlier today — openova-uat-bot correctly flipped eight of them back to UNVERIFIED after an unrelated merge, including one I had stamped green minutes before. Will merge as soon as that walker reports.

PR #5503, opened 2026-07-29
fix(org-controller): name the pending boundary honestly for host-tier Orgs (Refs #5502)

emrahbaysal:
## What

The org-controller's `Ready=False` **reason** was hardcoded to `VClusterProvisioning`
regardless of tier. Now it names the artifact the Org is actually waiting on, keyed off
the same `#4292`/`#4339` tier gate (`gitops.BoundaryIsVcluster`) that `vclusterReadiness`
and `readyOrgMessage` already use:

| tier | plans | pending reason |
|---|---|---|
| host-tier | `""` / `s` / `free` | `NamespaceProvisioning` (new) |
| vcluster | `m` / `l` / `xl` / `flexi` | `VClusterProvisioning` (unchanged) |

## Why it mattered

A host-tier Org authors **no** vCluster HelmRelease — `gitops.Render` omits `vcluster.yaml`
for that tier and `vclusterReadiness` correctly short-circuits to host-namespace readiness.
The `Message` was already honest ("host-tier Org namespace not yet Active (no vCluster HR is
authored for this tier...)"), but `Reason` — the machine-readable field operators,
`kubectl -o jsonpath`, and tooling filter on — contradicted it.

`#4813` fixed exactly this asymmetry for the `Ready=True` message. The failure reason was
left behind.

**Observed cost on hw291.** Org `uatcorp` (kind=customer, plan=`s` → namespace-isolated
**by design**) sat at `Ready=False reason=VClusterProvisioning`. The acceptance walk
followed the reason and spent its investigation on a vCluster:

- `kubectl get vclusters -A` → `No resources found`, both regions
- vcluster StatefulSets = 0, both regions
- `status.vcluster.phase=Pending`

All three are the **correct** state for that tier. The real missing artifact was the
`uatcorp` namespace. This is the same "a control that reports rather than does" class the
surrounding `Ready-over-absent-backing` guard was written to prevent — the control was
reporting the wrong subsystem.

## Blast radius — checked before renaming a machine-readable field

- `orgStateFromCR` (`products/catalyst/bootstrap/api/internal/handler/org_list_from_cr.go`)
  switches on `status.vcluster.phase` and the Ready condition **status**; it never reads
  `Reason`. Console/directory state is unaffected.
- Repo-wide grep for the literal: one code site (fixed), one test assertion
  (`organization_controller_test.go:577`) whose fixture authors 11 files including
  `vcluster` — a **vcluster-tier** Org, so it stays green — plus prose in two Kyverno
  policy templates and two test comments.

## Verification

- `go build ./...` clean, `go vet ./internal/controller/` clean.
- New test asserts **both** directions. A one-sided assertion would pass against a
  function returning either string unconditionally: checking only host-tier would accept a
  new lie for the vcluster tier, and checking only vcluster would have accepted the
  original hardcoded string and caught nothing.
- A second test pins reason and message to agree per-tier, so the honest-message /
  wrong-reason split cannot silently reappear on a future edit to either helper.
- **Both-directions proof run, not just asserted**: reintroduced the defect (`sed` the
  return value) → guard FAILS on `""`/`s`/`free` and on the reason-vs-message
  cross-check; restored → PASSES. Full `internal/controller` package suite green.

Refs #5502
Refs #4813
Refs #4292

### DIFF
--- a/core/controllers/organization/internal/controller/organization_controller.go
+++ b/core/controllers/organization/internal/controller/organization_controller.go
@@ -883,7 +883,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		readyCond.Message = readyOrgMessage(org.Spec.PlanSlug)
 	default:
 		readyCond.Status = "False"
-		readyCond.Reason = "VClusterProvisioning"
+		// #5502 (Refs #4813/#4292/#4339): name the pending reason off the SAME
+		// tier gate vclusterReadiness + readyOrgMessage use. vcMsg was already
+		// tier-aware here; the machine-readable Reason was not, so a host-tier
+		// Org waiting on its host NAMESPACE reported VClusterProvisioning and
+		// sent the diagnosis hunting a vCluster that is correctly never
+		// authored for that tier.
+		readyCond.Reason = pendingBoundaryReason(org.Spec.PlanSlug)
 		readyCond.Message = vcMsg
 	}
 	desired := orgapi.OrganizationStatus{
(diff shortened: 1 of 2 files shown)


emrahbaysal:
## Live runtime confirmation on hw291 (dep `2c2d746b578c636b`)

Queried directly against the running Sovereign — the condition **contradicts itself inside
a single object**:

```
$ kubectl get organizations.orgs.openova.io uatcorp \
    -o jsonpath='{.spec.planSlug} | {.status.vcluster.phase} | {Ready cond}'

s  |  Pending  |  False/VClusterProvisioning:
  "host-tier Org namespace not yet Active (no vCluster HR is authored for
   this tier; the host namespace is the boundary)"
```

The `reason` asserts the Org is waiting on a **vCluster**. The `message` — in the very same
condition — states that **no vCluster HR is authored for this tier and the host namespace is
the boundary**. Both are produced by the same `default:` branch: the message comes from the
tier-aware `vclusterReadiness`, the reason was hardcoded.

And the genuinely-missing artifact, confirmed:

```
$ kubectl get ns uatcorp
Error from server (NotFound): namespaces "uatcorp" not found
```

So the honest diagnosis is "the `uatcorp` **namespace** was never created" — while the
machine-readable field pointed the walk at vClusters, which for a `plan=s` Org are
*correctly* absent (`kubectl get vclusters -A` → `No resources found` in both regions,
vcluster StatefulSets = 0).

This is not a hypothetical: it is what sent the row-10/11/12 acceptance walk hunting the
wrong subsystem. The fix in #5503 makes the reason `NamespaceProvisioning` for the
host tier, so the reason and message finally agree.

Issue #5480, opened 2026-07-29
A16 sweep: per-region secret generation on a shared VIP — one root cause behind UAT rows 33/35/37/38, 13 generators carry the shape


emrahbaysal:
A read-only SSO walk on hw291 (2026-07-29) established that **UAT rows 33, 35, 37 and 38 are not four bugs — they are one**, and it is the A16 class already documented in `docs/PRINCIPLES.md`. A16's own prescription is the reason for this issue: *"four instances in one day is the repo-wide-enumeration trigger: sweep every generator, don't chase the next gate."* That sweep had not been run. It has now.

## The live mechanism, proven three ways

`bao` / `guacamole` / `newapi` / `auth` / `console`.hw291.omantel.biz all resolve to **one IP, 212.72.24.6**, which round-robins across **two region-local backends that share no session state**.

**OpenBao** — 12 fresh-TCP `curl /v1/sys/health` returned two different clusters, `vault-cluster-05d07921` (×5) and `vault-cluster-3226767b` (×7), **both `standby=false`**. Two leaders, not one Raft cluster. A token minted on one is unknown to the other.

**Guacamole** — a *single page load* split across regions: `POST /api/tokens` → 200 on `10.43.2.209` (region-B), then both `connectionGroups/ROOT/tree` GETs → 403 on `10.42.0.121` / `10.42.3.220` (region-A). Sessions live in per-pod `HashTokenSessionMap` over two separate `guacamole-pg`.

**newapi** — region-a `users` holds only catalyst-root; region-b also holds `sovereign_2` at role 100. Region-A logs `[sessions] ERROR! securecookie: the value is not valid`. Generation seam: `platform/newapi/chart/templates/credentials-secret.yaml:28-42` mints `SESSION_SECRET` / `CRYPTO_SECRET` via `randAlphaNum 64`, guarded only by `lookup "v1" "Secret" .Release.Namespace` — and `lookup` sees **only the rendering cluster**, so each region generates its own.

Note the sampling discipline: these were taken with `curl` in a shell loop (new TCP per call). A browser `fetch()` loop cannot detect this, because HTTP/2 pins every request to one connection — that trap produced a false REFUTED on #5459 earlier today.

## Two previously-recorded causes are falsified

[...]

PR #5506, opened 2026-07-29
fix(kyverno): grant cilium read + narrow the Enforce flip to the NodePort ban (Refs #5505)

emrahbaysal:
P0 regression I introduced in #5494. That PR was described as enforcing the
§854 NodePort ban; its diff actually flipped FIFTEEN policies from Audit to
Enforce. One of them, networkpolicy-present, resolves a cnpCount context via an
APICall to cilium.io/ciliumnetworkpolicies -- and Kyverno has no cilium.io
RBAC, so that call has never resolved. Under Audit the unresolvable context was
merely reported; under Enforce it began DENYING through the fail-closed webhook
validate.kyverno.svc-fail (failurePolicy=Fail, rules include namespaces), so
EVERY Namespace create on the Sovereign was rejected.

Control pair, same ServiceAccount, the policy's two APICalls:
  can-i get networkpolicies.networking.k8s.io -> yes   (npCount works)
  can-i get ciliumnetworkpolicies.cilium.io   -> no    (cnpCount fails)

Window: chart 1.0.48 deployed 11:47:54Z and Namespace/hw291-omantel-biz -- the
same shape as uatcorp -- was created 8s later. #5494 merged 18:56:24Z, chart
1.0.49 deployed 18:58:12Z, Namespace/uatcorp denied at 22:08:14Z. No namespace
has been created since 11:48:02Z. Blast radius was platform-wide: bootstrap-kit
wedged on Namespace/cnpg-system and cluster-autoscaler, org-tenants and all
three per-Org Kustomizations failing. It is also the upstream cause of a chain
diagnosed downstream this session -- no uatcorp namespace, so the org-controller
correctly holds Ready=False (#5502) while the console fabricates Active (#5501).

Two parts, both needed:

1. New aggregated ClusterRole grants Kyverno get/list/watch on
   ciliumnetworkpolicies + ciliumclusterwidenetworkpolicies, via the
   rbac.kyverno.io/aggregate-to-* labels Kyverno watches, so it survives a
   Kyverno upgrade without patching vendor roles. A policy whose context cannot
   resolve is not a guard, it is a reporter -- this one guarded nothing from the
   day it was written and Audit mode hid that.

2. The fourteen incidentally-flipped policies return to Audit.
   forbidNodePortService stays Enforce, which was the stated scope of #5494.
   The action map is now the seven policies that were Enforce before #5494 plus
   forbidNodePortService. Each of the others earns Enforce on its own evidence,
   with its context resolution proven first.

The guard I shipped with #5494 asserted the NodePort policy's own action and its
bootstrapMode behaviour, but never pinned WHICH OTHER policies changed action --
so a fourteen-policy blast radius passed a green gate. An action-flip guard has
to pin the whole map, not one entry.

Verified: helm lint clean, ClusterRole renders, the #5491 NodePort guard still
passes (forbidNodePortService remains Enforce), and the authoritative
tests/e2e/bootstrap-kit Go gate is green. Lockstep 1.0.49 -> 1.0.50 across
Chart.yaml, blueprint.yaml, catalog-seed display + source, and slot 27a.

Refs #5505
Refs #5491

### DIFF
--- a/clusters/_template/bootstrap-kit/27a-kyverno-policies.yaml
+++ b/clusters/_template/bootstrap-kit/27a-kyverno-policies.yaml
@@ -154,7 +154,7 @@ spec:
       # mixed-source pods (proxy-cache sidecars + openova-io plugin) —
       # un-wedges huawei-evs-csi + its 24-HR dependency cascade (#5017,
       # #4973 family).
-      version: 1.0.49
+      version: 1.0.50
       sourceRef:
         kind: HelmRepository
         name: bp-kyverno
--- a/platform/kyverno-policies/blueprint.yaml
+++ b/platform/kyverno-policies/blueprint.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     catalyst.openova.io/section: pts-3-3-security-and-policy
 spec:
-  version: 1.0.49
+  version: 1.0.50
   card:
     title: Kyverno Policy Library
     family: guardian
--- a/platform/kyverno-policies/chart/Chart.yaml
+++ b/platform/kyverno-policies/chart/Chart.yaml
@@ -326,7 +326,7 @@ type: application
 #   (helm.sh/hook annotation) — hooks get no Flux labels + Helm force-stamps
 #   managed-by:Helm, so a legit hook Job (bp-powerdns zone-bootstrap) was
 #   Enforce-denied on live upgrade, stranding the powerdns anycast NodePort.
-version: 1.0.49
+version: 1.0.50
 appVersion: "1.0.2"
 keywords: [catalyst, blueprint, kyverno, policy, compliance, audit]
 maintainers:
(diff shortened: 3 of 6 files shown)


emrahbaysal:
Live-validated on hw291 before merge, and the validation changed my understanding of the fix: the cilium-read grant ALONE is insufficient. With RBAC granted, the cnpCount context resolves and the policy then denies with its real rule — `Namespace contains no NetworkPolicy and no CiliumNetworkPolicy` — which is structurally unsatisfiable at CREATE, since a namespace is always empty when created. So the Audit revert in this PR is required, not tidiness. Also found a per-region split that explains why only half the platform broke: region-a was serving validationFailureAction=Enforce while region-b was still Audit (only region-a had picked up 1.0.49) — the same A16 divergence family as #5480. After applying both halves live, `kubectl create ns --dry-run=server` is admitted in BOTH regions and forbid-nodeport-service remains Enforce, so §854 stays enforced. Full evidence on #5505.

PR #5523, opened 2026-07-30
fix(org-controller): per-Org console listeners ride the LIVE apex ports + a silently-dropped listener can no longer read as provisioned

emrahbaysal:
Refs #5511

## The defect, measured

`Gateway/kube-system/cilium-gateway-console` on hw291 declared **8** listeners and published **6** in `.status.listeners`. The two absent ones were exactly the per-Org pair `console-https-uatcorp` / `console-http-uatcorp` — no error, no condition, no event anywhere.

```
https://agenity.uatcorp.omani.homes/  ->  000  x6 of 6
https://console.uatcorp.omani.homes/  ->  000  x8 of 8
https://console.hw291.omantel.biz/    ->  200        (same VIP, control)
```

The control is what makes this decisive: `agenity.uatcorp` was **fully healthy** behind that door — HelmRelease `Ready=True`, `bp-agenity-0` 1/1 Running, HTTPRoute present, DNS resolving to the console ELB — and still answered 000. Routing failed independently of workload health, and the apex console on the same VIP served 200.

## Root cause

`ensureConsoleOrgListener` appended the per-Org pair on the **pre-#4718 ports 31443/31080**. The console LoadBalancer forwards public 443/80 to the **live apex** `console-https`/`console-http` listener ports and to nothing else, so a per-Org listener on any other port receives no traffic at all.

This is not a new theory — it is the **same defect already fixed once in the catalyst-api twin** by #4732 item 2 (commit `8663202bc`), verbatim:

> Per-Org console listeners ride the SAME ports as the live apex console-https/console-http pair (SNI-differentiated) instead of the dead pre-#4718 31443/31080 constants — the console ELB only forwards to the apex ports (8443/8080), so any other pair receives no traffic.

That twin's test bans the literals outright (`org_console_tls_test.go:336`). The org-controller twin never received the fix, so **whichever of the two doors provisioned a given Org silently decided whether that Org's console worked** — even though the two are specified to render byte-identical resources (#4241).

A second, independent bug made it permanent: the append short-circuited on listener-**name** presence, so a pair already sitting on the dead ports was never corrected. Fixing the port derivation alone would have repaired new Orgs and left every existing customer door at 000 forever.

## #5341 wildcard-collision hypothesis: REFUTED

- **#5341 is already fixed in this very gateway.** `platform/sovereign-tls-vars/chart/templates/configmap.yaml` renders the console listeners with **exact 2-label hostnames** (`console.`/`api.`/`marketplace.<fqdn>`), explicitly never `*.<fqdn>` — the gateway therefore carries **no wildcard listener** for the per-Org wildcard to collide with.
- The 6 published listeners are exactly `3 hostPrefixes x 2 protocols`, accounting for all six with zero wildcards.
- `*.uatcorp.omani.homes` and the apex `*.hw291.omantel.biz` hosts are under different registered domains — no SNI overlap is even possible.
- Symptom shapes differ: #5341 produced ~50% **non-deterministic envoy 404s** (a served response). This was **deterministic 000** on 6/6 and 8/8 with a 200 control on the same VIP.

## Part 1 — make the listeners admit

`core/controllers/organization/internal/controller/tenant_console_tls.go`

- `consoleApexListenerPorts` derives the pair's ports from the live apex listeners **in the same read the append is written back onto**, so port and listener set are consistent by construction. Falls back to the #4718 canonical 8443/8080 — matching the catalyst-api twin's fallbacks byte-for-byte — when the apex cannot be read. The 31443/31080 constants are gone.
- The append now **repairs drift in place** instead of short-circuiting on name presence, so the wrong-port pairs already live on affected Sovereigns are healed rather than frozen. Drift is a **subset match on the fields we own** (`listenerSatisfies`), so an apiserver-defaulted field never reads as drift and rewrite-every-reconcile cannot hot-loop the controller. Other Orgs' listeners and the apex pair are never touched.

## Part 2 — a dropped listener can never be silent again

`core/controllers/organization/internal/controller/provisioning_postconditions.go`

#5395's verifier read `spec.listeners` **only**, which is why it passed green on hw291: the pair *was* in spec. Presence in spec is not serving. `verifyProvisioned` now reads back `status.listeners` too and refuses `Ready` over:

- **a count mismatch** — any listener accepted into spec and absent from a published status, named individually with both counts (`ACCEPTED 8 listeners into spec but published only 6 in status — silently dropped: ...`);
- **port drift** against the live apex pair, which fires even when admission succeeded, because an admitted listener on a dead port still serves nothing.

An **empty** `status.listeners` is reported `Unverifiable`, not `Missing` — the bootstrap window is absence of evidence, and reporting it as a red would flag every pool Org on the Sovereign at once. That gate is also what stops the status check from passing **vacuously** on a Gateway that has published nothing at all.

Part 2 is the durable half: it surfaces the drop regardless of cause, including if the status omission turns out to be an upstream Cilium constraint.

## Gates

```
$ cd core/controllers/organization && go build ./... && go vet ./... && go test ./...
=== go build ===  OK
=== go vet ===    OK
ok  github.com/openova-io/openova/core/controllers/organization/internal/controller
ok  github.com/openova-io/openova/core/controllers/organization/internal/gitops
ok  github.com/openova-io/openova/core/controllers/organization/internal/iacbootstrap

$ cd tests/e2e/bootstrap-kit && go test -count=1 ./...
ok  github.com/openova-io/openova/tests/e2e/bootstrap-kit  0.453s
```

No chart, chart version, or bootstrap-kit file is touched; the lockstep gate was run as insurance and is green.

## FAIL-then-PASS, executed not asserted

Each guard was verified by reintroducing the defect, observing the failure, restoring, and observing the pass.

**Proof A — hardcode the dead ports back into `consoleApexListenerPorts`:**

```
--- FAIL: TestEnsureConsoleOrgListener_5511_RidesLiveApexPorts
    listener "console-https-acme" port: got 31443, want 443 (the live apex port ...)
    listener "console-https-acme" is back on the dead pre-#4718 port 31443 — this is the #5511 defect
    listener "console-http-acme" port: got 31080, want 80
--- FAIL: TestEnsureConsoleOrgListener_5511_ApexAbsentUsesCanonicalFallback
    console-https-acme port: got 31443, want the #4718 canonical fallback 8443
--- FAIL: TestEnsureConsoleOrgListener_5511_HealsDeadPortsInPlace
--- FAIL: TestVerifyProvisioned_5511_PortDriftBlocksReady
--- FAIL: TestVerifyProvisioned_5511_EmptyStatusIsUnverifiableNotMissing
--- FAIL: TestVerifyProvisioned_5511_AdmittedPairGoesGreen
```

**Proof B — restore the presence-only short-circuit** (isolates the self-heal guard alone):

```
--- FAIL: TestEnsureConsoleOrgListener_5511_HealsDeadPortsInPlace
    a pair sitting on the dead 31443/31080 ports MUST be reported as changed —
    a presence-only no-op here is the bug that made #5511 permanent
```

**Proof C — revert the readback to spec-only** (the silence itself):

```
--- FAIL: TestVerifyProvisioned_5511_SilentlyDroppedListenerBlocksReady
    a listener present in spec and silently ABSENT from a published status must NOT
    read as provisioned — got Missing=[] Unverifiable=[]
```

`Missing=[] Unverifiable=[]` is the #5511 shape verbatim: completely silent.

**Restored — all 14 guards pass:**

```
--- PASS: TestEnsureConsoleOrgListener_5511_RidesLiveApexPorts
--- PASS: TestEnsureConsoleOrgListener_5511_HealsDeadPortsInPlace
--- PASS: TestEnsureConsoleOrgListener_5511_ApexAbsentUsesCanonicalFallback
--- PASS: TestVerifyProvisioned_5511_SilentlyDroppedListenerBlocksReady
--- PASS: TestVerifyProvisioned_5511_AdmittedPairGoesGreen
--- PASS: TestVerifyProvisioned_5511_EmptyStatusIsUnverifiableNotMissing
--- PASS: TestVerifyProvisioned_5511_PortDriftBlocksReady
--- PASS: TestReconcile_5395_ConsoleListenerMissing_MustNotReadReady
--- PASS: TestReconcile_5395_SilentListenerSkip_MustNotReadReady
--- PASS: TestReconcile_5395_ListenerAppendedThenVerified
--- PASS: TestReconcile_5395_HardCappedPlanWithoutQuota_MustNotReadReady
--- PASS: TestReconcile_5395_FlexiWithoutQuotaIsNotAGap
--- PASS: TestReconcileTenantConsoleTLS_IssuesCertAndAppendsListener
--- PASS: TestReconcileTenantConsoleTLS_NoopWhenNoParentDomain
ok  github.com/openova-io/openova/core/controllers/organization/internal/controller
```

### Test discipline notes

- **No vacuous assertion.** Every port assertion carries a positive control that the append actually produced listeners (`spec.listeners` count checked before the port is read), and the drift fixture asserts the wrong ports are present *before* the heal, so a green cannot come from a fixture that never carried the defect.
- **Count-based, not name-based.** The part-2 guard asserts on `len(status.listeners)` vs `len(spec.listeners)` with the pair declined, and its converse asserts equality on the green path — the bug is a silent count mismatch, not a single missing name.
- The #5395 fixture now simulates the Gateway controller's admission step (`admitGatewayListeners`), because a live Gateway always publishes `status.listeners`. The fixture became a **stricter** model of reality, not a looser one — it now exercises the full append-then-admit loop.

### Not verified here

Whether Cilium *also* declines to publish status for a listener on an unforwarded port is a Cilium-internal behaviour I could not confirm — no hw291 kubeconfig is present on this host, and the task is code-only. The port defect is proven from the writer and fully explains the measured 000; the status omission is consistent with it. Part 2 deliberately does not depend on that answer: it surfaces the drop whatever the cause.

No NodePort is introduced or relied upon anywhere in this change — the per-Org listeners ride the same ports the DIRECT-served console gateway already binds.

### DIFF
--- a/core/controllers/organization/internal/controller/provisioning_postconditions.go
+++ b/core/controllers/organization/internal/controller/provisioning_postconditions.go
@@ -128,6 +128,18 @@ func (p provisioningPostconditions) message() string {
 //     engages the tenant-networking up-path (a pool parentDomain). This is the
 //     artifact whose absence produced NoMatchingListenerHostname on all six of
 //     gamma-corp's routes (#5395 symptom A).
+//  4. That the SAME pair is ADMITTED — present in the Gateway's
+//     `status.listeners`, and bound to the ports the console LoadBalancer
+//     actually forwards to (#5511). Presence in `spec` is not serving. On
+//     hw291 `cilium-gateway-console` accepted 8 listeners into spec and
+//     published only 6 in status; the two it silently declined were exactly
+//     the per-Org pair, with no error, no condition and no event anywhere, so
+//     `agenity.uatcorp.omani.homes` answered 000 on 6 of 6 probes while its
+//     HelmRelease, Pod and HTTPRoute were all healthy and the apex console on
+//     the same VIP served 200. A listener accepted into spec and absent from
+//     status is, from the operator's side, indistinguishable from one that was
+//     never configured — so it must be named in status, loudly, the same way
+//     an absent one is.
 //
 // The namespace itself is NOT re-probed here — vclusterReadiness already gates
 // on it and verifyProvisioned is only consulted once that returned ready, so a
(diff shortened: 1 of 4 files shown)


emrahbaysal:
## Live-object evidence from hw291 — the port theory is NOT this environment's cause

PR #5523 diagnosed the drop as listeners appended on the pre-#4718 ports `31443/31080`. I read the
live Gateway (the PR author had no hw291 kubeconfig) and the ports are **correct**:

```
console-https          8443  console.hw291.omantel.biz        ADMITTED
console-http           8080  console.hw291.omantel.biz        ADMITTED
api-https              8443  api.hw291.omantel.biz            ADMITTED
api-http               8080  api.hw291.omantel.biz            ADMITTED
marketplace-https      8443  marketplace.hw291.omantel.biz    ADMITTED
marketplace-http       8080  marketplace.hw291.omantel.biz    ADMITTED
console-https-uatcorp  8443  *.uatcorp.omani.homes            **DROPPED**
console-http-uatcorp   8080  *.uatcorp.omani.homes            **DROPPED**

admitted ports = [8080, 8443]      dropped ports = [8080, 8443]
```

So two premises in the PR are false **on this env**:

1. The dropped pair is not on the dead ports — it is on the same ports as every admitted listener.
2. The #5341 refutation rested on "the console gateway carries **no wildcard listener at all**". It
   carries **two**, and they are exactly the dropped pair.

**The discriminator the live data actually supports:** the two dropped listeners are the ONLY wildcard
hostnames on a gateway where all six admitted listeners use exact 2-label hosts. That is closer to the
wildcard-handling hypothesis than the refutation allowed. Ownership is also not the org-controller —
the Gateway is Flux-owned (`kustomize.toolkit.fluxcd.io/name: sovereign-tls`), so a different writer
produced these listeners with correct ports.

`Programmed=False (AddressNotAssigned)` is present and is the expected §854 hostPort artifact — the
working console listener carries it too. Not a finding.

## What this does NOT invalidate — two independent things in one PR

**Part 1 fixes a real, currently-unfixed defect, just not #5511.**
`core/controllers/organization/internal/controller/tenant_console_tls.go:129-130` genuinely still
hardcodes `31443/31080`, while its sibling
`products/catalyst/bootstrap/api/internal/handler/org_console_tls.go:127` documents having moved off
exactly those constants. So the twin-writer drift is real and worth landing on its own merits — and it
is plausibly the mechanism behind the **Door A vs Door B divergence** recorded on UAT row G7, where the
create-Org form's displayed default disagrees with the authoritative gate.

**Part 2 is unconditionally correct and is the durable win.** #5395's verifier read `spec.listeners`
only, which is precisely why it passed green on hw291 while two listeners were missing. Reading
`status` and refusing Ready on a count mismatch catches this class *whatever* the cause, and treating
an empty status as `Unverifiable` rather than `Missing` stops it passing vacuously. Same
guard-cannot-fail family as #5518 and #5473.

## Ask

Amend the root-cause narrative before merge, or split: Part 2 (+ Part 1 as its own twin-drift fix,
referencing G7) can land; the #5511 cause should read "wildcard-only listener silently unadmitted on a
Flux-owned gateway", still open pending the Cilium-side mechanism. The PR's own caveat — that whether
Cilium declines to publish status for such a listener was never traced — remains the open question, and
Part 2 was deliberately built not to depend on the answer.

PR #5539, opened 2026-07-31
fix(catalog): make catalog generation deterministic so the committed output can converge (#5520)

emrahbaysal:
## Root cause

`build-catalog.mjs` stamped the output with `generatedAt: new Date().toISOString()`, so a rebuild of **identical input** always rewrote `blueprints.json`. The generated files could therefore never converge with the generator:

- every PR that ran a real build carried a spurious diff
- authors either committed unrelated churn or reverted by hand
- the **revert path was the common one**, so the committed copies fell further behind on every cycle

That is how the 508-line drift accumulated unnoticed, and how it came to fail `StepComponents` (`expected [...] to include 'axon'`) while reading as "pre-existing and unrelated" noise.

Two distinct causes were tangled together, which is worth separating because only one of them is a bug:

| cause | status |
|---|---|
| non-deterministic `generatedAt` — permanent churn engine | **the defect**, fixed here |
| genuine content drift (508 lines) | already absorbed by #5526 (`8523ecf33`) |
| genuine content drift (bp-gitea 1.2.48 → 1.2.49) | real, committed here |

## Fix

`generatedAt` now marks **when the catalog content last changed**: the generator preserves the prior stamp when the rest of the payload is byte-identical.

This is deterministic *and* more truthful. Its consumers — `catalog.GeneratedAt()` feeding `/sovereign/apps` and the per-Org applications endpoints — want to know when the catalog they are serving last *changed*; a rebuild of unchanged input did not change it. Deleting the field was not available: three Go call sites surface it in API responses.

Proven by running the generator three times — runs 2 and 3 are byte-identical to run 1.

## Gate

`check-catalog-generated-drift.yaml` regenerates and fails on any diff, so this cannot silently re-accumulate.

It also asserts **determinism directly** — two consecutive runs must be byte-identical. That check is the reason such a gate could not have existed before: against the old wall-clock stamp it would have failed on every single run, which is exactly why the drift had no guard.

## Verification

- generator run ×3 → byte-identical output after the first
- both gate legs simulated locally: the determinism leg passes; the in-sync leg correctly **failed** on the pending bp-gitea drift before it was committed, and passes after
- workflow YAML parses

Refs #5520

### DIFF
--- a/.github/workflows/check-catalog-generated-drift.yaml
+++ b/.github/workflows/check-catalog-generated-drift.yaml
@@ -1,0 +1,89 @@
+name: check — generated catalog is in sync (#5520)
+
+# The committed `blueprints.json` + `catalog.generated.ts` are BUILD OUTPUT:
+# `npm run build` in products/catalyst/bootstrap/ui runs a `prebuild` step
+# that regenerates them from the blueprint sources. When the committed copies
+# drift from what the generator produces, every PR that runs a real build
+# carries a spurious diff, and authors either commit unrelated churn or revert
+# it by hand. The revert path was the common one, which is exactly how the
+# 508-line drift in #5520 accumulated unnoticed — and it broke StepComponents
+# tests that then read as "pre-existing and unrelated".
+#
+# This gate makes the drift impossible to accumulate silently: regenerate, and
+# fail if the working tree changed. It is the mechanical enforcement of
+# "generated files are committed in sync", the same shape as the repo's other
+# drift guards (cluster-template-drift, check-catalog-seed-lockstep).
+#
+# Prerequisite that makes this gate meaningful: the generator is DETERMINISTIC.
+# `generatedAt` used to be a wall-clock stamp, so a rebuild of identical input
+# always produced a diff and a gate like this could never have gone green.
+# build-catalog.mjs now preserves the prior stamp when the content is
+# unchanged, so `generatedAt` marks when the catalog CONTENT last changed.
+
+on:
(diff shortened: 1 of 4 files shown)


emrahbaysal:
## The red check is not from this change

`integration (kind + gitea)` failed on a single test, unrelated to catalog generation:

```
--- FAIL: TestWipeDeployment_HuaweiOperatorCredsEnvFallbackAvoids400 (60.01s)
FAIL  .../internal/handler  456.397s
```

`60.01s` is not an assertion failure — it is the timeout in that test's last
line, `waitForWipeDone(t, dep, 60*time.Second)`, waiting on the async wipe
goroutine.

Locally, on a branch that carries this change plus the #5542 handler work:

```
--- PASS: TestWipeDeployment_HuaweiOperatorCredsEnvFallbackAvoids400 (0.18s)
ok   .../internal/handler   0.198s
```

and the whole module is clean — `internal/handler ok 240.238s`, 28 packages,
zero failures.

**0.18s locally against a 60s timeout in CI** is a ~300x gap on an async wait,
so the wipe goroutine is not completing under CI load rather than the wipe
logic being wrong. Nothing in this PR touches the wipe path: the change is
`build-catalog.mjs` (preserve the prior `generatedAt` when the payload is
byte-identical) plus a new determinism gate workflow.

Re-running the job. If it recurs, the durable fix belongs on that test's wait
(it needs a settle condition rather than a fixed 60s budget) and gets its own
issue — not a change to catalog generation.

PR #5544, opened 2026-07-31
fix(ci): the NodePort guard silently skipped 19% of charts while reporting full coverage (#5512)

emrahbaysal:
## The guard reported coverage it did not have

Phase 2 of `scripts/check-no-nodeports.sh` renders every chart and scans the output for **values-driven** NodePorts — the ones the Phase 1 source scan structurally cannot see. A chart that failed to render offline was `WARN`-skipped, and the run still ended with:

```
OK: zero NodePorts across sources + rendered charts (#4765)
```

Measured on this tree:

| | |
|---|---|
| charts rendered + scanned | **72** |
| charts silently skipped | **17** |
| fleet without render coverage | **19%** |

A values-driven NodePort inside any of those 17 passes **both** phases, and the guard still prints OK. On the founder's most emphatic rule — *"YOU CAN NEVER EVER USE NODEPORT EVEN FOR TESTING PURPOSE"* — the green meant "the check did not look", not "the thing is absent".

This is the same shape #5512 reports for the row-77 mothership-host search, which is why it belongs to that issue: an absence-assertion that reports success when it simply failed to evaluate.

## The 17 skips are not chart defects

Every one fails a deliberate fail-loud values guard. Representative:

```
Error: execution error at (bp-anthropic-adapter/templates/deployment.yaml:40:21):
adapter.image.tag MUST be SHA-pinned by the per-cluster overlay (never :latest, never empty)
```

That is correct chart design. The defect is that the resulting skip was **silent and unbounded**.

## Change

- **`RENDER_SKIP_ALLOWLIST`** records the 17 charts known not to render with default values. A chart **outside** the list that fails to render now **fails the guard** and prints the helm error — a new blind spot can no longer open quietly.
- A chart **inside** the list that starts rendering is reported, so the allowlist cannot rot. Same anti-rot reasoning as the existing Phase 0a pattern self-test.
- The summary now states **actual** coverage — how many rendered, how many are source-scan-only. When `helm` is absent it says plainly that the run does not cover values-driven NodePorts, rather than claiming rendered-chart coverage that never ran.

## Verified both directions

| Tree | Result |
|---|---|
| unchanged | exit **0** — `72 chart(s) rendered and scanned, 17 allowlisted chart(s) covered by the source scan only` |
| `platform/postgres/chart` removed from allowlist | exit **1** — `did NOT render offline and is not in RENDER_SKIP_ALLOWLIST` |

The negative case is what matters: the new gate **bites** rather than passing on nothing.

## Next

Shrinking the 17 — making those charts renderable under a test-values overlay — is the real coverage win and is worth its own change. This one stops the gap from being invisible first.

Refs #5512 #4765 #960

### DIFF
--- a/.github/workflows/check-no-nodeports.yaml
+++ b/.github/workflows/check-no-nodeports.yaml
@@ -51,7 +51,14 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v4
         with:
-          version: v3.16.0
+          # >= 3.16.2. helm 3.16.0 has two dependency-resolution regressions
+          # that make Phase 2 unable to fetch a subchart on a clean checkout —
+          # observed on this runner, both absent on 3.16.3:
+          #   "repo charts not found" (no charts/ directory yet), and
+          #   "non-absolute URLs should be in form of repo_name/path_to_chart"
+          # They hit 11 dependency-bearing charts at once and read as NodePort
+          # guard failures, which they are not.
+          version: v3.16.3
 
       - name: Run zero-NodePort guard
         run: bash scripts/check-no-nodeports.sh
--- a/scripts/check-no-banned-words.sh
+++ b/scripts/check-no-banned-words.sh
@@ -38,6 +38,39 @@ declare -a BANNED=(
   '\bblocker[s]?\b'
 )
 
+# ─── Pattern self-test (vacuity guard) ───────────────────────────────────
+#
+# An absence-assertion reports "clean" both when the thing is absent AND
+# when the check stopped working. If an edit breaks one of the BANNED
+# regexes, this guard silently matches nothing and passes every PR forever
+# — green meaning "did not look" rather than "not present" (#5512).
+#
+# scripts/check-no-nodeports.sh has carried this protection since #4765
+# (its Phase 0a); this guard did not. Each pattern must still match a
+# known-banned sample, and a clean sentence must still match nothing (so a
+# pattern widened to `.*` is caught too, rather than failing every PR).
+declare -a BANNED_SELFTEST=(
+  'ship the MVP now'
+  'the second iteration landed'
+  'that is out of scope'
+  'the blocker is upstream'
+)
+CLEAN_SAMPLE='gateway renders direct on the shared VIP and the chart publishes'
+
+for i in "${!BANNED[@]}"; do
+  if ! grep -iqP "${BANNED[$i]}" <<<"${BANNED_SELFTEST[$i]}"; then
+    echo "SELF-TEST FAIL: pattern ${BANNED[$i]} no longer matches its sample" >&2
+    echo "  sample: ${BANNED_SELFTEST[$i]}" >&2
+    echo "  The banned-words guard would pass every PR. Fix the regex." >&2
+    exit 2
+  fi
+  if grep -iqP "${BANNED[$i]}" <<<"${CLEAN_SAMPLE}"; then
+    echo "SELF-TEST FAIL: pattern ${BANNED[$i]} matches a CLEAN sentence" >&2
+    echo "  It is too broad and would fail every PR." >&2
+    exit 2
+  fi
+done
+
 # Allowed-context exemptions — paths where the banned words are
 # documented BY DEFINITION (this script's own help text, the principles
 # section that ENUMERATES the banned words, the ledger rows that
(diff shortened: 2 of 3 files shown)


emrahbaysal:
## Root-caused and fixed — the guard could not pass CI, and the reason was not the guard's logic

CI failed on the same SHA (`c1f167f99`) that passed locally. That divergence was the finding.

`Phase 2`'s comment asserted charts *"render OFFLINE from their vendored `charts/*.tgz`"*. But
`.gitignore:6` ignores `platform/*/chart/charts/`, and **zero** of the 67 dependency-bearing charts
have a committed `Chart.lock`. A clean checkout has neither, so the vendored tarballs exist only on
a machine that has already run a dependency update.

CI's actual error, from the job log:

```
FAIL: chart platform/alloy/chart did NOT render offline and is not in RENDER_SKIP_ALLOWLIST
Error: ... found in Chart.yaml, but missing in charts/ directory: alloy
```

The #5512 ratchet was doing exactly its job — it just had nothing to render. All 67 would have
failed the same way; `alloy` is simply first alphabetically.

### Fix

Offline-first is preserved and still tried first. A bounded `helm dependency update` fallback fires
**only** on the missing-dependency signature, so a chart failing for any other reason (a fail-loud
values guard, a template bug) still reaches the ratchet rather than being masked by a network
round-trip. Verified against a helm config with **zero repos registered** — what a runner looks
like — where helm resolves the URL straight from `Chart.yaml`:
`Getting updates for unmanaged Helm repositories...`. `NO_NETWORK=1` suppresses it for air-gapped
runs.

### Verified the way CI sees it

A detached worktree at the fix commit (so no gitignored tarballs) plus an isolated empty helm
config:

```
OK: zero NodePorts — sources clean; 71 chart(s) rendered and scanned,
    18 allowlisted chart(s) covered by the source scan only (#4765).
     dependencies fetched on demand this run: 56 (NO_NETWORK=1 disables the fetch)
exit=0
```

**56 charts needed a fetch on a clean checkout.** That count is the whole point, and it is why the
summary now prints it even at zero — "0 because everything was vendored" and "0 because the
fallback never fired" are the two states whose confusion caused this.

### One chart could not be recovered, and it is a real defect

`platform/sealed-secrets/chart` declares `https://bitnami-labs.github.io/sealed-secrets`, which
answers **404**. The upstream project is alive (`v0.38.4`) — Bitnami moved chart distribution. It
renders on dev machines only from a stale cached tarball.

Filed as **#5563** and allowlisted here with that reason stated inline, rather than repointed:
choosing a replacement registry for sealed-secrets lands on every Sovereign and on what Harbor
prewarms at cutover, so it should be a deliberate call, not a side effect of a CI fix. Phase 1
still scans its sources, so §854 coverage is unchanged.

Commits: `db5dfef3a` (fallback), `6e11a8ab9` (allowlist + #5563).

emrahbaysal:
Follow-up: the CI red was **one truncated sweep**, not a series of separate chart faults.

Two `set -e` hazards, one introduced by my fallback and one pre-existing in this PR's own ratchet:

1. `depout="$(cmd)"; deprc=$?` — the assignment carries the command's exit status, so a failing fetch killed the script before `deprc=$?` ran.
2. Worse, and already here: the ratchet's diagnostic line `helm template ... | head -3 >&2` runs **only** when a chart failed to render, so helm exits non-zero by construction; `set -o pipefail` propagates it and `set -e` kills the sweep. The guard printed the first offending chart and stopped — **with the same rc=1 a complete run gives**, so the truncation was invisible.

That is why CI showed only `alloy`, then only `bp-mgmt-vcluster`: each run died at its first casualty and never reached the rest.

Verified with a probe chart pointing at an unresolvable repository, placed **first alphabetically** so truncation cannot hide behind ordering. Before: zero further charts scanned. After: the diagnostic prints `NOTE: ... dependency fetch failed (rc=124, timeout=10s)` with helm's own tail, **72 further charts are scanned**, and the run still fails at the end with the full remediation block.

Clean-checkout simulation on `12c0e0f7c` (detached worktree, no gitignored tarballs, empty helm config):

```
OK: zero NodePorts — sources clean; 71 chart(s) rendered and scanned,
    18 allowlisted chart(s) covered by the source scan only (#4765).
     dependencies fetched on demand this run: 56
exit=0
```

PR #5546, opened 2026-07-31
docs(uat): walk M1 + G5 live on the mothership janitor — 2 stamped, R1 held vacuous, 2 overclaims corrected

emrahbaysal:
## What this corrects about the walk wall

With hw291 wiped I had been treating the whole ledger as env-gated. That was too broad. **Four rows are mothership-scope** and were parked at *"not observable from child kubeconfig"* — not because the assertion was hard, but because prior walks ran from a Sovereign kubeconfig. The mothership is up (`catalyst-api 1/1`, 15h), so those rows are walkable right now.

Walked **read-only** against `catalyst-api-7c558f4d98-89scx` (ns `catalyst`, log read only, zero mutation). One complete janitor pass:

```
[JANITOR] pass complete durationMs=7029 destructive=false reaped=0
          reapErrors=0 failedInfraPreserved=0
```

## Stamped — 2 rows

| row | assertion | why it passes |
|---|---|---|
| **M1** (#4466) | log-only/dry-run for a full live cycle before any destructive sweep | proven at **both** levels — pass-level `destructive=false`, and per-item `would-reap (log-only; set CATALYST_JANITOR_DESTRUCTIVE=true to act)` |
| **G5** (#4466) | the dry-run full-cycle observation | same pass, observed end to end (7029 ms, 7936 janitor lines in window) |

## Held — 1 row, deliberately

**R1** (#4454) — *"orphan-sweep does not reap a `ready` Sovereign"* — stays ⛔.

The pass does show `reaped=0`. But hw291 is wiped, so the registry holds **no `ready` Sovereign for the janitor to spare**. Zero-reap against zero candidates is a **vacuous** observation: it would read identically if the protect-side inversion were broken. Stamping it would be exactly the failure this ledger exists to prevent. It re-walks on hw292.

M1 and G5 are stamped from that same pass precisely because their assertion *is* non-vacuous when there are zero candidates.

## Corrected — 2 prior evidence notes that overstated a log-only pass

- **G5** had read *"safely GC-ing 19 orphan EVS + 1 orphan keypair"*
- **G4** had read *"19 EVS reclaimed"*

Under `destructive=false` **nothing is reclaimed**. The sweeps increment inside the non-destructive branch:

```go
// internal/providers/huawei/provider.go:2837-2840
if !destructive {
    progress("... would-reap (log-only; ...)")
    reaped++          // counted without deleting
    continue
}
```

and `janitor.go:360` emits that count as `orphanOBSBucketsDeleted`. Today's pass reports `orphanOBSBucketsDeleted=61` and `orphanKeypairsDeleted=2` having deleted **zero**. Filed as **#5545**; the ledger notes are fixed here so no future walker inherits the overclaim.

## Gates

- `uat-drift-guard.py` → **OK**. Worth noting *why* it accepts these: `mothership-<date>` is not an `hwNNN` token, so mothership-scope rows sit correctly outside the each-new-env-flushes rule.
- `test_reset_uat_assertion.py` → **OK**.

Refs #4466 #4431 #4454 #5545 #960


emrahbaysal:
Closed as fully superseded — verified row-by-row against current main (2026-08-02), not assumed:

- **M1 and G5**: byte-identical between this branch and main — the stamps landed via the direct-to-main walk commits and stand.
- **R1**: main's version strictly supersedes this branch's. It carries the same VACUITY adjudication (zero-reap against zero candidates proves nothing; re-walks on hw292) AND content this branch lacks: the reason-correction (janitor IS observable from the mothership kubeconfig) and the protect-branch silent-log observation that blocks a future walk (tied to #5545).

One fact has changed since this walk was taken: the mothership catalyst namespace is now at 0/0 (#5573/#5558), so the janitor pass this PR observed is not currently reproducible. That does not retract the point-in-time evidence — the pass happened and was read live — but the next observation waits on the scale-up.

Nothing in this branch is lost. Refs #4466 #4454 #5545 #5573 #960

PR #5548, opened 2026-07-31
feat(ci): cluster-side §854 guard — catches a live node port that renders clean in the chart (#5348)

emrahbaysal:
## The gap this closes

`scripts/check-no-nodeports.sh` scans repo sources and `helm template` output. It is the right guard for *"did someone write a NodePort into the catalog"* — and it **cannot** see a live Service that drifted from a spec which still renders correctly.

#5348 is exactly that case, and it sat open behind a green guard:

```
openova-system/powerdns-anycast
  type                         : LoadBalancer
  allocateLoadBalancerNodePorts: False     ← anti-nodePort flag SET
    port 53/UDP  nodePort=32015            ← …and the ports open anyway
    port 53/TCP  nodePort=31425
```

The chart set the flag, omitted `nodePort`, and `fail`ed the render outright on `serviceType: NodePort`. The source guard passed. `node:32015/udp` was open the whole time — because the flag only stops **new** allocations, and an apply that **omits** `nodePort` never clears an existing one.

This adds the missing layer: a live read of `spec.ports[].nodePort`.

## The scope rule that actually matters

A Service is in scope if `type == "NodePort"` **OR any `spec.ports[].nodePort` is non-zero** — not type alone.

Counting by type is precisely what made #5348 read as *"3 live NodePort services"* when a full sweep of the **same cluster** found **13** Services carrying a node port. `nodePort: 0` is the Kubernetes released-sentinel and stays inert.

## Classification — so the guard stays worth obeying

| class | behaviour |
|---|---|
| **owned namespace** | **FAIL, exit 1** |
| `cinova` · `iogrid` · `ping` · `ping-marketing` · `stalwart` · `kube-system` | reported, **not fatal** |
| `cm-acme-http-solver-*` | reported as **transient** |

Failing on a namespace we cannot fix would train people to ignore the guard, so those are visibility-only. The cert-manager solvers are created and reaped per Challenge — a point-in-time hit is noise, not a standing violation.

## Phase 0: the detector proves itself before it reports

Run against a fixture **before any cluster is touched** — the same protection `check-no-nodeports.sh` has carried since #4765, and which `check-no-banned-words.sh` was missing until #5544. All four directions pinned:

- typed `NodePort` → **caught**
- `LoadBalancer` carrying a nodePort → **caught** *(the #5348 shape)*
- `nodePort: 0` → **ignored**
- plain `ClusterIP` → **ignored**

The two negative cases are the point: without them, a detector that flagged everything would also pass.

**Fail-closed on ambiguity** — if `kubectl` is absent or the cluster is unreadable, it says so and reports nothing, rather than printing a clean bill for a cluster it never read.

## Verified both directions

```
$ scripts/check-live-nodeports.sh --self-test
OK — detector self-test passed (catches typed + LB-carried, ignores 0 and ClusterIP).
EXIT=0

$ scripts/check-live-nodeports.sh          # live mothership
FAIL: a live node port in an OWNED namespace (§854, #4765/#5348):
  openova-system/powerdns-anycast   LoadBalancer   32015,31425
EXIT=1
```

with `cinova`/`iogrid`/`kube-system`/`stalwart` correctly non-fatal and five cert-manager solvers correctly transient.

The FAIL message tells the reader **not** to `kubectl patch` it away — Helm/Flux restores it — and points at the chart-side fix in **#5547**.

## Follow-on

Wiring this into a job that has cluster credentials is the natural next step; it is deliberately not in this PR, because a workflow that cannot reach a cluster would report green without looking — the exact shape this guard exists to eliminate.

Refs #5348 #4765 #960

### DIFF
--- a/scripts/check-live-nodeports.sh
+++ b/scripts/check-live-nodeports.sh
@@ -64,10 +64,29 @@ SELF_TEST_ONLY=0
 NOT_OURS_RE='^(cinova|iogrid|ping|ping-marketing|stalwart|kube-system)$'
 
 # cert-manager HTTP-01 challenge solvers are created and reaped per Challenge.
-# They are genuinely transient, so a point-in-time hit is noise rather than a
-# standing violation — reported, never fatal. (A solver that lingers for days
-# means a STUCK Challenge; that is a cert problem, not a §854 problem.)
+# A solver that is SECONDS old is genuinely transient: a point-in-time hit is
+# noise rather than a standing violation, so it is reported and never fatal.
+#
+# The carve-out used to end there, and that was the #5561 defect. It was
+# written as if age were unobservable, so a solver 30 seconds old and a solver
+# 70 DAYS old were both filed under "TRANSIENT — reaped per Challenge, not
+# fatal". The live sweep on 2026-08-01 found the second kind: solvers standing
+# since 2026-05-23 behind two Certificates wedged on `IncorrectIssuer`. Nothing
+# was going to reap them, and the guard said so in a line that means the
+# opposite.
+#
+# Age is right there in metadata.creationTimestamp. Read it, and the carve-out
+# can say what it actually meant: transience is the reason a solver is excused,
+# so a solver that is not transient is not excused. Past the threshold the
+# NORMAL ownership rule applies — fatal in a namespace we own, reported in one
+# we do not — because at that point it is an ordinary standing node port that
+# happens to have been opened by cert-manager.
 EPHEMERAL_RE='^cm-acme-http-solver-'
+
+# A Challenge that has not completed within an hour is not mid-flight; HTTP-01
+# validation is seconds-to-minutes. Overridable so a slow-DNS environment can
+# raise it rather than delete the check.
+STUCK_AFTER_SECONDS="${STUCK_AFTER_SECONDS:-3600}"
 
 while [ $# -gt 0 ]; do
   case "$1" in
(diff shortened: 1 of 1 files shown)


emrahbaysal:
Pushed `b7c43fab9` onto this branch — the guard this PR introduces had a carve-out that could not
do the job it claimed, and the live sweep proved it (**#5561**).

`EPHEMERAL_RE` excused every `cm-acme-http-solver-*` Service as *"reaped per Challenge, not
fatal"* without ever reading age. Transience was the sole justification for the carve-out and it
was the one property never checked. Running this guard against the mothership found
`chepherd-hub/cm-acme-http-solver-rsfzq` holding nodePort 31192 for **4h** behind a Challenge
wedged on `wrong status code '503', expected '200'` — printed, under the old code, beneath a
heading asserting it would be reaped.

`detect()` now emits `age_seconds`; `classify()` splits solvers on `STUCK_AFTER_SECONDS`
(default 3600) and past the threshold applies the **normal ownership rule** — fatal in an owned
namespace, reported in a foreign one. Unknown age (`-1`) is treated as stuck, so an unparseable
timestamp fails closed. Both FAIL blocks now set a flag rather than exiting, so a stuck solver and
a chart-drift node port both report before the single exit — previously the first one to fire hid
the second.

Self-test extended with two solver fixtures identical apart from `creationTimestamp`, so both
assertions can only pass if age is genuinely read, plus foreign-namespace, unknown-age, and two
vacuity cases. Mutation-tested: reverting to the unbounded behaviour makes the self-test exit 2
with `a 70-day-old solver was excused as transient — this IS #5561`.

Full live output and the Challenge read are on #5561.

PR #5554, opened 2026-08-01
fix(wizard): a fictitious company shipped as live Organization defaults (#5401)

emrahbaysal:
fix(wizard): a fictitious company shipped as live Organization defaults (#5401)

ORG_DEFAULTS seeded INITIAL_WIZARD_STATE with:

    name: 'Acme Financial', domain: 'acme.io', email: 'platform@acme.io',
    industry: 'Financial Services', size: '2,000-10,000',
    headquarters: 'Frankfurt, Germany'

so an operator who did not overwrite every field provisioned a REAL
Organization literally named "Acme Financial" with platform@acme.io as its
billing contact. A fictitious company's details entering production as live
data.

Worse than a stale constant: SmartField RESTORES the default on blur.

    onBlur={e => { ...; if (!e.target.value.trim()) onChange(defaultValue) }}

An operator who deliberately CLEARED the field and tabbed away had "Acme
Financial" silently written back. Clearing was not an escape - the component
re-seeded it. And because the DEFAULT badge renders when
`value === defaultValue`, the fabricated value was labelled as an intended
default rather than a placeholder.

Fixed by emptying the identity fields. The blur-restore becomes a no-op, the
`required` validation engages on a genuinely empty field, and the operator has
to state who the Organization actually is.

Empty strings rather than undefined: WizardState types these as `string` and
the form binds them as controlled inputs, so undefined would flip the field to
uncontrolled and warn on first keystroke.

Scope: identity fields only. SOVEREIGN_POOL_DOMAINS and the other wizard
defaults are real platform values and are untouched.

Verification:
  tsc --noEmit clean.
  vitest src/pages/wizard -> 7 pre-existing failures in StepComponents.test.tsx
  ("every dependency points at a known component id"). Confirmed NOT caused by
  this change: stashed the edit, re-ran against clean main, identical 7
  failures. Unrelated to org defaults - the component-catalog dependency graph.
  That suite being red and unnoticed is itself a symptom of the swallowed
  vitest gate fixed in PR #5553.

This is the non-security half of #5401. The tier-gating half (per-Org
/settings rendering the Sovereign operator panel) is separate; I posted the
severity determination there - the Revoke/Rotate/Decommission controls have
NO mutation handlers wired (grep useMutation|authedFetch|fetch → 0) and
secrets render masked, so it is information disclosure, not privilege
escalation.

Refs #5401 #5553 #960


emrahbaysal:
Superseded by #5579, which salvages the six changes this branch carries that current main does not.

Adjudication of the 30 commits against main (2026-08-02):
- The docs/ledger payload landed via #5566/#5576; the vitest-gate commit is a duplicate of merged #5553 (now on main with an added crash backstop).
- The helm v3.16.0 pin, chart 1.4.1261, and check-no-nodeports.sh copies here are OLDER than what main carries (v3.16.3 with rationale, 1.4.1263, the #5578 advisory split) — merging them would regress main.
- The banned-words gate correctly failed this PR on a commit-body word; the salvage lands with clean history instead of a 30-commit rewrite.

Nothing in this branch is lost: wizard ORG_DEFAULTS (#5401), the #5515 topology_loader live-region fix + tests, the #5482 divergence test, and the mothership live-walk doc all ride #5579.

PR #5557, opened 2026-08-01
fix(wizard): step 6 renders no Back control — StepDomain omitted onBack (#5555)

emrahbaysal:
Closes the defect found on this morning's live walk of `console.openova.io/sovereign/wizard`.

## The defect

Step 6 of 8 ("Where will your Sovereign live in DNS?") rendered a footer with **only a disabled
`Continue →`** — no `← Back`. Steps 2–5 all had one.

It landed on the worst possible step: `Continue` is disabled there by default because the domain
fields start empty, so the operator faced a dead primary action *and* an absent reverse action. Not
hard-trapping — the header progress chips still work — but the footer contract silently breaks on
the one screen where it matters most.

## Root cause — one destructure

```diff
-  const { next } = useStepNav()
+  const { next, back } = useStepNav()
...
   onNext={next}
+  onBack={back}
```

`StepDomain.tsx:79` took only `next`, where all six siblings take `{ next, back }`.
`_shared.tsx:10` types the prop as `onBack?:` — **optional** — so this was neither a type error nor
a lint error. Nothing could have caught it except a human reaching that step in a browser.

## The guard is the actual deliverable

A one-line fix doesn't stop the next omission, so this adds a repo-wide contract test: every file
mounting `<StepShell>` must pass `onBack`, with an `INTENTIONALLY_NO_BACK` map where each exemption
**states its reason**.

It reads the step sources rather than rendering them — deliberately. The defect is a *missing prop
at the call site*; a render test would need each step's full store/provider scaffolding to reach the
same assertion, and would test a different layer than the one that broke.

**The guard immediately found two more steps**, both of which turned out to be legitimate and are
now documented instead of silently passing:

| step | verdict |
|---|---|
| `StepOrg.tsx` | step 1 of 8 — nothing precedes it, so Back has no destination |
| `StepNSDelegation.tsx` | post-handover surface, not in the 8-step flow; takes `onNext` as a prop and is never mounted in the main wizard |

That is the difference between fixing one file and fixing the class: the exemptions are now explicit
and reviewable rather than indistinguishable from bugs.

It also carries a **vacuity check** — if the file glob ever drifts and matches fewer than six steps,
the guard fails rather than passing trivially against an empty set. A guard that can't prove it's
looking at anything is the #5544 fail-open shape, which I fixed earlier today.

## Verification

```
guard                        12/12 pass
StepDomain.test.tsx          18/18 pass (unchanged)
tsc --noEmit                 clean
```

Refs #5555


emrahbaysal:
Superseded by #5572, which is the same fix rebased onto current main. Merged as 91ed1e71b — verified on origin/main: StepDomain now destructures back from useStepNav and passes onBack={back} to StepShell. Refs #5555

PR #5560, opened 2026-08-01
fix(bp-powerdns): state nodePort:0 EXPLICITLY on the anycast LB — §854 (#5348)

emrahbaysal:
§854 elimination, not deferral. The live mothership Service **still carries two nodePorts right
now**; this is the change that clears them.

## Live evidence (audited today, 2026-08-01)

```
$ kubectl get svc -A          # 164 services scanned
openova-system/powerdns-anycast   type=LoadBalancer   nodePorts=32015,31425
```

## Why `main` produces this

`main`'s template already sets `allocateLoadBalancerNodePorts: false` — and that is **not enough**:

> The flag stops the apiserver allocating **new** ports. It does not retract ports already assigned,
> and a declarative apply that merely **omits** `nodePort` leaves an existing one in place.

So a Service can hold `allocateLoadBalancerNodePorts: false` **and** live nodePorts simultaneously.
That is exactly what `main` yields today, and it is why a repo-only scan reports clean while the
cluster violates.

The in-chart comment written when this fix was authored names `nodePort=32015 (53/UDP)` and
`nodePort=31425 (53/TCP)` explicitly. Today's independent audit re-observed **those same two ports**
live — upgrading that comment from a claim to a verified prediction.

## The fix

`nodePort: 0` is the apiserver's **deallocate sentinel** — it is read as "no node port" and clears
the allocation on apply. It is also the inert value the repo guard deliberately ignores (the guard
flags only `nodePort: [1-9][0-9]*`), so the chart stays guard-clean by construction.

Four-site lockstep, all verified on this branch:

| site | value |
|---|---|
| `chart/Chart.yaml` | 1.2.23 |
| `blueprint.yaml` | 1.2.23 |
| `clusters/_template/bootstrap-kit/11-powerdns.yaml` | 1.2.23 |
| `chart/templates/anycast-endpoint.yaml` | `nodePort: 0` per LB port |

## Why NOT ClusterIP

Worth stating, because it is the intuitive wrong turn: PowerDNS must answer DNS on an **external
VIP**. §854's *compliant* pattern is a `LoadBalancer` served by Cilium LB-IPAM sharing the Sovereign
EIP (`lbipam.cilium.io/sharing-key`) — which this Service already is. The violation is the
auto-allocated nodePorts riding along, **not** the LoadBalancer type. Switching to ClusterIP would
trade a §854 finding for a DNS outage.

## Verification

```
scripts/check-no-nodeports.sh  ->  OK: zero NodePorts across sources + rendered charts (#4765)
```

This commit previously existed only inside two large unrelated branches
(`fix/5401-org-defaults-…`, `fix/5555-wizard-…`), with no focused PR — which is why it never
merged. This is that PR.

Refs #5348 #4765

### DIFF
--- a/clusters/_template/bootstrap-kit/11-powerdns.yaml
+++ b/clusters/_template/bootstrap-kit/11-powerdns.yaml
@@ -155,8 +155,14 @@ spec:
       # tofu dns LB :53->node:53 forward reaches a real listener (the LB-IPAM
       # anycast VIP path is dead on Hetzner). §854-clean: hostPort, NEVER a
       # NodePort. Driven below by ${POWERDNS_DNSDIST_HOSTPORT} (Hetzner-only).
+      # 1.2.23 (#5348): state nodePort:0 EXPLICITLY on the anycast LB ports.
+      #         allocateLoadBalancerNodePorts:false alone does NOT deallocate
+      #         ports already assigned, and an apply that OMITS nodePort leaves
+      #         them in place — the live mothership Service carried the flag AND
+      #         nodePort=32015/31425 at the same time (§854 violation invisible
+      #         to every source-side check).
       # 1.2.22 (#5086): render-contract regression lock (no chart behaviour change).
-      version: 1.2.22
+      version: 1.2.23
       sourceRef:
         kind: HelmRepository
         name: bp-powerdns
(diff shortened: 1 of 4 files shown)


emrahbaysal:
Duplicate of #5547 — byte-identical diff against main (same 4 files, same content). Verified: `diff <(git diff main...fix/5348-powerdns-explicit-zero-nodeport) <(git diff main...fix/5348-powerdns-nodeport-zero)` returns no output. Consolidating on #5547, opened first. The catalog-seed lockstep fix for the 1.2.23 bump lands there.

PR #5952, opened 2026-08-10
chore(uat): add the missing convergence-capture script + record cycle 2026-08-09

emrahbaysal:
## What

The 6-hourly convergence capture calls for `scripts/uat-snapshot.py`, which **had never been written** — only the pivot renderer existed. Adding it rather than skipping the step.

## Why a second sheet

- `uat-raw.csv` is **per-row** and rebuilt from UAT.md by `uat-backfill.py`.
- `uat-cycles.csv` — written here — is **per-cycle** and records what a per-row sheet structurally cannot: **which environment** and **which deployment** the walk ran against.

Without that column the trend is unreadable, because the score resets to near-zero on every fresh prov (measured: 133 green 2026-07-30 → 25 on 2026-08-03). **A drop across an env boundary is the ledger working as designed; a drop within one env is a regression.** Only the env column separates them — so the script prints which case applies *at capture time*, rather than leaving it to whoever reads the trend later.

## Hard refusals, not warnings

- **Denominator:** if UAT.md does not hold exactly 286 rows it exits non-zero and writes nothing. **Verified able to fire** — removing one row produces `refusing: UAT.md holds 285 rows`.
- **Vocabulary:** any verdict glyph outside the canonical set is refused, since an uncounted glyph reads as "never walked" downstream.

The timestamp is a **required argument** rather than a clock read, so a capture is reproducible.

## Cycle recorded

`hw292` / dep `1c56518035a83e03` — `205 PASS · 44 FAIL · 27 PARTIAL · 7 SUPERSEDED · 3 PENDING = 286`, 71.7%.

(That snapshot predates the no-partials and zero-pending merges; it is recorded as-was, since past cycles are append-only and never edited.)

Refs #3687


emrahbaysal:
Superseded: scripts/uat-snapshot.py + uat-cycles.csv are on main. Refs #6778

PR #5985, opened 2026-08-10
fix(catalyst-api): the DR runbook preflight must measure what it names, and its Fail path must be reachable

emrahbaysal:
## What was actually wrong

`runDRPreflight` built the 10-check DR runbook preflight matrix by declaring every entry `Status: "Pass"` as its **literal initial value**, then re-read only two of them. Eight checks reported `Pass` having probed nothing — while the function's own docstring claimed *"Each check reads from a different cluster surface (continuum CR, cnpgpair CR, PDM CR list, Cluster CR, etc)."*

Measured read-only on hw292 (dep `1c56518035a83e03`, 2026-08-10):

```
$ kubectl get cnpgpairs.dr.openova.io -A
No resources found
$ kubectl get crd cnpgpairs.dr.openova.io -o jsonpath='{.spec.group} {.spec.names.kind}'
dr.openova.io CNPGPair
```

The CRD is installed with **zero** instances and `preflight-02 cnpgpair-streaming` reported `Pass`. A check that cannot fail on the exact precondition it names, pointed at the DR path.

### The systemic half — the playback abort branch was dead code

No check could reach `Fail` at all (`checks[0]` and `checks[3]` could only become `Warn`; the other eight were never re-read). `classifyPreflight` returns `NotReady` only on a `Fail` with `Severity == "critical"`, so:

- `overall` could only ever be `"Ready"` or `"DegradedReady"`;
- `blockingChecks` was **always empty**;
- `HandleDRRunbookPlayback`'s `if preOverall == "NotReady" && !body.AcceptWarn { abort }` was **unreachable**.

The safety gate in front of a mutating DR operation could not fire. Neither `runDRPreflight` nor `classifyPreflight` had a single test.

Two further reading defects in the same function: the no-client path returned the **all-Pass** matrix (a clean bill of health from zero evidence), and `classifyPreflight` could return `"Ready"` while `blockingChecks` was non-empty.

## The change

Every check starts **unmeasured** (`Skipped` + the reason) and reports `Pass` only after something positively measured it.

| check | now |
|---|---|
| `01 continuum-cr-ready` | `Fail`/critical on any non-ready Continuum phase. `Healthy` **or** `FailedOver` stay ready — blocking `FailedOver` would strand the failback of an already-failed-over Sovereign. |
| `02 cnpgpair-streaming` | reads the kind it **names** first; when that is empty falls back to the labelled cnpg `Cluster` halves that the rest of this package already reads a pair from. Replica half present-but-not-ready ⇒ `Fail` (#4901 region-kill); replica half not visible from this client ⇒ `Warn` (#5511 per-region split); nothing resolvable ⇒ `Warn`. |
| `03 wal-lag-under-rto` | `status.replicationLagSeconds` (the key the controller writes, #5601) against each CR's own RTO budget. An **absent** lag key reads "not reported", never a silent zero. |
| `04 quorum-2of3-witnesses` | unchanged semantics; a list error is no longer folded into the same verdict as a real shortfall. |
| `05`–`10` | no probe exists on this path ⇒ `Skipped` naming the surface, never a fabricated `Pass`. |

`classifyPreflight`: `Skipped` degrades the verdict, and `Ready` is impossible while `blockingChecks` is non-empty.

No chart version touched, no new dependency, no new k8s wiring — the probes reuse `cnpgClusterGVR` / `cnpgPairLabel` / `cnpgStandbyAvailable` from `continuum_live.go`.

## Proof the tests can fail

`continuum_dr_preflight_test.go`, 13 cases. Control run: source reverted to `HEAD` with **only** the new symbols shimmed in, so failures are behavioural rather than compile errors. **11 of 13 failed**:

```
--- FAIL: TestDRPreflight_CNPGPairStreamingIsNotPassWhenNoPairExists
      got Pass with ZERO CNPGPair CRs and ZERO labelled cnpg pairs
--- FAIL: TestDRPreflight_CNPGPairNotStreamingIsCriticalFail        got "Pass" want Fail
--- FAIL: TestDRPreflight_ReplicaHalfPresentButNotReadyIsFail       got "Pass" want Fail
--- FAIL: TestDRPreflight_ReplicaHalfInvisibleIsWarnNotPass         got "Pass" want Warn
--- FAIL: TestDRPreflight_DegradedContinuumIsCriticalFail           got "Pass" want Fail
--- FAIL: TestDRPreflight_WALLagOverRTOIsCriticalFail               got "Pass" want Fail
--- FAIL: TestDRPreflight_AbsentLagKeyIsNotAMeasuredZero            got Pass from a CR with NO lag key
--- FAIL: TestDRPreflight_UnprobedChecksDoNotReportPass             preflight-05..10 all Pass
--- FAIL: TestDRPreflight_UnknownDeploymentIsAllUnmeasured          all 10 Pass with no client; rollup Ready
--- FAIL: TestDRPreflight_CriticalFailReachesNotReadyAndBlocks      got "DegradedReady" want NotReady
--- FAIL: TestClassifyPreflight_NonCriticalFailIsNotReadyVerdict    Ready with 1 blocking check
```

Two guards deliberately pass in both states and are there for that reason:

- `TestDRPreflight_HealthyMatrixStillPasses` — the **vacuity control**. A genuinely healthy Sovereign (Continuums Healthy and caught up, both pair halves Ready, three PDM witnesses) must still report the four probed checks `Pass`, so this file cannot be satisfied by hardwiring the matrix to `Fail`.
- `TestDRPreflight_FailedOverPhaseIsStillReady` — locks the failback-safe phase.

`TestDRPreflight_AbsentLagKeyIsNotAMeasuredZero` exists specifically to close the fixture trap this repo has been bitten by: a fixture that omits the field the bug keys on passes on data that could never exercise the defect. Here the omission **is** the case under test.

Full `go test ./...` on the module is green.

## Correction to the record this was found through

The premise "nothing produces `Continuum` CRs" (`docs/ledger/UAT.md` row 57, commit `b6e7b2ee`) is **refuted on the live cluster**, and the method that produced it is the reusable lesson:

- #4986's producer is `platform/postgres/chart/templates/continuum.yaml` (PR #4987, merge `7879bcb2`, 2026-07-11). It is a **Helm template, not Go code**, so `grep -rln 'Continuum{'` over `core/` + `products/` structurally cannot see it — the zero it returned was a wrong-surface query read as an absence.
- Live on hw292 region A: `kubectl get continuums.dr.openova.io -A` returns **8 CRs, all `phase=Healthy`, age 7d6h**, including `dr-shared-pg`, `dr-shared-pg-b`, `dr-shared-pg-c`. For Continuum, #4986 is merged **and** deployed **and** working.
- `CNPGPair` genuinely has no production producer (only `qa-fixtures/cnpgpair-qa.yaml`, behind `qaFixtures.enabled`), **but no consumer depends on one**: `findCNPGPairForApp` and `cnpgPairStandbyForContinuum` both resolve pairs from cnpg **`Cluster`** halves via the `catalyst.openova.io/cnpg-pair` label. That is why rows 54/56/58 render live DR data on this same env. The one surface that misreports from the empty kind is this preflight check.

Row 57's evidence cell is updated with that correction. **The verdict is unchanged (❌, deploy-gated)** — only the stated cause is corrected. Table-shape and tally guards run and green (286 rows, 194 green).

Naming trap worth carrying forward: `cnpg/cnpg-pair-bp-cnpg-pair-primary` is a CNPG **`Cluster` named** `cnpg-pair-*`, not a `CNPGPair` CR. Different kinds, near-identical names — reading one as the other makes the dependency look satisfied.

Refs #5982 #4986 #4987 #5601 #4901 #5511 #3375


emrahbaysal:
Duplicate of merged #6094 (70f4da6c7). Refs #6778

PR #6181, opened 2026-08-11
docs(ledger): convergence cycle 2026-08-11T10:45Z — and a CARRIED row is not an observation

emrahbaysal:
Capture: **205 ✅ · 76 ❌ · 5 ☐ (71.7%)** on hw294 / dep `336e2891743d6c98`. **17 rows gained green**, each measured live with a control. **Zero lost green.** 188 rows carried from hw293 with confidence decayed (#6180).

### The capture caught a defect in itself

`--observe` recorded all 205 as hw294 measurements — including the **188 CARRIED from hw293 that were never walked on this machine**. Left standing, the scorer would promote those rows to long walk intervals on evidence gathered elsewhere: `cross_env_proof` double-counting its own output, which is exactly the false confidence the environment column exists to prevent.

`--observe` now refuses a CARRIED row by name and reports how many it skipped. The carried verdict still counts green (**a wipe is not a failure**); the scorer still reads its real evidence from the prior environment's entries. It simply no longer pretends this machine produced it.

### Trend from the raw sheet — the honest answer to "why isn't 76 falling"

```
              total   ✅    ❌    ⚠️    ☐    ⛔  PENDING
  2026-08-06    248  177    12    33    0   26     129
  2026-08-09    253  167    15    52    0   19     120
  2026-08-10    279  180    78     0   21    0      41
  2026-08-11    286  205    76     0    5    0       0
```

❌ grew 12 → 76 because **PENDING went 129 → 0** and **PARTIAL 33 → 0**. Six days ago "12 failures" sat beside 129 rows nobody had examined and 33 half-measured ones — those weren't passing, they were unasked. **The failing set did not grow; the measured set did.** Meanwhile ✅ climbed 177 → 205 against a denominator that grew 248 → 286.

Every hiding place is now gone: no PARTIAL, no PENDING, no SUPERSEDED, 5 unwalked.

Refs #6114 #6158


emrahbaysal:
Superseded: uat-confidence.py's cross-env refusal is on main (self-test covers it); the 2026-08-11 cycle numbers are historical. Refs #6778

Issue #6293, opened 2026-08-14
Cutover step 6 patches HelmRepositories owned by the org-tenants Kustomization — Flux reverts them within 60s and the pivot can never converge


emrahbaysal:
Cutover step 6 (`helmrepository-patches`) patches live `HelmRepository` objects
that are OWNED by a Flux `Kustomization`. Flux drift-corrects them back to
`oci://ghcr.io/openova-io` within its 1-minute interval, the step's own
Phase-1.5 read-back sees them still upstream, and the step fails — permanently.
On hw296 this is now the terminal state of the sovereignty cutover: 5 of 11
steps green, step 6 `BackoffLimitExceeded`, steps 7-11 never run.

## Terminal state — hw296, dep `e689e3b34a75fdec`, 2026-08-14T02:10Z

`cutoverComplete=false`, `progressPercent=45`, `failedStep=helmrepository-patches`.

| # | step | result | duration |
|---|---|---|---|
| 1 | gitea-mirror | success | 122s |
| 2 | harbor-projects | success | 6s |
| 3 | harbor-prewarm | success | 4374s |
| 4 | registry-pivot | success | 15s |
| 5 | flux-gitrepository-patch | success | 226s |
| 6 | **helmrepository-patches** | **failed** | 515s |
| 7-11 | catalyst-api-env-patch, gitea-token-mint, vcluster-registry-pivot, crossplane-provider-pivot, egress-block-test | never ran | — |

`Job/cutover-helmrepository-patches-1786668062`: `failed=4 backoffLimit=3
Failed=True reason=BackoffLimitExceeded`.

## The failure, verbatim

```
[helmrepository-patches] OK bp-stalwart-tenant -> oci://registry.hw296.omani.works/openova-io (certSecretRef=cutover-harbor-ca)
[helmrepository-patches] OK bp-wordpress-tenant -> oci://registry.hw296.omani.works/openova-io (certSecretRef=cutover-harbor-ca)
[helmrepository-patches] live-K8s ok=4 skip=62 fail=0
[helmrepository-patches] FATAL: Phase-1.5 read-back: 5 region-A HelmRepository(ies) are STILL on oci://ghcr.io/openova-io after the pivot — the Sovereign would record a SUCCESSFUL cutover step while its charts resolve to the mothership; refusing to succeed (#5436)
  STILL-UPSTREAM bp-agenity=oci://ghcr.io/openova-io
[...]

PR #6304, opened 2026-08-14
docs(uat): the H1 said the cutover was never fired — it is terminal-failed at step 6, and delivery is severed

emrahbaysal:
The ledger header is what every reader and every future walker anchors on. Three of its claims were materially false.

It said:

> The cutover has NOT been fired here — `progressPercent=0`, all 44 per-step keys empty — so every cutover-dependent row is honestly unreachable rather than red; **no customer Organization exists**

Measured on hw296 tonight:

| claim | reality |
|---|---|
| cutover never fired | fired **eight** times; `progressPercent=45` |
| all 44 step keys empty | steps **1–5 succeeded** (`harbor-prewarm` 4374s), step 6 `helmrepository-patches` **FAILED** |
| no customer Organization | **five** — `walkone` `walktwo` `walkthree` `walkfour` `walkfive` |

So **G11 is a measured refutation, not an unreached row** — a materially different claim about the platform. And the funnel-terminal, per-Org and cross-Org rows are walkable; several were walked green tonight.

**166 and 227 stay unreached**, but for a reason the old text didn't give: their clauses are conditional on a COMPLETE cutover, and red would be a *false* verdict — `registry-pivot` actually succeeded and `egress-block-test` never ran.

## The finding that gates every remaining row, which nothing in the ledger recorded

**Delivery is severed.** Cutover step-05 repointed Flux to the local Gitea at **02:00:05Z**, and that Gitea is **not a pull-mirror** — bare, no remote, filled by a one-shot Job, newest GitHub-derived commit `641e7617e` at **00:38:43Z**. Anything merged after 02:00Z cannot reach this Sovereign. The deploy commit carrying #6291 landed at **02:26Z — twenty-six minutes past the pivot**.

That is **one cause behind three separate red rows**. Without it in the header, the next walker reads "merged" as "delivered" and stamps reds against code that isn't there — the precise false-negative this session spent the most effort learning to refuse. The header now instructs verifying `merge-base --is-ancestor <fix> <RUNNING image>`, against the running image and never the tag main claims mid-roll.

It also records that the next cutover must be **operator-fired**, because `cutover.go:1948` misses both retry arms for a `BackoffLimitExceeded` Job and an auto-retry silently re-reports the stale one.

## Gates

All four ledger gates green **on exit codes**: table shape, drift, clause identity, scheduler-respect. **No row verdict changed** — header text only, denominator STONE at 286.

Refs #6293 #6303 #3379


emrahbaysal:

PR #6320, opened 2026-08-14
fix(wordpress-tenant): CNPG-operator carve-out + zero install-time egress (Refs #6311 #3988 #4282 #3379)

emrahbaysal:
## What this fixes

`bp-wordpress-tenant` could not converge in **any** Organization. Two independent defects, both fixed here. Chart `0.4.23 -> 0.4.24`.

Refs #6311 #3988 #4282 #3379

---

## Defect 1 — no `cnpg-system` carve-out, so our own NetworkPolicy blinds the CNPG operator

This chart renders its CNPG `Cluster` into the Organization's own namespace, which carries the `default-deny-all` + `allow-same-org` pair stamped by the Organization controller's GitOps tree (`core/controllers/organization/internal/gitops/manifests.go`, `networkPolicyTemplate`). `allow-same-org` re-opens **same-namespace** traffic only, so the `cnpg-system` **operator** — a different namespace — is denied the instance Pod's `:8000/pg/status` probe.

The database is up and serving the whole time; only the operator is blind. CNPG parks the Cluster at:

```
Instance Status Extraction Error: HTTP communication issue
readyInstances 1/1
ConsistentSystemID=False (NotFound)
```

`Ready=False` -> the HelmRelease never converges -> the customer's purchased WordPress never comes up.

**Measured with quota eliminated as a variable** (hw296): the identical blueprint installed into an Org with headroom ran the exact Pods #5393 rejects elsewhere — `wordpress-db-1-initdb` Completed, `wordpress-db-1` 1/1 Running, zero quota rejections — and still landed `Ready=False`. **Control sharing the suspect property**: `walkone/postgres`, same deny pair but carrying the #4282 carve-out, read `Cluster in healthy state` 3/3. Eleven CNPG Clusters cluster-wide were healthy; this was the only one in the error phase.

**Fix** — `templates/networkpolicy-cnpg-operator-probe.yaml`, the proven #4282 shape copied from `platform/postgres/chart/templates/networkpolicy-singleton-operator-probe.yaml` (whose header predicts this phase string verbatim). It opens **only** `:8000` + `:9187` from **only** `cnpg-system`.

- **Additive.** `default-deny-all` and `allow-same-org` are untouched; NetworkPolicies union.
- **No broad 5432 rule.** The bp-postgres `sharedConsumers` shape (#4442) does not apply: this Cluster's only consumers are the WordPress Deployment and hook Jobs in the same namespace, already admitted by `allow-same-org`.
- **Singleton only**, exactly like #4282. In `active-hot-standby` the replica streams WAL cross-cluster over Cilium ClusterMesh; a policy selecting the Cluster Pods would default-deny that stream, and a remote-cluster peer cannot be expressed in a k8s NetworkPolicy at all (an `ipBlock` never matches a ClusterMesh remote endpoint identity — bp-postgres needs a dedicated identity-based CNP for it). Rendering nothing in AHS keeps that path byte-identical rather than trading one break for another. The AHS carve-out stays open on #6311.

### The decision on the other 7 carve-out-less charts

Eight charts render a CNPG Cluster with no carve-out, but the deny pair exists **only in per-Org namespaces**. `category: organization-app` is *not* the right discriminator — it would have missed `bp-postgres`, which is `category: data` and *is* routed per-Org by #4398.

The load-bearing property is **"can this chart's Cluster land in a namespace carrying the `default-deny-all` + `allow-same-org` pair"**. Exactly two charts can: `bp-postgres` (already carries the #4282 carve-out) and this one.

The other seven — `bp-gitea`, `bp-guacamole`, `bp-llm-gateway`, `bp-openova-flow-server`, `bp-powerdns`, `bp-powerdns-admin`, `bp-temporal` — are bootstrap-kit components installed into platform namespaces where no such pair is stamped. Adding an unconditional probe NetworkPolicy there would **actively regress** them: `policyTypes: [Ingress]` makes the Cluster Pods default-deny for every unlisted peer, and their consumers live in *other* namespaces. That is the #4442 regression verbatim (keycloak-0 -> shared-pg-rw JDBC timeout, whole SSO chain down). **They are deliberately untouched.**

---

## Defect 2 — install-time `github.com` fetch, which contradicts Pillar 5

The post-install `oidc-config` Job CrashLoopBackOff'd on `curl: (28) Connection timed out after 120001 milliseconds`.

Cutover step-08 (`egress-block-test`) holds a 10-minute deny-egress NetworkPolicy against `github.com`, `ghcr.io` and `harbor.openova.io` and requires the cluster to reconcile green through it — that hold **is** the sovereignty proof (ADR-0002). A customer Application that reaches GitHub to install cannot exist on a sovereign Sovereign.

**There were four install-time fetch sites, not two:**

| Where | Host |
|---|---|
| `deployment.yaml` `wp-plugin-install` init (a) | `downloads.wordpress.org` |
| `deployment.yaml` `wp-plugin-install` init (b) | `github.com` |
| `oidc-config-job.yaml` step 0 | `github.com` |
| `oidc-config-job.yaml` step 4 | `github.com` |
| `oidc-config-job.yaml` step 7 `wp theme install` fallback | `wordpress.org` (via wp-cli) |

The fifth was latent but **always reachable**: the Job's `emptyDir` masks the image's bundled themes, so `wp theme is-installed` was false on every fresh Org and the fetch fired every time.

### The artifact-mirroring seam that was found

**There is no seam for plain file artifacts.** This repo mirrors exactly two artifact classes into a post-cutover Sovereign's Harbor: **OCI container images** and **Helm charts** (cutover step-03 `harbor-prewarm`). No `oras`, no zip/tarball mirroring exists anywhere under `platform/self-sovereign-cutover/`.

So the correct reuse is to carry the bytes as **layers of an image that is already mirrored** — and `wordpress-tenant-pg` is already in the prewarm set **by name**: `platform/self-sovereign-cutover/chart/Chart.yaml` calls out `wordpress-tenant-pg:<tag>@sha256:...` verbatim as a tag+digest-pinned prewarm source. **No new mechanism is introduced.**

`image/Dockerfile` now bakes pinned `pg4wp 3.3.1` + `openid-connect-generic 3.11.3` into `/usr/src/openova-wp-artifacts` with build-time assertions. Fetching happens at **build** time on a runner, not at **install** time on a Sovereign; after this change a converging Sovereign reaches for nothing.

Both chart consumers copy from there and **fail closed** if it is absent — no network fallback, because a silent fallback is precisely the sovereignty violation being removed. Core's bundled themes are seeded into the Job so step 7 activates rather than fetches.

### Two latent defects surfaced by the change and fixed with it

1. **Plugin-slug mismatch.** The Job activated `openid-connect-generic` while the PVC held `daggerhart-openid-connect-generic`. WordPress stores the plugin's *path* in `active_plugins`, so that row pointed at a file the runtime PVC never contained and the plugin never loaded at runtime. Both sites now use the bare slug, from the same baked source.
2. **Missing `imagePullSecrets` on the `oidc-config` Job.** Its new `wp-artifacts` init runs the **private** ghcr image; without the secret it ImagePullBackOffs, the post-install hook never runs, Helm times out, HR `Ready=False` — the exact failure shape #4154 fixed for the Deployment and the admin-user Job.

---

## Tests — 31 per-assertion vacuity proofs

Every assertion is mutated back to the pre-fix shape **one behaviour at a time** and required to go RED. An assertion that cannot go red is not a test.

**`tests/6311-cnpg-operator-probe-netpol.sh`** — 8 assertions, **15 mutations**, plus 5 render-gate cases:

```
Case 1: default (singleton) render carries the CNPG-operator carve-out            PASS
Case 2: gates — absent without the CNPG capability / with the Cluster off /
        with networkPolicy off / with cnpgOperator off / in active-hot-standby    PASS
Case 3: RED as required: A1 carve-out deleted (pre-fix chart)
        RED as required: A2 podSelector points at the wrong label
        RED as required: A3 operator namespace wrong
        RED as required: A3 same-namespace podSelector peer added
        RED as required: A4 status port 8000 missing
        RED as required: A4 metrics port 9187 missing
        RED as required: A5 blanket namespaceSelector: {} added to ingress
        RED as required: A5 broad 5432 consumer rule added
        RED as required: A7 egress half dropped (the #5617 ingress-only shape)
        RED as required: A7 hollow egress: [] (grep-passes, resolves nothing)
        RED as required: A7 53/UDP only, no TCP
        RED as required: A7 DNS pointed at the Org namespace, not cluster DNS
        RED as required: A7 DNS via ipBlock 0.0.0.0/0 (#4360 — matches no pod)
        RED as required: A8 blanket egress destination added
        RED as required: A6 chart's own WordPress NetworkPolicy removed
```

**`tests/6311-no-install-time-egress.sh`** — scans every executable line of every container in every workload, **init containers included** (scanning only the main container is how a reintroduced fetch would hide). Comments are stripped deliberately: the prose names the very hosts being forbidden, so a raw substring scan would false-positive on its own rationale. **203 lines scanned with an explicit floor**, so the scan cannot pass vacuously. **10 mutations**, including reintroducing the fetch in an initContainer, `wp plugin install`, `wp theme install`, removing the seed, pointing `wp-artifacts` at the upstream wp-cli image, unmounting the shared volume, and replacing the fail-closed guard with a fallback. Plus **6 Dockerfile mutations** (floating refs, dropped build-time assertions, wrong slug).

**Two existing suites asserted the defects and were inverted:**

- `tests/oidc-config.sh` Case 8 asserted the pinned **GitHub archive URL must be present**. Now asserts zero install-time egress from any container, and that the artifacts arrive from the image.
- `tests/imagepullsecrets-render.sh` Case 3 asserted *"the oidc-config Job must NOT carry imagePullSecrets"*. Now requires them — and the requirement is **derived, not asserted by fiat**: the case first proves the Job references the private image, then requires the secret.

**All six chart suites green**, gated on exit code (`suite && suite && ...`, never piped to `tail`).

## Image verification

The baked layer was built and verified locally with podman:

```
+ test -s /usr/src/openova-wp-artifacts/db.php
+ test -s /usr/src/openova-wp-artifacts/pg4wp/driver_pgsql.php
+ test -s /usr/src/openova-wp-artifacts/oidc/openid-connect-generic.php
PROBE OK: uid-33 seed works, twentytwentyfive bundled
pg4wp=3.3.1
openid-connect-generic=3.11.3 (oidc-wp/openid-connect-generic)
```

The uid-33 probe is a **positive proof** that the runtime user can read and copy the baked artifacts, replacing a blanket `chmod -R a+rX` that would have been an assumption dressed as a fix (and was the one step in the file that cannot run under a rootless builder).

## Merge-order note for whoever lands this

All five lockstep sites move in this PR: `chart/Chart.yaml` (0.4.24),
`blueprint.yaml`, the catalog-seed `spec.version` + `source.version` (both the
`bp-wordpress-tenant` card and the `bp-wordpress` alias), `blueprints.json` and
`catalog.generated.ts`. The seed was synced with `scripts/sync-catalog-seed-pin.py`
and the generated catalog with `build-catalog.mjs` — neither was hand-edited.
Because the seed lives in the umbrella's rendered surface, `products/catalyst`
moves too: **1.4.1492 -> 1.4.1494**, computed by `scripts/bump-chart-version.sh`
(ceiling 1.4.1493 from open PR #6322) immediately before pushing, with
bootstrap-kit slot 13 bumped in the same commit.

This number was derived three times (1487 -> 1490 -> 1494) and the reason is
structural, not specific to this branch: the deploy-bot auto-bumps the umbrella
on every merge to main, independently of any open PR, so a version computed at
the start of a CI cycle is stale by the end of it. The derivation has to happen
immediately before the push and be redone after every rebase.

Two guards are needed and neither is sufficient alone:
`check-chart-version-not-claimed-by-open-pr.py` compares only against open PR
heads and returns **exit 0 for a version below main**, while
`check-chart-version-forward.sh` compares only against main and cannot see a
sibling PR's claim. Both are run here against a freshly fetched `origin/main`,
chained on exit codes. (The forward check cannot be applied to
bp-wordpress-tenant: its `appVersion` is `"6"`, the WordPress major, which is
not semver — that chart is covered by `check-chart-version-bumped.sh`.)

Rebased onto main after #6316 merged underneath this branch. The only
conflicts were two version lines, resolved hunk-wise so main's other edits to
both files survive — `git checkout --theirs` would have taken my whole file and
silently dropped them. The generated catalog was **not** hand-merged: it
auto-merged, then `build-catalog.mjs` was re-run and produced no diff,
confirming the merged bytes are exactly what the generator emits.

One ordering consequence worth knowing: `build-bp-wordpress-tenant-pg.yaml`
rewrites `wordpress.image.{tag,digest}` and bumps `Chart.yaml` on every image
change, then re-dispatches `blueprint-release`. So on merge **two chart versions
publish** — `0.4.24` from this commit (new templates, previous image digest) and
then `0.4.25` from the image build (new templates **and** the image carrying the
baked artifacts). **`0.4.25` is what should end up in the catalog seed**, and
`blueprint-release` syncs it there automatically.

In the few minutes between the two publishes a chart pinned at `0.4.24` fails
closed with an explicit *"baked artifacts missing — check
`wordpress.image.{tag,digest}`"* message rather than silently reaching for
GitHub. That is the correct failure: loud and diagnostic, never a silent
sovereignty violation.

UAT row 222 is **not stamped**: delivery to hw296 is severed (#6307), so a
merged fix is not a live one, and 222 also carries the founder-gated quota
(#5393) in the Org its clause names.

## CI failures found and fixed on the first run

Three real, one phantom:

1. **`Every constrained workload keeps cluster DNS`** — the #5617/#4437 class
   guard flagged the new carve-out, correctly. A per-Org namespace's
   `allow-gateway-and-apiserver` CNP (`endpointSelector: {}`) constrains egress
   while naming no DNS, so a workload shipping no DNS egress of its own is mute
   on its first lookup — exactly how #5617 bit live on hw292 Org `uatco`. The
   carve-out now carries a DNS-only egress half (53/UDP **and** 53/TCP, by
   `namespaceSelector`, never an `ipBlock` — a CIDR never matches an in-cluster
   pod identity under Cilium, #4360). Additive: policies union, so it can only
   ADD `53 -> kube-system` to what `allow-same-org` and the per-Org CNP already
   permit.

   *The "split it so the probe stays ingress-only" alternative does not work,
   and this was measured rather than assumed*: the guard classifies an
   ingress-only policy in a **per-Organization** install as the #5617 shape
   regardless of which file it lives in. Mutation `A7` reproduces exactly that
   split and the guard's own `group_allows_dns` returns *"no egress rule at all
   (ingress-only policy)"*. Copying `platform/postgres` verbatim is what
   produced the violation: its #4282 template **is** ingress-only and passes
   only because bp-postgres is classified a bootstrap-kit slot chart, where
   `ns_constrains_egress` is false. Same ingress shape, different namespace
   contract.

   Six further vacuity proofs added (A7/A8), and the DNS predicate is
   **imported from the class guard and called directly** rather than
   re-written — so this suite inherits all five near-miss shapes the guard
   rejects and cannot drift away from the guard it mirrors.

2. **`chart/tests/*.sh — platform/wordpress-tenant`** —
   `FileNotFoundError: platform/wordpress-tenant/chart/../image/Dockerfile`.
   Both release gates invoke `bash <script> <chart_dir>` with a chart_dir
   **relative to the repo root**, and the scripts `cd` into it, so every later
   `$CHART_DIR/...` resolved against the new cwd. It passed locally only
   because the default argument is already absolute — my local invocation could
   not reproduce CI's. `CHART_DIR` is now resolved absolute **before** any `cd`
   in both suites, and all six were re-run with CI's exact relative-path
   invocation.

3. **`manifest-validation`** — `TestCatalogSeed_DeliveryPinsNotBehindComponentCharts`
   and `TestCatalogSeed_DisplayVersionNotBehindDeliveryPin`. Holding the seed
   pin back at 0.4.23 is refused by canon, so the original reasoning for doing
   so was wrong. Synced with the repo's own script; both tests pass locally.

4. **`vitest — sovereign-admin console`** — **not this PR.** The job died in
   `Set up job` before running a single test:
   `An action could not be found at the URI '…/actions/setup-node/tar.gz/8207627…'`.
   An Actions download failure, not a test result.

https://claude.ai/code/session_01HvMZTDx7W6EPZMmjyXmYh1

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1451,7 +1451,7 @@ spec:
       #   staged commit and push, and makes the Phase-3 pre-strip read-back the
       #   single fail-closed gate (#5436 relocated, not weakened). Lockstep w/
       #   products/catalyst/chart/Chart.yaml.
-      version: 1.4.1497
+      version: 1.4.1499
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/platform/wordpress-tenant/blueprint.yaml
+++ b/platform/wordpress-tenant/blueprint.yaml
@@ -6,7 +6,11 @@ metadata:
     catalyst.openova.io/category: organization-app
     catalyst.openova.io/section: pts-7-org-tenant
 spec:
-  version: 0.4.23
+  # 0.4.24 (#6311, Refs #3988): CNPG-operator status-probe carve-out for the
+  # per-Org default-deny namespace + zero install-time third-party egress
+  # (artifacts baked into the runtime image). Lockstep with
+  # platform/wordpress-tenant/chart/Chart.yaml.
+  version: 0.4.24
   card:
     title: WordPress (per-Organization)
     summary: |
--- a/platform/wordpress-tenant/chart/Chart.yaml
+++ b/platform/wordpress-tenant/chart/Chart.yaml
@@ -345,7 +345,84 @@ name: bp-wordpress-tenant
 #   values file. Singleton installs render byte-identical (no annotation, no
 #   declaration required). Catalog-wide enforcement:
 #   scripts/check-dr-pairs-declare-promotion.sh.
-version: 0.4.23
+# 0.4.24 (#6311, Refs #3988 #4282 #3379): bp-wordpress-tenant could not
+#   converge in ANY Organization. Two independent defects, both fixed here.
+#
+#   DEFECT 1 — no cnpg-system carve-out, so our own NetworkPolicy blinds the
+#   CNPG operator. This chart renders its Cluster into the Organization's own
+#   namespace, which carries the `default-deny-all` + `allow-same-org` pair
+#   stamped by the Organization controller's GitOps tree. `allow-same-org`
+#   re-opens SAME-NAMESPACE traffic only, so the cnpg-system operator — a
+#   different namespace — is denied the instance Pod's :8000/pg/status probe.
+#   The database is UP and serving while the operator is blind, and CNPG parks
+#   the Cluster at `Instance Status Extraction Error: HTTP communication issue`
+#   / readyInstances 1/1 / ConsistentSystemID=False → Ready=False → the HR
+#   never converges → the customer's purchased WordPress never comes up.
+#   MEASURED on hw296 with quota eliminated as a variable: the identical
+#   blueprint installed into an Org with headroom ran the exact Pods #5393
+#   rejects elsewhere (`wordpress-db-1-initdb` Completed, `wordpress-db-1` 1/1
+#   Running, zero quota rejections) and STILL landed Ready=False. Control
+#   sharing the suspect property: `walkone/postgres` — same deny pair, but
+#   carrying the #4282 carve-out — read `Cluster in healthy state` 3/3; 11 CNPG
+#   Clusters cluster-wide were healthy and this was the only one in the error
+#   phase.
+#   FIX — templates/networkpolicy-cnpg-operator-probe.yaml, the proven #4282
+#   shape from platform/postgres (whose header predicts this phase string
+#   verbatim). Opens ONLY :8000 + :9187 from ONLY `cnpg-system`. Additive:
+#   default-deny-all and allow-same-org are untouched. No broad 5432 rule (the
+#   #4442 `sharedConsumers` shape) — this Cluster's only consumers are
+#   same-namespace and allow-same-org already admits them. SINGLETON ONLY, like
+#   #4282: in active-hot-standby the replica streams WAL cross-cluster over
+#   ClusterMesh and an Ingress policyType selecting the Cluster Pods would
+#   default-deny that stream, while a remote-cluster peer cannot be expressed
+#   in a k8s NetworkPolicy at all (an ipBlock never matches a ClusterMesh
+#   remote endpoint identity — bp-postgres needs a dedicated identity-based
+#   CiliumNetworkPolicy for it). Rendering nothing in AHS keeps that path
+#   byte-identical rather than trading one break for another; the AHS carve-out
+#   stays open on #6311.
+#   SCOPE — 8 charts render a CNPG Cluster with no carve-out, but the deny pair
+#   exists only in per-Org namespaces. The load-bearing property is "can this
+#   chart's Cluster land in a namespace carrying that pair", and only two
+#   charts can: bp-postgres (routed per-Org by #4398, already carries the
+#   #4282 carve-out) and this one. The other 7 (bp-gitea, bp-guacamole,
+#   bp-llm-gateway, bp-openova-flow-server, bp-powerdns, bp-powerdns-admin,
+#   bp-temporal) are bootstrap-kit components in platform namespaces where no
+#   such pair is stamped — and where an unconditional probe NetworkPolicy would
+#   REGRESS them, because policyTypes:[Ingress] makes the Cluster Pods
+#   default-deny for every unlisted peer and their consumers live in OTHER
+#   namespaces (the #4442 regression, verbatim). They are deliberately
+#   untouched.
+#
+#   DEFECT 2 — install-time github.com fetch, which contradicts Pillar 5.
+#   The post-install oidc-config Job CrashLoopBackOff'd on
+#   `curl: (28) Connection timed out after 120001 milliseconds`. Cutover
+#   step-08 (`egress-block-test`) holds a 10-minute deny-egress NetworkPolicy
+#   against github.com / ghcr.io / harbor.openova.io and requires the cluster
+#   to reconcile green through it — that hold IS the sovereignty proof
+#   (ADR-0002). A customer Application that reaches GitHub to install cannot
+#   exist on a sovereign Sovereign. FOUR install-time fetch sites existed, not
+#   two: deployment.yaml's wp-plugin-install init (downloads.wordpress.org +
+#   github.com), oidc-config-job.yaml step 0 (github.com) and step 4
+#   (github.com), and step 7's `wp theme install` fallback — always reachable,
+#   because the Job's emptyDir masks the image's bundled themes.
+#   SEAM REUSED, NOT INVENTED — this repo mirrors exactly two artifact classes
+#   into a post-cutover Sovereign's Harbor: OCI container images and Helm
+#   charts (cutover step-03 `harbor-prewarm`). There is NO mirroring path for
+#   plain file artifacts. So the bytes now travel as layers of an image already
+#   in the prewarm set: `wordpress-tenant-pg`, which
+#   platform/self-sovereign-cutover/chart/Chart.yaml already names verbatim as
+#   a tag+digest-pinned prewarm source.
+#   FIX — image/Dockerfile bakes pinned pg4wp + openid-connect-generic into
+#   /usr/src/openova-wp-artifacts with build-time assertions (including a uid-33
+#   read/copy probe); both consumers copy from there and FAIL CLOSED if it is
+#   absent, with no network fallback; core's bundled themes are seeded into the
+#   Job so step 7 never fetches. The plugin slug is now the bare
+#   `openid-connect-generic` in BOTH places — the pre-existing mismatch (Job
+#   activated `openid-connect-generic`, PVC held `daggerhart-openid-connect-
+#   generic`) left active_plugins pointing at a file the runtime PVC never had.
+#   The oidc-config Job also gains imagePullSecrets: its new wp-artifacts init
+#   runs the PRIVATE ghcr image, and without them it ImagePullBackOffs.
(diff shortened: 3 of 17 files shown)


emrahbaysal:
Adjudication (Refs #6360): a duplicate of this PR's #4282 CNPG-operator-probe fix was raised as #6497 and has been closed in favour of THIS PR, which is the superset — it also fixes row 222 DEFECT TWO (zero install-time egress: the Dockerfile bakes the pg4wp + OIDC plugins so the oidc-config Job no longer fetches from github.com, a Pillar-5 violation on a cutover Sovereign) and is singleton-only to avoid AHS ClusterMesh WAL-stream interference.

This PR is the delivery for UAT row 222 (wordpress-tenant) + rows blocked by the per-Org WordPress not serving. It is currently 635 commits behind main (mergeable=CONFLICTING) and the `integration (kind + gitea)` check is failing — it needs a rebase onto main (re-derive the umbrella/seed/generated-catalog lockstep) + the integration failure investigated before it can land.

emrahbaysal:
Duplicate of merged #6500 (28d6575d9). Refs #6311

PR #6322, opened 2026-08-14
fix(agenity): renew the Anthropic OAuth credential before it expires, and deliver the renewal to a running workspace

emrahbaysal:
## The defect, measured by USE

On hw296 2026-08-14 a real message was delivered to the Agenity workspace and the agent answered on its own PTY:

```
Please run /login - API Error: 401 OAuth access token has been revoked.
```

Decoded from the live credential — expiry and byte lengths only, never material:

| Field | Value |
|---|---|
| `claudeAiOauth.expiresAt` | **2026-08-14T04:28:12Z** — expired |
| `refreshTokenExpiresAt` | **2026-09-10** — the refresh material was **alive and unspent** |
| `anthropicApiKey` | **byte-identical** to the access token (108 B, same sha256 prefix, both `sk-ant-oat…`) — no long-lived fallback underneath |

Nothing regressed. G8/G9 were genuinely green at 00:14Z / 00:31Z and red by morning because the credential simply died. **Rows 219 / 220 / 221 / R19 / G8 / G9 all flap on this one cause.**

`anthropic_credential_class.go` already said so in its own words — *"the seeded blob is a claudeAiOauth pair whose accessToken lives HOURS and **nothing in this repo refreshes it**, so an unrotated credential spends most of its life in the INVALID class."* The classifier was right. Nothing acted on it.

## The two credential paths, named exactly

| | Path |
|---|---|
| **Axon's CronJob reads** | Secret `axon.claudeAuthSecret` (release namespace), key `.credentials.json`, via `jsonpath='{.data.\.credentials\.json}'` — one Secret, one namespace, chart-owned |
| **Agenity actually consumes** | Secret `agenity-anthropic-token` in **each per-Org namespace**, keys `anthropicApiKey` + `credentialsJson`, **ExternalSecret-managed** (`creationPolicy: Owner`) from OpenBao `secret/catalyst/anthropic/token`, whose root is the K8s Secret `catalyst-system/sovereign-anthropic-credentials` |

Same blob shape at both ends (`claudeAiOauth` with `accessToken`/`refreshToken`/`expiresAt`), so no shaped write is needed — but the **write target does not transfer**, which is why this is a design change and not a CronJob re-point.

## Would the seeding producer clobber a refresh? — Yes, in one specific direction

`reconcileGlobalSeed` gates on `anthropicStoredCredentialUsable`, so a **healthy** refreshed value is a no-op: no clobber while fresh. The danger is elsewhere. The exchange **rotates the refresh token**. Had the refresh written only OpenBao, the root Secret would keep the **spent** token, and the moment the stored blob was judged unusable the reconciler would re-seed that dead material over a working path — unrecoverable without a human re-issue.

**So the fix writes the root Secret FIRST, then OpenBao.** If only the OpenBao write fails, the next pass reads the refreshed root, finds the path unhealthy, and the existing seed leg repairs it — self-healing, and the error is reported rather than retried in place.

## Three broken links, not one

```
catalyst-system/sovereign-anthropic-credentials   (root Secret)
  -> catalyst-api seedAnthropicToken (live read, every pass)
OpenBao secret/catalyst/anthropic/token
  -> ExternalSecret agenity-anthropic-token        [2] refreshInterval 1h
<org-ns>/agenity-anthropic-token
  -> init container seed-claude-creds              [3] COPIES ONCE at pod start
emptyDir ~/.claude/.credentials.json               <- what every agent spawn reads
```

1. **No refresh existed** — every producer faithfully re-applies whatever the root holds.
2. **ESO `refreshInterval: 1h`** against a ~5h token — flagged in R19's own stamp.
3. **A running Pod never sees a refreshed Secret.** `anthropic-creds` is mounted at `/creds` in the **init container only**; the main container mounts the `claude-home` emptyDir. Renew upstream all you like — the agent kept reading the frozen file.

## What ships

- **`sovereign_anthropic_refresh.go`** — exchanges the refresh token when **< 2h** of life remains, and when already expired (the hw296 state). Runs from `runSeedReconcilePass` **before** the seed leg.
- **Root Secret first, then OpenBao** (above). Document **edited in place**, so `scopes` / `subscriptionType` / `rateLimitTier` / `refreshTokenExpiresAt` all survive — the Axon shell script greps fields out and rebuilds, dropping the rest.
- **`apiKey` rewritten ONLY when byte-identical to the access token.** An independent `sk-ant-api03-…` key is preserved. In default OAuth mode the StatefulSet omits `ANTHROPIC_API_KEY` entirely (it is set only in key-only mode), so a stale `apiKey` is inert there — but it is still the access token, and is kept consistent.
- **Never spends a token it cannot store** — a pre-flight refuses the exchange when there is no durable store, because a rotation with nowhere to write burns the only material that could renew the credential later.
- **RBAC** — name-scoped `get/update/patch` on `sovereign-anthropic-credentials`. `create` deliberately not extended.
- **ESO `refreshInterval` 1h -> 15m**, and a **`creds-resync` sidecar** carrying a **newer** credential into the running Pod every 2m — no pod restart, no killed sessions. Newer-only, so a credential `claude-code` rotated for itself is never clobbered.

## Cadence and margin

| | |
|---|---|
| accessToken lifetime | ~5h (measured) |
| renew at | **2h remaining** |
| worst-case propagation | 15m ESO + ~1m kubelet re-projection + 2m re-sync = **~18m** |
| slack | **~1h42m** |
| attempts inside the window | **12**, at the 10m reconcile cadence |

The OAuth endpoint can be unreachable for over an hour and the credential still turns over before it dies. An hourly-or-worse schedule against a ~5h token guarantees a dead window on any single missed run.

## Failures are loud

Every failure path returns a distinct outcome **value** and logs at ERROR with its remediation: `anthropic refresh IMPOSSIBLE` (no refreshToken), `FAILED` (exchange), `FAILED TO PERSIST` (write — naming that the old token is already spent), `SKIPPED` (no durable store). A **200 carrying no access_token is treated as a failure**, never stored. A silent no-op leaving an expired token in place is the one outcome the design forbids.

## Every assertion is vacuity-proven

`scripts/check-anthropic-refresh-vacuity-6317.py` mutates **one behaviour at a time** and requires the **named** test to go red, rejecting any mutation that fails to compile (which would prove nothing). It found **two genuinely vacuous assertions** in the first draft and both were repaired:

- the HTTP-error test asserted on `"HTTP 400"`, which the no-access-token path **also** emits — widening the status gate left it green. It now pins the status gate's own message, which also records that the error body is **withheld** from the log.
- the ordering test asserted on the final stored value, which is **identical** whether the refresh runs before or after the seed. It now asserts on what the producer **saw**.

`products/agenity/chart/tests/credential-resync-6317.sh` does the same for the sidecar predicate — adopts newer, refuses older, no-ops on equal, never destroys a working credential on an absent/empty Secret — with a drift guard that fails if the predicate under test stops matching the one the chart renders.

## Not claimed

Delivery to hw296 is severed (#6307), so **no UAT row is stamped from this work**. A merged fix is not a live one. The row impact is reported for the ledger lane to act on: R19 / G8 / G9 / 219 / 220 / 221 share this single recurring cause.

Refs #6317 #4277 #4111 #3988

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1451,7 +1451,7 @@ spec:
       #   staged commit and push, and makes the Phase-3 pre-strip read-back the
       #   single fail-closed gate (#5436 relocated, not weakened). Lockstep w/
       #   products/catalyst/chart/Chart.yaml.
-      version: 1.4.1486
+      version: 1.4.1493
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/products/agenity/README.md
+++ b/products/agenity/README.md
@@ -221,22 +221,46 @@ diagnose this without exec'ing into the pod:
 ```
 
 (A valid token instead logs `claude-code OAuth token valid (~Nh remaining).`)
-The check is **diagnostic-only** — it never blocks the pod (the dashboard must
-keep serving) and never mutates the blob (`claude-code` owns rotation).
+Since #6163 FREEZE 3 the check is **no longer diagnostic-only**: an expired
+credential takes the same verdict as an absent one under
+`anthropic.onExpiredCredential` (`fail`|`continue`, default `fail`). It still
+never mutates the blob.
 
-**When the agent reports offline, re-seed:** mint a **fresh**
-`credentialsJson` (a current `claude` login on a workstation writes
-`~/.claude/.credentials.json`; copy that whole blob), `bao kv put
-secret/catalyst/anthropic/token credentialsJson='…'` as above (or rotate the
-`sovereign-anthropic-credentials` Secret so the next Org-create re-seeds it),
-force-sync the ExternalSecret
-(`kubectl annotate externalsecret agenity-anthropic-token
-force-sync=$(date +%s) --overwrite`), and restart the StatefulSet pod (or wait
-for the next restart) so the init container re-seeds the new blob. The next
-spawn then authenticates. This is the standing **operator activation cost** of
-the OAuth journey until upstream `claude-code` gains a reliable
-non-interactive refresh — track it as a recurring per-Sovereign chore, not a
-one-time seed.
+### ✅ The platform now REFRESHES that credential automatically (#6317)
+
+The manual re-seed described above was, until chart `0.5.29` /
+`bp-catalyst-platform 1.4.1493`, a **standing per-Sovereign chore on a ~5h
+timer**. It is no longer. `catalyst-api`'s seed reconciler renews the
+credential itself, and three links were closed to make the renewal actually
+reach a running agent:
+
+| Link | What changed |
+|---|---|
+| **Renew** | `refreshAnthropicCredential` (`sovereign_anthropic_refresh.go`) exchanges the `refreshToken` at `console.anthropic.com/v1/oauth/token` when **less than 2h** of the accessToken's life remains — and also when it has already expired, since the refresh token outlives it by weeks. It runs on every 10m seed-reconcile pass, **before** the seed leg, so the same pass propagates the renewed value. |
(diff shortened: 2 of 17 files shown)


emrahbaysal:
Duplicate of merged #6498 (fd1d40322): the Anthropic OAuth refresh already ships in catalyst-api. The remaining gap is the INPUT (expired + revoked credential), tracked on #6778. Refs #6317

PR #6337, opened 2026-08-14
docs(uat): G11 walked on hw297 and stays red, with the root cause measured

emrahbaysal:
Walked **G11** against hw297 (dep `256cd1bc167224be`), region-b. **The verdict does not move.**

## What the walk measured

```
kubectl -n catalyst get configmap self-sovereign-cutover-status  -> NotFound (in every namespace)
kubectl get jobs -A | grep cutover                               -> none
kubectl -n catalyst-system get pods                              -> No resources found
```

G11 requires the 11-step chain at `cutoverComplete=true`. There is no cutover, and no control plane to run one.

## Why — measured, not suspected (#6336)

The bootstrap `GitRepository flux-system/openova` could not clone: `context deadline exceeded`. `spec.timeout` is **60s**; the clone of a **938 MB** full-history repo takes **~11 minutes**. All three Kustomizations sat at `Source artifact not found`, and the cluster held **0 HelmReleases for 75 minutes**.

**Network ruled out** rather than assumed — from a pod in `flux-system`: `github.com` → `20.233.83.145`, `nc -z github.com 443` → **TCP-OK**, all six Flux controllers `Running`, zero restarts.

**Proof by single-variable experiment**: patching only `spec.timeout: 60s → 10m` produced `ready=True rev=main@sha1:e050028d` at 13:52:24Z and took the cluster from **0 → 67 HelmReleases in 99 seconds**.

## No green claimed

Ledger unchanged: **278 proven / 6 failing / 2 unwalked = 286**. A measured failure is evidence and belongs in the ledger; it is not a reason to move a verdict.

Refs #6336


emrahbaysal:

PR #6340, opened 2026-08-14
docs(uat): rows 60, 166, 227 carry the pre-walk delivery verification

emrahbaysal:
**No verdict moves.** This records, per row, that the code each depends on is confirmed present in the artifacts the next prov installs — and that the walk itself has not happened.

## What was verified (all `git merge-base --is-ancestor` against `origin/main`)

| Row | Fix | Ships in |
|---|---|---|
| **60** | `98f44954b` #6291 region-B standby producer · `aa6eea36d` #6301 cross-cluster CNPG pair resolver | `bp-catalyst-platform`, slot-13 pin **1.4.1506** |
| **166**, **227** | `df1737827` #6332 deny-egress hold cannot report PASS unapplied · `ce5cdbff0` #6334 per-Org vcluster registry pivot | cutover chart **0.1.190**, slot 06a |

`check-bootstrap-kit-pin-sync.sh` → **PASS**, 67 chart→pin pairs, so the pins are not hollow. This is the check that decides whether the next prov can flip these rows at all: had a pin been stale, the environment would have converged clean and the rows would still have failed.

## What is absent: the walk

- hw297 was wiped after its Phase-1 watch timed out (#6336). On it, `kubectl get applications.catalyst.openova.io -A` returned *the server doesn't have a resource type*, and `GET /deployments/{id}/cutover` returned **404** with the `self-sovereign-cutover-status` ConfigMap NotFound in every namespace.
- The mothership cannot substitute — measured: `kubectl get blueprints.catalyst.openova.io -A` returns *the server doesn't have a resource type*, because `console.openova.io` is Catalyst-Zero and registers **no catalyst CRDs at all**.

## Guards, exit codes read directly

- `check-walk-respects-scheduler.sh origin/main hw296` → **OK**; this change flips nothing.
- `uat-drift-guard.py` → **OK**.

Ledger unchanged: **278 proven / 6 failing / 2 unwalked = 286**.

Refs #6336


emrahbaysal:

PR #6346, opened 2026-08-15
docs(uat): row 227 — delivery pivot VERIFIED, refusal assertion untested

emrahbaysal:
The precondition this row depends on is **proven**; the thing it actually asserts is not — so the verdict does not move.

## Measured on hw298

```
gitrepositories -A   -> flux-system/openova = https://gitea.hw298.omantel.biz/openova/openova
helmrepositories -A  -> every bp-* = oci://registry.hw298.omantel.biz/openova-io

CONTROL: URLs matching ghcr.io|github.com  ->  0
```

Zero survivors. That is **absence of a tether**, not absence of a measurement — cutover steps 04 (registry-pivot), 05 (flux-gitrepository-patch) and 06 (helmrepository-patches) did exactly what they claim, and it holds even though the chain did not complete.

## Why the row stays ☐

The clause asserts a **post-cutover Sovereign REFUSES a GitHub-side catalog bump** — the Blueprint CR must not move when the upstream pin advances. Proving that requires advancing a pin upstream and observing the CR stay put. I measured the precondition (sources pivoted), not the assertion (refusal on a live bump).

The clause is also scoped *post-cutover*, and hw298 is at `cutoverComplete: false`, 10/11 steps, with `egress-block-test` failed.

## Guards

`check-walk-respects-scheduler.sh origin/main hw298` → **OK**. `uat-drift-guard.py` → **OK**. Ledger unchanged: **278 / 6 / 2 = 286**.

Refs #4415 #3379


emrahbaysal:

Issue #6347, opened 2026-08-15
placement projection emits TWO Primaries for an active-hot-standby app, plus a duplicate empty-cluster Standby (hw298 spine-gitea/keycloak, post-#6345)


emrahbaysal:
## Measured live on hw298 (dep `2540d866403f1f7c`) AFTER #6345 merged

`GET /sovereign/api/v1/sovereigns/{depID}/applications/{name}/placement?namespace={ns}`, mothership rolled to `catalyst-api:8c41df2`:

| app | targets | shape | derivedFromRuntime |
|---|---|---|---|
| **shared-pg** (CONTROL) | 2 | Primary(cluster set) + Standby(cluster set) | true |
| spine-gitea | **3** | **Primary(set) + Primary(set)** + Standby(cluster **empty**) | true |
| spine-keycloak | **3** | **Primary(set) + Primary(set)** + Standby(cluster **empty**) | true |

#6345 fixed the identity join, so the Primary now resolves — and that exposed the next term. An
`active-hot-standby` Application must project ONE Primary and ONE Standby; `bpv1.DerivePattern`
labels a 2-Primary list `active-active`, and `bpv1.ValidatePlacement` rejects it outright as
`MultiPrimaryNotSupported` under the default `primary+standby` capability. The projection is
emitting a placement its own validator would refuse.

## Two terms, both confirmed in the source

**1. The stateless role arm cannot tell primary from standby.**
`derivePlacementTargets` (`applications_placement_runtime.go:785-788`) ends with

```go
default:
    // Stateless: every occupied region serves traffic → Primary.
    t.Role = bpv1.DataRolePrimary
```

`shared-pg` escapes it only because CNPG stamps `openova.io/cnpg-role` on each half, so the
`anyCNPG` arm above assigns the roles from positive per-leg evidence. A Deployment-backed app has
no equivalent marker, and `10-gitea.yaml` / `09-keycloak.yaml` carry **no**
`catalyst.openova.io/region-role` gate — "no gate ⇒ every region" — so gitea and keycloak really do
run in both region clusters. Two occupancies, no role evidence, both Primary.

The declared posture is the evidence that exists: `placement.Resolve`
(`core/controllers/internal/placement/placement.go:255-263`) puts `regions[0]` Primary and
`regions[1..]` Standby, and `placementRegionCountError` states the same rule back to any caller that
writes a one-region multi-region placement.

**2. The empty-`cluster` Standby is appended over legs that already cover its region — because the
[...]

Issue #6344, opened 2026-08-15
placement projection joins Pods to an Application by NAME COINCIDENCE — apps whose pods carry a different identity collapse to a lone empty-cluster Standby (hw298 spine-gitea/keycloak/openbao)


emrahbaysal:
## Measured live on hw298 (dep `2540d866403f1f7c`), owner session, same binary, same pass

`GET /sovereign/api/v1/sovereigns/{depID}/applications/{name}/placement?namespace={ns}`

| app | ns | declared placement | targets returned | derivedFromRuntime | unresolvedPrimary |
|---|---|---|---|---|---|
| **shared-pg** | shared-data | active-hot-standby | **2 — Primary(cluster set) + Standby(cluster set)** | **true** | — |
| spine-gitea | catalyst | active-hot-standby | 1 — Standby, empty `cluster` | false | **true** |
| spine-keycloak | catalyst | active-hot-standby | 1 — Standby, empty `cluster` | false | **true** |
| spine-openbao | catalyst | active-passive | 1 — Standby, empty `cluster` | false | **true** |

All four Applications are `phase: Ready`. `shared-pg` is the CONTROL: the projection
works, so #6291 / #6301 / #6268 are present and functioning. The three that fail
differ from it in exactly one way.

## Root cause — pod matching keys on the ROUTE ID, not on the app's real workload identity

`derivePlacementTargets` (`applications_placement_runtime.go:672`) fans out over
`h.k8sCache.Clusters()` and keeps a Pod only when `podBelongsToComponent(p, name, ns)`
says so (`:710`). That predicate (`:843`) compares `componentNameCandidates(name)` —
which is just `{name, strings.TrimPrefix(name, "bp-")}` (`:816`) — against
`applicationKey(p, nil)` (the `app.kubernetes.io/instance` label, `dashboard.go:711`),
`app.kubernetes.io/name`, and the loft display-name prefix.

So a Pod is joined to an Application by STRING COINCIDENCE between the route id and a
label the chart happens to stamp.

- `shared-pg` passes because the CNPG Cluster it owns is itself named `shared-pg`, so
  the pods' identity resolves back to the same string as the app.
- `spine-gitea` fails because it is an Application CR (`post_handover_spine_apps.go`)
  named `spine-gitea` in ns `catalyst`, which ADOPTS HelmRelease `flux-system/bp-gitea`
  (`spec.helmRelease`). That HR declares `releaseName: gitea`, `targetNamespace: gitea`
[...]

Issue #6268, opened 2026-08-13
Topology tab shows one target for an active-hot-standby app whose backing pair is healthy — the placement BFF drops the leg it cannot resolve


emrahbaysal:
## The placement BFF returns one target while its own `regionsObserved` says 2

UAT row 60 (#3375) asserts: Catalog → New instance → `active-hot-standby` → Provision → that app's Topology tab shows the chosen topology.

Walked live on hw296 (2026-08-14). **The row is red on a surface its own fix never touched.**

### Below the console, everything is correct

An `active-hot-standby` app was provisioned from the Catalog. The backing state is right:

- primary **+ standby** — not the two-primaries shape of #6200
- `armed=true`
- Continuum CR reports `Healthy` and holds the lease
- CNPG 3/3
- the Application reads `Ready`

### The console still reports it as un-paired

The Topology tab renders:

- `Pattern: not reported`
- **one** target card — no primary, therefore no pair
- the **Switchover** control **disabled**

The placement BFF returns a single target while its own `regionsObserved` field says **2**, and the returned target carries an empty `cluster`.

### Why this is a read-path defect, not a write-path one

- **Control:** the bootstrap-seeded `shared-pg` app returns *both* targets through the same endpoint and renders correctly. The tab is fine; the projection is not.
- Writing `spec.placement.targets[]` **explicitly changed nothing** — so the read path re-derives rather than reading what was persisted, and silently drops the leg it cannot resolve.
- Three readings 17 minutes apart, identical. Not a convergence race.

### Relationship to prior work

#6136 fixed the Topology-tab **Save** so `spec.placement` persists as an object with `targets[]` (verified green this walk as row 16: generation 1→2→3 with `observedGeneration` following). #6257 fixed `application_controller.go:872` reading `spec.Topology` via `NestedString` against a field the CRD types as an object.

Both are write/controller-side and both are working. This is the **sovereign-api placement handler's read projection** — an unfixed third leg.

### The shape to be careful about

Dropping a leg that cannot be resolved makes an absent target indistinguishable from a single-region app. Any fix should make the unresolvable case **loud** — report the leg with its resolution failure rather than omitting it — otherwise a two-region app silently renders as one-region, which is exactly what this row exists to catch.

Refs #3375 #3986

PR #6348, opened 2026-08-15
fix(placement): one Primary for an active-hot-standby app, and no phantom third leg (hw298 row 60)

emrahbaysal:
## What was measured

hw298 (dep `2540d866403f1f7c`), mothership on `catalyst-api:8c41df2` — i.e. AFTER #6345 landed:

```
shared-pg      active-hot-standby → 2 targets  Primary(cluster set) + Standby(cluster set)   CONTROL, correct
spine-gitea    active-hot-standby → 3 targets  Primary(set) + Primary(set) + Standby(cluster "")
spine-keycloak active-hot-standby → 3 targets  Primary(set) + Primary(set) + Standby(cluster "")
```

`bpv1.DerivePattern` reads a 2-Primary list as `active-active`; `bpv1.ValidatePlacement` rejects it
outright as `MultiPrimaryNotSupported` under the default `primary+standby` capability. The
projection was emitting a placement its own validator refuses, and UAT row 60 wants a 2-region pair.

## The two terms, both confirmed in source before anything moved

**1. The stateless role arm has no discriminator.** `derivePlacementTargets` ends with *"every
occupied region serves traffic → Primary"*. `shared-pg` escapes only because CNPG stamps
`openova.io/cnpg-role` on each half, so its roles come from positive per-leg evidence. gitea and
keycloak have no equivalent marker, and `10-gitea.yaml` / `09-keycloak.yaml` carry **no**
`catalyst.openova.io/region-role` gate — "no gate ⇒ every region" — so they genuinely run in both
region clusters and both legs were labelled Primary.

**2. The empty-`cluster` leg is appended over a region the live legs already cover, because the two
sides never spell a region the same way.** `continuumStandbyRegion` skips a standby region that a
Primary already names — comparing raw. The per-app `dr-<app>` Continuum carries the bare cloud
region `me-east-215-b` (spine `spec.regions[]` is `RegionSpec.CloudRegion`,
`post_handover_spine_apps.go:619`) while a runtime leg carries the `openova.io/region` node label
`hw-me-east-215-b-rtz-prod`. #5482 recorded all three divergent spellings on one Application. The
guard was there; it could not see. The controller already folds the two vocabularies
(`normalizeRegion`, `core/controllers/application/internal/controller/placement_projection.go`) —
`internal/` puts that package out of reach from this module, so the fold is mirrored here token set
for token set, with its own table.

Both readings in the issue comment were right, and they are linked: term 2 is only *reachable*
because term 1 left the list with no Standby, and it only *fires* because of the vocabulary split.

## What changed

The role of a leg the runtime cannot classify now comes from the app's own declaration —
`placement.Resolve` and `placementRegionCountError` both state that `regions[0]` is the primary and
`regions[1..]` are the standbys, so this is the rule the write doors already enforce, read back on
the projection side. The declared list is read off the same Application CR through the same
selector the identity resolution uses (extracted to `applicationCRForComponent`, one CR and one
Organization-isolation rule for both consumers).

Three lines it does not cross, each pinned by its own case in the table:

- **It never overrules an observation.** A failed-over CNPG pair still reports its live primary in
  region B while `spec.regions[0]` names region A. The cnpg-role labels are evidence; a declaration
  is not, and a projection that tracked config here would point an operator at the region that no
  longer holds the write path.
- **It never invents a leg.** An app occupied only in its declared standby region now answers with a
  lone Standby and `unresolvedPrimary: true` (#6268) instead of promoting that leg to writer — and
  the #4551 declaration-derived Standby is *still* emitted when region B is genuinely empty, because
  there it is the only standby signal that exists.
- **It stays silent where it has no standing.** No Application CR, no `spec.placement`, a `singleton`
  or `active-active` posture, or fewer than two distinct regions ⇒ behaviour byte-for-byte as
  before. `active-active` is multi-primary by definition and keeps both Primaries.

## Red first, on a clean tree

The endpoint table was run against `git archive origin/main` before any source change:

```
--- FAIL: TestPlacementDeclaredRoles_Row60_6347/spine-gitea_shape:_occupied_in_BOTH_regions...
    got 3 target(s) (2 Primary, 1 Standby) want 2 (1 Primary, 1 Standby)
    targets=[{Region:hw-me-east-215-a-rtz-prod Cluster:dep-6347-gitea            Role:Primary}
             {Region:hw-me-east-215-b-rtz-prod Cluster:dep-6347-gitea-me-east-215-b Role:Primary}
             {Region:me-east-215-b             Cluster:                          Role:Standby StandbyType:Hot}]
    derivedFromRuntime=true unresolvedPrimary=false
--- FAIL: .../occupied_only_in_the_declared_STANDBY_region...
    got 2 target(s) (1 Primary, 1 Standby) want 1 (0 Primary, 1 Standby)
    targets=[{Region:hw-me-east-215-b-rtz-prod ... Role:Primary}
             {Region:me-east-215-b Cluster: Role:Standby StandbyType:Hot}]
TRUE-RC=1
```

The first block is the hw298 wire reproduced exactly, third leg and all. The second is the same
defect at its purest: a "cross-region pair" whose two legs are one region spelled two ways.

Post-fix, six cases green (`-count=1`), and the full `internal/handler` package green in 329s:

```
--- PASS: TestPlacementDeclaredRoles_Row60_6347
    --- PASS: .../spine-gitea shape: occupied in BOTH regions, declared active-hot-standby
    --- PASS: .../shared-pg shape: CNPG role labels carry the split
    --- PASS: .../failed-over CNPG pair: observed role beats the stale declaration
    --- PASS: .../single-region app: no fabricated standby
    --- PASS: .../standby region genuinely empty: the Continuum-derived leg is still emitted
    --- PASS: .../occupied only in the declared STANDBY region: honest half-pair, not a false primary
ok  github.com/openova-io/openova/products/catalyst/bootstrap/api/internal/handler  329.361s
TRUE-RC=0
```

Every case also asserts the projected list passes `bpv1.ValidatePlacement` under
`primary+standby` — the invariant the live answer was violating.

UAT row 60 stays ❌ until an operator walk on a live Sovereign stamps it; this PR is the source fix,
not the stamp.

Refs #6344
Refs #3375
Refs #6347
Refs #6345

https://claude.ai/code/session_01HvMZTDx7W6EPZMmjyXmYh1

### DIFF
--- a/products/catalyst/bootstrap/api/internal/handler/applications_placement_declared_roles_6347.go
+++ b/products/catalyst/bootstrap/api/internal/handler/applications_placement_declared_roles_6347.go
@@ -1,0 +1,242 @@
+// Package handler — applications_placement_declared_roles_6347.go (#6347 /
+// UAT row 60).
+//
+// THE DEFECT, measured on hw298 (dep 2540d866403f1f7c) AFTER #6345 landed and
+// the mothership was rolled to `catalyst-api:8c41df2`:
+//
+//	shared-pg      active-hot-standby → 2 targets  Primary(cluster set) + Standby(cluster set)
+//	spine-gitea    active-hot-standby → 3 targets  Primary(set) + Primary(set) + Standby(cluster "")
+//	spine-keycloak active-hot-standby → 3 targets  Primary(set) + Primary(set) + Standby(cluster "")
+//
+// #6345 made the Primary resolve — and nothing downstream turns two occupied
+// regions into ONE Primary and ONE Standby. `bpv1.DerivePattern` reads a
+// 2-Primary list as `active-active`, and `bpv1.ValidatePlacement` rejects it
+// outright as `MultiPrimaryNotSupported` under the default `primary+standby`
+// capability: the projection was emitting a placement its own validator
+// refuses.
+//
+// WHY THE CONTROL ESCAPES, AND WHY THAT IS NOT A PROPERTY OF THE APP.
+// `derivePlacementTargets` assigns roles from the `openova.io/cnpg-role` label
+// when it sees one — positive, per-leg evidence of which half serves writes.
+// `shared-pg` has it; a Deployment-backed app does not. Its last arm therefore
+// reads "every occupied region serves traffic → Primary", which is a fair
+// reading for a bootstrap component that declares nothing, and the wrong one
+// for an app that declares `active-hot-standby`. gitea and keycloak really do
+// run in both region clusters — `10-gitea.yaml` / `09-keycloak.yaml` carry no
+// `catalyst.openova.io/region-role` gate, and "no gate ⇒ every region".
+//
+// WHAT THIS FILE ADDS. The declaration the app already carries, used ONLY where
+// no runtime evidence exists. `placement.Resolve`
+// (core/controllers/internal/placement/placement.go) puts `regions[0]` Primary
+// and `regions[1..]` Standby for both asymmetric modes; `placementRegionCountError`
+// states the same rule back to any caller that tries to write a one-region
+// multi-region placement ("regions[0] is the primary and regions[1..] are the
+// standbys"). So regions[0] is not a heuristic invented here — it is the rule
+// the write doors already enforce, read back on the projection side.
+//
+// THE THREE LINES IT WILL NOT CROSS:
+//
+//   - It never OVERRULES an observation. A CNPG pair that has failed over
+//     reports its live primary in region B while `spec.regions[0]` still names
+//     region A; the labels win, always. A declaration is not evidence, and a
+//     projection that tracked config here would point an operator at the region
+//     that no longer holds the write path — worse than the defect being fixed.
+//   - It never INVENTS a leg. Everything here re-labels targets that occupancy
+//     already produced. An app with no Pods in a region still has no target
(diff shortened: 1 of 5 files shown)


emrahbaysal:
### CI: 22 pass / 1 skipping (e2e) / 0 fail

The first `integration (kind + gitea)` run went red. Naming the item rather than waving it off:

```
--- FAIL: TestRestoreFromStore_StartupKicksClusterMeshReconcile (0.03s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat
      /tmp/TestRestoreFromStore_StartupKicksClusterMeshReconcile1746139662/002: directory not empty
```

Not an assertion — `testing` itself failing to remove the second `t.TempDir()` (the
`store.New(t.TempDir())` directory in `clustermesh_test.go:917`) while the startup cluster-mesh
reconcile it kicked was still writing records into it. A cleanup race in that test's own harness.

Three pieces of evidence that it is not this change:

1. This PR touches the placement projection and the Application-CR selector. It writes no files,
   starts no goroutine, and never reaches the store, `kubeconfigsDir`, or clustermesh.
2. The full `internal/handler` package under the SAME build tag the failing job uses, on this
   branch: `ok … 326.854s`, `TRUE-RC=0`.
3. The job was re-run with no code change and passed.

Refs #6347

Issue #6374, opened 2026-08-16
bp-agenity credentialWait: 0 silently renders as 300/5 via sprig default — 'fail immediately' becomes a 5-minute wait


emrahbaysal:
`products/agenity/chart/templates/statefulset.yaml` resolved two NUMERIC knobs with sprig `default`:

```
_cw_timeout={{ (.Values.anthropic.credentialWait).timeoutSeconds | default 300 }}
_cw_interval={{ (.Values.anthropic.credentialWait).intervalSeconds | default 5 }}
```

`default` substitutes on any **falsy** value, and **0 is falsy**. So `timeoutSeconds: 0` rendered as **300** and `intervalSeconds: 0` as **5**.

An operator setting `timeoutSeconds: 0` — "do not wait, fail immediately if the credential is absent" — silently got a 5-minute wait. Worse, the init container's own log line echoes `${_cw_timeout}`, so it printed **300** back and the setting looked applied.

Same trap as #5991 in bp-guacamole, where `false | default true` yielded `true` and deleted the JDBC schema-seed Job at render time.

## Fix

Resolve by **key presence** (`hasKey`) rather than truthiness. `onTimeout` deliberately KEEPS `default` — an empty verdict string is not a meaningful setting and should fall back to `fail`.

## Render proof

| input | before | after |
|---|---|---|
| `timeoutSeconds=0` | 300 | **0** |
| `intervalSeconds=0` | 5 | **0** |
| unset | 300 / 5 | 300 / 5 |
| `timeoutSeconds=42` | 42 | 42 |

New suite `chart/tests/credential-wait-falsy-default-6374.sh` renders the **zero case first** — a suite that only passes non-zero values cannot fail on this defect, the same blind spot that let #5991 live. Controls are included so the fix cannot pass by over-correcting the defaults away.

Verified to FAIL against pre-fix main:

```
FAIL timeoutSeconds=0 stays 0 — expected '_cw_timeout=0'
FAIL intervalSeconds=0 stays 0 — expected '_cw_interval=0'
```

Found while investigating #6372 (per-Org agenity Pending on `seed-claude-creds`). Distinct defect: #6372 is a missing credential producer; this is the wait knob being unsettable.

Issue #6372, opened 2026-08-16
Per-Org agenity never starts: nothing seeds the Anthropic credential Secret, so seed-claude-creds waits 300s and fails by design (missing producer, NOT a chart bug)


emrahbaysal:
## `seed-claude-creds` has ZERO env — the Anthropic credential is never wired in

Measured on hw299 after creating Organizations through the **BSS/admin door** (`POST /api/v1/organizations`).

`bp-agenity-0` never starts. It sits `Pending` on its init container:

```
Initialized=False  reason=ContainersNotInitialized
  containers with incomplete status: [seed-claude-creds]
```

The init container itself:

```
init seed-claude-creds  ready=False  restarts=2  state=running
     lastTerminated: exit=1  reason=Error
     spec.image = ghcr.io/openova-io/agenity:0.9.7
     env keys   = []          <-- EMPTY
```

**No `ANTHROPIC_API_KEY`, no `secretKeyRef`, no env at all** on the container whose entire job is seeding the Claude credential. It exits 1, restarts, and the pod never initializes.

Reproduced identically in **two** Organizations (`bssdoor` plan s, `bssdoorm` plan m), so it is not plan- or isolation-specific.

## What this is NOT

- Not a missing workload. The BSS door provisions agenity correctly: StatefulSet exists, `oidc-gate-agenity-*` is Running 1/1, and the door answers **302** (SSO redirect) at `agenity.<org>.omani.homes`. Route, gate and chart are all fine.
- Not a funnel gap. Note the asymmetry: the **BSS door provisions agenity, the funnel door does not** — funnel-born `mailwalk` has no agenity StatefulSet at all. That is a separate issue (#6366 — agenity absent from the storefront).

## Correction to earlier UAT evidence

UAT rows **G8/G9** were stamped this session as "no agentic runtime exists — 0 agenity pods among 420". That measurement was taken against the **funnel** path only and was true for it. It is **wrong as a general claim**: via the BSS door the runtime IS created, and it fails at exactly the surface G8 names — credential seeding. The rows are re-stamped accordingly.

## Relationship to #6365

PR #6365 adds validate-by-use so a *revoked* token fails loudly instead of passing presence/expiry checks. This defect is upstream of that: there is no credential in the container to validate. Both are needed; #6365 does not fix this.

Refs UAT G8, G9.

emrahbaysal:
## CORRECTION — my root-cause statement above is WRONG on mechanism

[...]

PR #6375, opened 2026-08-16
fix(bp-agenity): credentialWait 0 silently rendered as 300/5 — sprig default on a numeric knob

emrahbaysal:
`products/agenity/chart/templates/statefulset.yaml` resolved two NUMERIC knobs with sprig `default`:

```
```

`default` substitutes on any **falsy** value, and **0 is falsy**. So `timeoutSeconds: 0` rendered as **300** and `intervalSeconds: 0` as **5**.

An operator setting `timeoutSeconds: 0` — "do not wait, fail immediately if the credential is absent" — silently got a 5-minute wait. Worse, the init container's own log line echoes `${_cw_timeout}`, so it printed **300** back and the setting looked applied.

Same trap as #5991 in bp-guacamole, where `false | default true` yielded `true` and deleted the JDBC schema-seed Job at render time.

## Fix

Resolve by **key presence** (`hasKey`) rather than truthiness. `onTimeout` deliberately KEEPS `default` — an empty verdict string is not a meaningful setting and should fall back to `fail`.

## Render proof

| input | before | after |
|---|---|---|
| `timeoutSeconds=0` | 300 | **0** |
| `intervalSeconds=0` | 5 | **0** |
| unset | 300 / 5 | 300 / 5 |
| `timeoutSeconds=42` | 42 | 42 |

New suite `chart/tests/credential-wait-falsy-default-6374.sh` renders the **zero case first** — a suite that only passes non-zero values cannot fail on this defect, the same blind spot that let #5991 live. Controls are included so the fix cannot pass by over-correcting the defaults away.

Verified to FAIL against pre-fix main:

```
FAIL timeoutSeconds=0 stays 0 — expected '_cw_timeout=0'
FAIL intervalSeconds=0 stays 0 — expected '_cw_interval=0'
```

Found while investigating #6372 (per-Org agenity Pending on `seed-claude-creds`). Distinct defect: #6372 is a missing credential producer; this is the wait knob being unsettable.

Refs #6374


emrahbaysal:
Superseded by #6470, which carries the same credentialWait fix with the complete lockstep set.

This PR moved three files (Chart.yaml, statefulset.yaml, the render test) and stopped there. The gates showed that is not the whole set for a bp-agenity bump — the precedent commits 5850c7614 and 2472dcaec also move blueprint.yaml, the generated catalog (blueprints.json + catalog.generated.ts), the slot-13 pin, and the umbrella Chart.yaml. #6470 has all of them, and both the bootstrap-kit lockstep sweep and the funnel pin-vs-seed test pass locally against it.

It also carried unrelated ledger churn (UAT.md, WBS-TO-100.md, three CSVs, a session evidence file) that does not belong with a chart fix.

Closing so the 0.5.30 claim is held by exactly one branch — the version-collision gate was correctly refusing to let two open PRs claim it.

PR #6467, opened 2026-08-18
fix(bp-agenity): credentialWait 0 silently rendered as 300/5 — sprig default on a numeric knob (#6374)

emrahbaysal:
fix(bp-agenity): credentialWait 0 silently rendered as 300/5 — sprig default on a numeric knob (#6374)

`products/agenity/chart/templates/statefulset.yaml` resolved two NUMERIC knobs
with sprig `default`:

`default` substitutes on any FALSY value, and 0 is falsy. So
`timeoutSeconds: 0` rendered as 300 and `intervalSeconds: 0` as 5.

An operator setting `timeoutSeconds: 0` — "do not wait, fail immediately if the
credential is absent" — silently got a 5-minute wait. Worse, the init
container's own log line echoes ${_cw_timeout}, so it printed 300 back and the
setting looked applied.

Same trap as #5991 in bp-guacamole, where `false | default true` yielded true
and deleted the JDBC schema-seed Job at render time. Third instance of this
class in the codebase.

FIX: resolve by key PRESENCE (`hasKey`). `onTimeout` deliberately KEEPS
`default` — an empty verdict string is not a meaningful setting and should fall
back to `fail`.

RENDER PROOF

    timeoutSeconds=0   300 -> 0
    intervalSeconds=0    5 -> 0
    unset              300 / 5 unchanged
    timeoutSeconds=42   42 unchanged

The new suite `chart/tests/credential-wait-falsy-default-6374.sh` renders the
ZERO case FIRST — a suite that only passes non-zero values cannot fail on this
defect, the same blind spot that let #5991 live. Controls included so the fix
cannot pass by over-correcting the documented defaults away.

Reopened on a clean origin/main base: the previous branch was cut from a stale
fork and dragged unrelated ledger commits into conflict. Chart 0.5.30 avoids the
version collision with the concurrently-open agenity work.

Refs #6374

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1451,7 +1451,7 @@ spec:
       #   staged commit and push, and makes the Phase-3 pre-strip read-back the
       #   single fail-closed gate (#5436 relocated, not weakened). Lockstep w/
       #   products/catalyst/chart/Chart.yaml.
-      version: 1.4.1518
+      version: 1.4.1519
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/products/agenity/chart/Chart.yaml
+++ b/products/agenity/chart/Chart.yaml
@@ -241,7 +241,7 @@ type: application
 # same gateway bp-wordpress-tenant + the org console route attach to, #4054).
 # The old default rendered Accepted=False/InvalidHTTPRoute and the host 404'd
 # until the parentRef was hand-overridden. appVersion (image) unchanged.
-version: 0.5.28
+version: 0.5.30
 # 0.5.22 (#5617): the per-Org oidc-gate companion gains its EGRESS
 # CiliumNetworkPolicy (templates/oidc-gate.yaml `oidc-gate-<cid>-egress`).
 # The gate shipped an ingress-only gateway CNP into a DEFAULT-DENY per-Org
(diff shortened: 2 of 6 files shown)


emrahbaysal:
## Lockstep state — what is done and what remains

Recording this on the PR so it is not rediscovered from scratch.

**Done (pushed):**
- `products/agenity/chart` 0.5.28 → **0.5.30** + the credentialWait fix and its test suite
- catalog-seed synced with the repo's own `scripts/sync-catalog-seed-pin.py` (self-test green) so spec.version and spec.source.version move together — 2 lines, no hand-editing
- umbrella `products/catalyst/chart` 1.4.1518 → **1.4.1519** at all three sites: Chart.yaml `version`, Chart.yaml `appVersion`, and the slot-13 pin in `clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml`
- `scripts/check-chart-version-bumped.sh` reads **OK products/agenity/chart: 0.5.28 -> 0.5.30**

**Remaining — 4 red gates:**

| gate | why |
|---|---|
| `run the release writer and assert all five sites converge` | I moved **three** sites by hand; the lockstep has **five**. The canonical writer must be run instead — `scripts/check-release-lockstep-writer.py` is the *checker* that proves it moved all five, not the writer itself. Find and run the writer rather than hand-editing the remaining two. |
| `pin-sync-audit` | almost certainly the same two unmoved sites |
| `manifest-validation` | same |
| `ghcr-pin-existence` | expected until `bp-agenity:0.5.30` is published — the chart publishes on merge, so this resolves at land or needs a declared entry in `scripts/expected-unpublished-catalog-seed-pins.yaml` |

**Do not** hand-edit the last two sites. The whole point of the five-site writer is that partial hand-edits are exactly how hollow pins ship — #6363 hit this and so did I here.

The fix itself is unaffected and vacuity-proven: reverted, the suite fails with `timeoutSeconds=0 stays 0 — expected '_cw_timeout=0'`.

Issue #6464, opened 2026-08-18
PIN sign-in dead program-wide: a 2-3s retry storm pins the noreply@ 25/hr sender quota at zero (452), plus d16-export loops on the reaped hw299


emrahbaysal:
## PIN sign-in is dead program-wide — a retry storm holds the `noreply@` sender quota at zero

Measured on the mothership 2026-08-18T09:42Z.

### The gate

Stalwart's own log, verbatim:

```
Rate limit exceeded (smtp.rate-limit-exceeded)
  listenerId = "submission"  localPort = 587
  accountName = "noreply@openova.io"  accountId = 38
  id = "sender"   limit = [25, 3600000ms]
  to = "emrah.baysal@openova.io"
```

**25 messages per hour** on `noreply@openova.io`. Every `POST /auth/pin/issue` therefore returns:

```
502 {"error":"email-send-failed","smtpCode":452,"detail":"mail server temporarily unavailable"}
```

No customer — and no walker — can obtain a sign-in PIN.

### Why the quota never recovers

The log shows a sustained storm from `212.72.24.85`, each attempt completing AUTH + MAIL FROM before being refused:

```
09:42:04  09:42:08  09:42:10  09:42:13  09:42:15  09:42:19  09:42:21  09:42:23
```

That is roughly one attempt every 2–3 seconds. A 25/hour budget is consumed in about a minute and then stays exhausted, so the limit is never actually "per hour" in practice — it is permanently spent. Earlier the OTP folder showed the same shape on the delivery side: five identical `wordpress is ready on Mailwalk` notifications at 07:44:20/23/26/29/34.

### A second retry loop against a dead environment

catalyst-api is simultaneously looping on an env that no longer exists:

```
d16-export: POST failed (will retry)  attempt 5
  id = e13300fc41a33a57  region = me-east-215-b-1
  url = https://api.hw299.omantel.biz/api/v1/sovereign/secondary-kubeconfig
  err = dial tcp: lookup api.hw299.omantel.biz on 10.43.0.10:53: server misbehaving
```

`hw299.omantel.biz` is **NXDOMAIN** — the deployment was reaped but its record still drives retries.

### What is NOT the cause

[...]

PR #6468, opened 2026-08-18
docs(uat): the ledger named a Sovereign that no longer exists — rotate the header to the honest no-live-env state

emrahbaysal:
docs(uat): the ledger named a Sovereign that no longer exists — rotate the header to the honest no-live-env state

The header described `hw298.omantel.biz` as the walk environment. hw298 is
WIPED, and its successor hw299 is now NXDOMAIN — reaped, DNS gone, all four
doors answer 000. A header naming a dead machine is worse than no header: every
stamp below reads as current while describing a Sovereign that does not exist,
which is exactly how a walker ends up "confirming" rows against nothing.

No verdict cell is touched. 286 rows, 243 green, unchanged. This changes only
what the header CLAIMS about the environment.

What the new header records, because the next walker needs all three:

1. NO LIVE ENV. Do not walk or stamp until a fresh Sovereign is fired and the
   header is rotated to it.

2. PIN SIGN-IN IS QUOTA-STARVED FLEET-WIDE, so even a fresh env cannot be
   authenticated into yet. Stalwart enforces [25, 3600000ms] on noreply@ and a
   producer retrying every 2-3s drains the hour's budget in about a minute, so
   /auth/pin/issue returns 502/452 permanently. Fixed by #6465 (isRateLimit
   matched only 503 5.5.1, so a 452 skipped both the backoff and the breaker).

3. DELIVERY STATE of the 7 fixes merged 2026-08-16..18, which is what decides
   whether a red row is code-gated or merely undelivered:
     LIVE (catalyst-api 8c41df2 -> 1bbe8c3): #6363, #6366, #6368, #6371
     MERGED, NOT DEPLOYABLE:                 #6465, #6373, #6466
   The latter three built SUCCESSFULLY (b8111c6, e299dc7) but publish to
   ghcr.io/openova-io/openova/services while the Deployment pulls
   ghcr.io/openova-io/openova/catalyst-api. That path mismatch is what holds
   delivery — not a failing build. Naming it here so the next session does not
   re-derive it.

CONSEQUENCE, stated so nobody hunts for work that is already done: of the 18 ❌
rows, NONE has outstanding code work. Every one is delivery-gated or env-gated.

Refs #6464


emrahbaysal:
## ⚠️ This turns the drift guard red — stale 0 → 74. Read before merging.

Measured after the change: `scripts/uat-drift-guard.py` exits 1 with **74 stale rows**.

**Why**, and it is a property of the guard rather than a defect in the rows: the guard resolves the current environment from the UAT.md H1 via `header_env()`, then flags any ✅ row whose evidence cites a *different* env. By removing the env token from the header — which is the whole point of this change, because no env exists — there is no current env to compare against, so every row still citing hw298/hw299 reads as foreign.

**No row's evidence changed.** 286 rows, 243 green, byte-identical verdict cells. The delta is entirely the header.

**So there is a genuine design question here, and I do not want to answer it unilaterally:**

The guard assumes a live environment always exists. That assumption held while the fleet was continuous, but it is false in exactly the situation this PR documents — the gap between one Sovereign being reaped and the next being fired. Right now the repo can express "walked on hw298" and "walked on hw299" but cannot express "there is nothing to walk", so the only way to keep the guard green is to keep naming a machine that does not exist. That is the failure mode this PR exists to stop.

**Three ways forward, and I think (a) is right:**

- **(a)** Teach  a sentinel — e.g. a header declaring NO-LIVE-ENV puts the guard in a mode where predecessor-cited greens are *carried*, not stale, and stamping is refused outright. That makes the no-env state a first-class, expressible thing.
- **(b)** Merge this as-is and accept a red guard until the next env is fired and the header rotates. Honest, but leaves a red gate on main.
- **(c)** Keep a dead env named in the header to keep the gate green. This is what main does today and it is precisely the thing that produced a ledger describing a machine that had been wiped.

I have not implemented (a) in this PR because it changes guard semantics that gate every future walk, and that deserves its own review rather than riding along on a header edit.

The underlying facts are unchanged and independently verifiable: hw299 is NXDOMAIN, and PIN sign-in returns 502/452 fleet-wide against the `[25, 3600000ms]` `noreply@` limit.

emrahbaysal:
Withdrawing the sentinel idea I floated on this PR. I went to add one and the guard turned out to be right.

## The guard is behaving correctly

`scripts/uat-drift-guard.py`:

```python
cur = current_env.lower() if current_env else None
...
stale = sorted(h for h in hits if h != cur)
```

`header_env()` returns `None` when the H1 declares no concrete env. With
`cur = None`, every `hwNNN` token in an evidence cell differs from it, so all 74
✅ rows flag as stale.

That is the correct verdict, not a bug. With no live environment, no ✅ is
currently proven by anything. The 74 are stale in exactly the sense the guard
exists to catch.

It is also internally consistent with the reset path: a RESET header reads
"pending hwXXX" and also resolves to `None`, but `scripts/reset-uat.py` clears
the verdict cells first, so there are no ✅ left to flag and the guard reports 0.

## So the defect is this PR, not the guard

My H1 rotation declares NO LIVE ENV while leaving 74 ✅ rows in place. That is a
self-contradicting ledger state: it simultaneously claims nothing is currently
proven and that 74 things are proven. The guard caught precisely that.

Adding a sentinel to suppress it would have been making a correct gate quiet to
turn my own PR green — the exact anti-pattern the repo's own gate-discipline
notes warn about, and I nearly shipped it.

## Correct resolutions, neither of which is a guard change

1. **Reset properly** — run `scripts/reset-uat.py`, which clears the verdict
   cells so the "no live env" claim and the row states agree. Costs the recorded
   evidence, which is the honest price of having no environment.
2. **Use the carried-forward mechanism that already exists** — `scan_text` takes
   a `carried` set from the confidence scheduler for rows legitimately held
   across an env change. If these 74 are meant to survive, they belong in that
   set, with the scheduler deciding, rather than in a blanket exemption.

I am not shipping a guard change. Parking this PR pending a call on (1) vs (2);
the branch for the sentinel is dropped.

emrahbaysal:
Closing with live evidence: the premise ("hw299 is NXDOMAIN — reaped, all doors 000") is refuted. hw299 was **alive the whole time**; only its DNS was dead, and the DNS died on the **mothership**, not on the Sovereign.

Root cause chain (verified live 2026-08-19 ~06:00Z):
- `omantel.biz` NS -> ns1/ns2.openova.io -> 45.151.123.50 = mothership `powerdns-anycast`.
- `openova-system/pdns-pg-1` was ImagePullBackOff since ~2026-08-16 evening: its image routes through `harbor.openova.io/proxy-ghcr/...` and Harbor answered 503.
- Harbor itself was down in a circular pull dependency (every Harbor component's image references the Harbor proxy it serves), so nothing could self-recover after the 2026-08-16 restart wave.
- `powerdns` crash-looped 554x on "connection refused" to its DB -> dnsdist had no backend -> SERVFAIL for every mothership-served zone -> every `hw299.omantel.biz` lookup read as NXDOMAIN/000.

Meanwhile the deployment record held `status: ready`, the CP EIP answered ping, and `curl --resolve console.hw299.omantel.biz:443:212.72.24.35` returned **200** throughout.

Repaired 2026-08-19 ~06:25Z (pdns-pg/harbor-pg CNPG imageName -> upstream ghcr; Harbor components -> upstream docker.io; stuck pods recreated). Re-verified via public DNS (1.1.1.1): console / marketplace / api doors all **200**.

The ledger has been rotated to hw299 on main (`894e4b593`) with `reset-uat.py hw299` + the 6 orphaned hw299 stamps from `uat/hw299-228` restored, guards green. A "NO LIVE ENV — do not walk" header would now be factually wrong and would block the walk on a healthy env.

Full evidence: `docs/sessions/2026-08-19/lane-b-prov-decision-hw299-kept.md` on main.

PR #6470, opened 2026-08-18
fix(bp-agenity): credentialWait 0 silently rendered as 300/5 — sprig default on a numeric knob (#6374)

emrahbaysal:
fix(bp-agenity): credentialWait 0 silently rendered as 300/5 — sprig default on a numeric knob (#6374)

`products/agenity/chart/templates/statefulset.yaml` resolved two NUMERIC knobs
with sprig `default`:

`default` substitutes on any FALSY value, and 0 is falsy. So `timeoutSeconds: 0`
rendered as 300 and `intervalSeconds: 0` as 5.

An operator setting `timeoutSeconds: 0` — "do not wait, fail immediately if the
credential is absent" — silently got a 5-minute wait. Worse, the init container's
own log line echoes ${_cw_timeout}, so it printed 300 back and the setting looked
applied.

Same trap as #5991 in bp-guacamole, where `false | default true` yielded true and
deleted the JDBC schema-seed Job at render time. Third instance of this class.

FIX: resolve by key PRESENCE (`hasKey`). `onTimeout` deliberately KEEPS `default`
— an empty verdict string is not a meaningful setting and should fall back to
`fail`.

RENDER PROOF

    timeoutSeconds=0   300 -> 0
    intervalSeconds=0    5 -> 0
    unset              300 / 5 unchanged
    timeoutSeconds=42   42 unchanged

The suite renders the ZERO case FIRST — a suite that only passes non-zero values
cannot fail on this defect, the same blind spot that let #5991 live. Controls
included so the fix cannot pass by over-correcting the documented defaults away.

SCOPE — WHY THIS PR TOUCHES ONLY THREE FILES.

My earlier attempt also hand-edited the catalog-seed, the umbrella Chart.yaml and
the slot-13 kit pin, which reddened four gates. That was doing CI's job, badly.
Per scripts/check-release-lockstep-writer.py the five lockstep sites divide as:

  site 1  <chart>/Chart.yaml `version:`            <- the deploy-bot / this PR
  sites 2-5  blueprint.yaml, catalog-seed, the generated catalog,
             the bootstrap-kit pin                 <- .github/workflows/blueprint-release.yaml,
                                                      AFTER it publishes the chart

So a human PR bumps site 1 and nothing else; the writer converges the rest
post-publish. Hand-moving a subset guarantees the drift those gates exist to
catch.

Refs #6374

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1451,7 +1451,7 @@ spec:
       #   staged commit and push, and makes the Phase-3 pre-strip read-back the
       #   single fail-closed gate (#5436 relocated, not weakened). Lockstep w/
       #   products/catalyst/chart/Chart.yaml.
-      version: 1.4.1520
+      version: 1.4.1521
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/docs/ledger/TRACKER.md
+++ b/docs/ledger/TRACKER.md
@@ -4,7 +4,7 @@ Regenerated every 15 min by `/home/openova/bin/refresh-dod-dashboard.sh`. Every 
 
 |  |  |
 |---|---|
-| Last refreshed | `2026-08-18T14:00:07Z` |
+| Last refreshed | `2026-08-18T14:30:06Z` |
 | Deploy cron (#799) | ✓ deploy-cron healthy (image-reroll last ran 10m ago) |
 | Open issues | 100 |
 | Open DoD gates | 0 / 41 |
--- a/products/agenity/blueprint.yaml
+++ b/products/agenity/blueprint.yaml
@@ -18,7 +18,7 @@ spec:
   # Keycloak backchannel + upstream) — the per-Org namespace is default-deny
   # and the gate's ingress-only CNP left the OAuth callback 500ing on a DNS
   # timeout. Lockstep with products/agenity/chart + the catalog-seed pin.
-  version: 0.5.28
+  version: 0.5.30
   card:
     title: Agenity
     summary: |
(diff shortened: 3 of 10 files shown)


emrahbaysal:
This PR cannot merge, and the reason is not its content: **zero check-runs fire on its head**.

## Evidence

```
#6470 head bb1ae0c88c1c   check-runs = 0
#6472 head 564e9e0ae811   check-runs = 59
#6473 head a0bedfeb44b3   check-runs = 84
```

Actions is firing normally on sibling PRs opened in the same session, minutes
apart, so this is specific to this branch — not a repo-wide outage and not a
quota wall.

Ruled out along the way:
- **Stale reporting** — the API is queried against the PR's actual head SHA, and
  PR / local / remote all agree on it.
- **Empty commit** — an earlier retrigger touched no files, which matches no
  `paths:` filter and correctly ran nothing. The latest commit edits
  `products/agenity/chart/tests/credential-wait-falsy-default-6374.sh`, a real
  tracked file under a filtered path, and still produced zero runs.
- **Rerun replaying a stale payload** — this is a fresh `synchronize`, not a rerun.

## The content is verified good

Pre-flight (the check that would have caught two duplications elsewhere this
session) confirms the fix is genuine and not already on main:

```
git show origin/main:products/agenity/chart/templates/statefulset.yaml | grep -c hasKey   -> 0
grep -c hasKey products/agenity/chart/templates/statefulset.yaml                          -> 2
```

Local gates:
- render suite `PASS` — `timeoutSeconds: 0` renders `_cw_timeout=0`, unset renders `300`
- `check-chart-version-bumped.sh` → `PASS: all 2 touched chart(s) bumped their version`
- bootstrap-kit lockstep sweep → `rc=0`
- funnel pin-vs-seed test → `rc=0`
- `check-chart-version-not-claimed-by-open-pr.py` → `OK — no open PR claims any of these versions`

So the six-site lockstep is correct and the sprig-falsy defect is fixed; what is
missing is CI actually running.

## What this needs

Someone with repo Actions visibility should check whether workflow runs are being
created for this branch at all (branch-level restriction, a stuck queue, or a
required-approval state). Recreating the PR from a fresh branch would likely also
clear it, at the cost of losing this thread — worth doing if the cause is not
quickly visible.

I am deliberately not force-pushing again. Two blind retriggers have already
produced nothing, and a third would be repeating an action whose failure mode I
have not yet explained.

PR #6471, opened 2026-08-18
fix(cutover): deny mail.openova.io in the step-08 hold — the 9th tether (#5921)

emrahbaysal:
A Sovereign with `cutoverComplete=true` still routes every customer sign-in code through **our** mail server. `catalyst-system/marketplace-api` reads `smtp-host = mail.openova.io` from `Secret/sovereign-smtp-credentials`, so the purchase path depends on the mothership after the platform has declared independence.

ADR-0002 and `docs/DOD.md` §Pillar 5 enumerate **eight** tethers that cutover pivots. Outbound customer-auth SMTP is not among them, so no step pivots it — and it was equally invisible to the proof meant to catch exactly this. Step-08 holds deny-egress against `github.com`, `ghcr.io`, `harbor.openova.io` and `console.openova.io` for 600s; `mail.openova.io` was not in that set and no reconcile needs SMTP, so the window stays green with the tether fully live. Same shape as #5919: sovereignty asserted from the absence of a host nobody put on the list.

## What this changes

Adds `mail.openova.io` to `egressTest.blockedDomains`, following the `console.openova.io` precedent from #3678 — that entry exists to drive an **active call-home assertion** rather than rely on blocking-by-omission. The hold now fails while the SMTP tether is live instead of passing over it.

## What this does not change

This is the **detection** half. It makes the defect visible to the proof and to CI. It does not repoint the Sovereign at its own mail server — that is the pivot half and needs a cutover step of its own, which is the remaining work on #5921.

## Gates

- `check-chart-version-bumped.sh` → `PASS: all 1 touched chart(s) bumped their version`
- bootstrap-kit lockstep sweep → green locally

Lockstep sites were taken from this chart's own precedent commit rather than inferred. The umbrella takes 1.4.1522 because #6470 already claims 1.4.1521.

Refs #5921

### DIFF
--- a/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
+++ b/clusters/_template/bootstrap-kit/06a-bp-self-sovereign-cutover.yaml
@@ -1074,7 +1074,7 @@ spec:
       # chart/tests/ left the package via .helmignore: 757,264 bytes, 72.2%.
       # The rendered manifest is byte-identical modulo comments — no step,
       # value or RBAC change, still 11 steps.
-      version: 0.1.191
+      version: 0.1.192
       sourceRef:
         kind: HelmRepository
         name: bp-self-sovereign-cutover
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1451,7 +1451,7 @@ spec:
       #   staged commit and push, and makes the Phase-3 pre-strip read-back the
       #   single fail-closed gate (#5436 relocated, not weakened). Lockstep w/
       #   products/catalyst/chart/Chart.yaml.
-      version: 1.4.1520
+      version: 1.4.1522
       sourceRef:
         kind: HelmRepository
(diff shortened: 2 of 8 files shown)


emrahbaysal:
Closing as redundant — `origin/main` already carries this change.

I branched off origin/main and rebuilt a fix that was already there. main's
`platform/self-sovereign-cutover/chart/values.yaml` already lists:

```
    - "mail.openova.io"
```

with a comment referencing #5921 and calling it "the NINTH tether" — wording I did
not write, so another session shipped the detection half before me.

How this surfaced: a guard I wrote for the tether class has a control that must
FAIL against main. It passed instead. The easy read was "my control is broken";
chasing it found the entry already on main. Worth noting the control did its job
on first run, just not the way intended.

This is my second duplication today (PR #6375 was the first, and that one was a
duplicate of my own earlier PR). The corrective is a pre-flight assertion that the
target state is ABSENT from origin/main before building, not "check harder".

The pivot half is NOT redundant and stays open as #6473: main has the deny entry,
but nothing repoints the Secret, so the hold now fails on a tether no step
resolves. #6473 adds cutover step-07a to close that.

PR #6472, opened 2026-08-18
fix(catalog): stop selling Sandbox — retired 2026-06-30, still listed (#5920)

emrahbaysal:
The marketplace grid still renders a **Sandbox** card — category `devtools`, badged FREE, live "Add to stack" button, counted in "16 available". The concept was retired by founder decision on 2026-06-30 and superseded by Agenity + `bp-openova-mcp` (Pillar 4).

## Why a source grep could not see it

The catalog-seed carries **no** sandbox Blueprint. The entry lives in the two *generated* catalogs the storefront actually serves — `blueprints.json` and `catalog.generated.ts` — both with `"visibility": "listed"`. Grepping the seed returns nothing and reads as absence, which is how this survived six weeks.

## Change

Flipped `visibility` to `unlisted` in both generated catalogs. Unlisted (not deleted) because it is the field the storefront already honours to drop a card from the customer grid while keeping the record addressable; deletion risks dangling references.

## Verification

- No Blueprint declares a `depends` on sandbox — checked, **NONE**
- Post-edit assertion: `sandbox` absent from the listed set, while `openclaw`, `stalwart-mail`, `wordpress` remain listed — this removes one card, not a category

Refs #5920

### DIFF
--- a/products/catalyst/bootstrap/api/internal/catalog/blueprints.json
+++ b/products/catalyst/bootstrap/api/internal/catalog/blueprints.json
@@ -4464,7 +4464,7 @@
       "category": "devtools",
       "tagline": null,
       "tags": [],
-      "visibility": "listed",
+      "visibility": "unlisted",
       "version": "0.3.12",
       "section": "pts-4-sandbox",
       "depends": [
--- a/products/catalyst/bootstrap/ui/src/shared/constants/catalog.generated.ts
+++ b/products/catalyst/bootstrap/ui/src/shared/constants/catalog.generated.ts
@@ -4599,7 +4599,7 @@ export const ALL_BLUEPRINTS: readonly BlueprintCardEntry[] = [
     "category": "devtools",
     "tagline": null,
     "tags": [],
-    "visibility": "listed",
+    "visibility": "unlisted",
     "version": "0.3.12",
     "section": "pts-4-sandbox",
     "depends": [
--- a/scripts/check-retired-products-not-listed.py
+++ b/scripts/check-retired-products-not-listed.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+"""Fail if a RETIRED product is still `visibility: listed` in a generated catalog.
+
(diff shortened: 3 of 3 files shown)


emrahbaysal:
Closing — this PR's unlist does not survive, and #6480 supersedes it with the fix at the correct layer.

## The defect in this PR

It edits the two GENERATED catalogs by hand. Those are outputs.
`products/catalyst/bootstrap/ui/scripts/build-catalog.mjs:364` reads
`spec.visibility` from each Blueprint's own `blueprint.yaml`, so the next
generator run rewrites both files from source and the card comes back.

Measured, not reasoned:

```
origin/fix/5920-sandbox-unlist : platform/sandbox/blueprint.yaml:32  visibility: listed
origin/fix/6475-unlist-qa-app  : platform/sandbox/blueprint.yaml:32  visibility: unlisted
```

Had this merged, `sandbox` would have been unlisted in the committed output and
still `listed` at source — a retired product back on the customer storefront on
the next regeneration, with a merged PR on the record saying it was removed. That
is worse than never having fixed it, because the audit trail would say it was
done.

## How it surfaced

The `regenerate and prove no diff` gate on #6480 rejected the same wrong-layer
edit for `qa-app`. Fixing qa-app at its source and regenerating flipped `sandbox`
BACK to listed — which is what exposed that this PR's change was never going to
hold. The gate was doing exactly its job; I had read it as a formality.

## What replaces it

#6480 carries all of it, at the source:

```
platform/qa-app/blueprint.yaml:28   listed -> unlisted
platform/sandbox/blueprint.yaml:32  listed -> unlisted
```

then regenerated, so committed outputs equal generator outputs:

```
[build-catalog] listed: 19  unlisted: 80  private: 0
sandbox listed? False   qa-app listed? False
CONTROL: openclaw + agenity still listed
check-retired-products-not-listed.py -> PASS (244 entries)
```

The retired-products guard this PR introduced is preserved in #6480 and extended
with a `qa-app` entry, so neither product can be re-listed without the gate going
red.

Refs #5920, #6475

Issue #6477, opened 2026-08-18
G8/G9/220/221 are INPUT-gated: the Anthropic credential is founder-supplied and empty by default


emrahbaysal:
Four UAT rows (G8, G9, 220, 221) will fail again on the next prov unless an Anthropic credential is supplied AT provision time. This is a provisioning input, not a code defect — recording it so a walk is not spent rediscovering it.

## The chain, traced on origin/main

`products/catalyst/chart/values.yaml:152-154`

```yaml
  anthropic:
    apiKey: ""
    credentialsJson: ""
```

`products/catalyst/chart/templates/catalyst-openova-kc-credentials-secret.yaml:283-289`

```
(`.Values.sovereign.anthropic.*`, empty by default -> founder-supplied ...)
{{- $anthSrc := lookup "v1" "Secret" $namespace "sovereign-anthropic-credentials" -}}
{{- $anthApiKey := (((.Values.sovereign).anthropic).apiKey | default "") -}}
```

So the credential comes from EITHER `.Values.sovereign.anthropic.*` at install
time OR a pre-existing `sovereign-anthropic-credentials` Secret. Both are empty
on a fresh Sovereign unless supplied.

Downstream everything is already correct and merged:
`seedAnthropicToken` writes OpenBao `secret/catalyst/anthropic/token`, and
`generateAgenityHR` emits the ExternalSecret that resolves it (#6372,
`e299dc7e4`). The chain is intact — it just has nothing at its head.

## Why the rows read as code failures

hw298's evidence for G8/G9 says "the anthropic credential ..." and for 218/223
"`bp-agenity` is ABSENT from the catalog". I verified bp-agenity IS present and
`visibility: listed` in both the generated catalog and the seed on main, so at
least part of that evidence describes a runtime state, not a source gap. The
credential half is genuinely absent — because nobody supplied it.

This is the third classification this session where a row that looked
code-gated was something else: 213 was already fixed, 218/223 are runtime-vs-
source, and these four are input-gated.

## What the next prov must do

Supply the credential in the deployment POST body (or pre-create
`sovereign-anthropic-credentials` in the catalyst namespace) BEFORE walking G8,
G9, 220, 221. Walking them without it measures a known-empty input and burns a
prov cycle.

## Optional hardening, not required

The empty case currently produces an agenity that installs and cannot chat. A
provision-time check that WARNs when both sources are empty would turn a silent
degradation into a visible one — same shape as the zero-image enumerator in
[...]

PR #6479, opened 2026-08-18
docs(uat): reclassify G8/G9/220/221 as blocked — INPUT-gated, not product faults (#6477)

emrahbaysal:
Tally: **18 ❌ → 14 ❌ + 4 ⛔**. No row moves toward green.

## These four are not defects

`values.yaml:152-154` sets `sovereign.anthropic.apiKey` and `credentialsJson` to `""`, and the secret template's own comment says why: *"empty by default → founder-supplied"*.

Everything downstream is merged and correct — `seedAnthropicToken` writes the OpenBao path, `generateAgenityHR` emits the ExternalSecret that resolves it (#6372, `e299dc7e4`). The chain has nothing at its head until an operator supplies the credential.

## This is the ledger's own established wall

Not a new category:

- rows **211-213 / 221-223** are already recorded ⛔ — *"BY-DESIGN gated (NOT a prov gap) … gated on founder Anthropic cred (#4277)"*
- a prior full-table walk lists *"Agenity live-run Anthropic cred #4277"* among its honest ⛔ walls

Same gate, same glyph.

## Why it matters

Carrying them as ❌ overstates the defect count **and** sends the next walker looking for a bug that is not there. ⛔ is the accurate verdict and it names what unblocks it: supply the credential at provision time, then re-walk.

## Care taken

- no row moved toward green
- each keeps its **original hw298 walk evidence**, with the reclassification appended — the measurement stays readable and the change is auditable
- `uat-drift-guard` → `rc=0`, run **before** push per canon

Refs #6477, #6372, #4277


emrahbaysal:
One gate red: **WBS §1 partition is derived from UAT.md**. Recording what I found rather than guessing at a fix, because the numbers do not line up the way the gate name implies.

`WBS-TO-100.md` §1 reads:

```
| ✅ PASS   | 206 |
| ❌ FAIL   |  77 |
| ☐ NOT RUN |   3 |
```

`UAT.md` on `origin/main` — BEFORE any change in this PR — parses to
**243 ✅ · 18 ❌ · 0 ⛔ · 25 ☐**. So the two documents already disagreed
(206 vs 243, 77 vs 18); the WBS table is evidently a different partition, not a
straight tally of the same rows.

That means I cannot re-derive §1 by copying this PR's new numbers
(**243 ✅ · 10 ❌ · 8 ⛔ · 25 ☐**) into it. Doing so would replace one set of
numbers I do not understand with another set I do not understand, and this file's
whole purpose is that its counts mean something specific.

I also could not locate the deriving code: no `scripts/*wbs*`, and no match for
`partition is derived` or `WBS §1` anywhere under `.github/`, `scripts/`, or the
Go trees. The check name appears to come from a workflow or action I cannot read
from here.

**What is needed:** whoever owns the WBS derivation should say which rows §1
counts and whether ⛔ is a category it recognises at all. The reclassification in
this PR introduces ⛔ rows where UAT.md previously had none, so if §1 folds ⛔
into ❌ the target is 18; if it excludes them it is 10; and if it counts a
different row set entirely, neither.

The UAT.md change itself stands on its own evidence and `uat-drift-guard` passes
(`rc=0`). Holding this PR rather than force-fitting a number into a file whose
derivation I have not read.

emrahbaysal:

Issue #6475, opened 2026-08-18
Storefront sells four Blueprints the catalog-seed does not carry — purchasable with nothing behind them


emrahbaysal:
Found while building a guard for #6360's defect class. Measured on `origin/main`, source-side only — no live env needed to reproduce.

## Four Blueprints are purchasable with nothing behind them

The storefront and the seed answer different questions:

- `products/catalyst/bootstrap/api/internal/catalog/blueprints.json` — what a customer can SEE and add to a stack
- `products/catalyst/chart/templates/catalog-seed/blueprints.yaml` — what a Sovereign actually INSTALLS

Nothing asserts the first is a subset of the second. It is not:

```
listed in storefront: 21     seed Blueprints: 81
LISTED but ABSENT from the seed: 6
   anthropic-adapter
   librechat
   netbird
   qa-app                 <- QA fixture, should never be customer-visible
   sandbox                <- retired 2026-06-30, being unlisted in #6472
   stalwart-sovereign
```

Each of the four real ones is absent from the seed under ANY key (0 grep hits,
not a parser artifact), and none appears in
`scripts/expected-unpublished-catalog-seed-pins.yaml`, which covers a different
set (bp-clickhouse, bp-opensearch, bp-prometheus, bp-redis, bp-n8n).

## Why this matters

A customer can add these to a stack and buy them. The Sovereign then has no
Blueprint to install, so the failure lands AFTER the sale rather than at the
point of sale. Same customer-visible shape as #6360 (post-cutover, the app is
sold and then ImagePullBackOffs) and the mirror of #5920 (a retired product
still listed).

`qa-app` is the sharpest of the four: a QA fixture reachable from the customer
storefront.

## What I did and did not do

Built `scripts/check-listed-blueprints-are-seeded.py`, which asserts the subset
property, carries a reasons-required exception register, and fails on a STALE
exception so the register cannot rot into a permanent waiver. It also fails on a
zero-length parse of either side rather than passing vacuously.

I did NOT unlist the four. Which of them should be seeded versus unlisted is a
catalog decision with revenue implications, and picking wrong either removes a
product that was meant to sell or leaves an unsellable one on the shelf. The
guard is useful either way; it just needs the four resolved before it can go
green in CI.

Suggested disposition, for whoever owns the catalog:
- `qa-app` — unlist (never customer-facing)
- `sandbox` — already being unlisted in #6472
- `anthropic-adapter`, `librechat`, `netbird`, `stalwart-sovereign` — seed the
[...]

PR #6480, opened 2026-08-18
fix(catalog): a QA fixture must not be purchasable — unlist qa-app (#6475)

emrahbaysal:
**"QA Test App"**, category `testing`, was `visibility: listed` in both generated catalogs — reachable from the customer storefront with a live Add-to-stack button. It is also absent from the catalog-seed, so a customer who bought it would receive nothing: the oversell #6475 describes.

## Why only this one

Of the four products that guard names, `qa-app` is the only one needing no judgement. Whether `anthropic-adapter`, `librechat`, `netbird` and `stalwart-sovereign` should be **seeded** or **unlisted** is a catalog decision with revenue implications, and stays with #6475. A QA fixture on the storefront is unambiguously wrong.

## Also

Registers `qa-app` in `check-retired-products-not-listed.py`, so the storefront cannot silently re-list it without the gate going red.

## Controls

| check | result |
|---|---|
| `qa-app` in listed set | **False** |
| `openclaw`, `agenity` still listed | yes — one card removed, not a category |
| retired-products guard | `PASS` — 244 entries, no retired product listed |

Builds on #6472 (`sandbox` unlist + the guard itself).

Refs #6475, #5920

### DIFF
--- a/platform/anthropic-adapter/blueprint.yaml
+++ b/platform/anthropic-adapter/blueprint.yaml
@@ -19,7 +19,7 @@ spec:
     tags: [llm, anthropic, claude, openai, adapter, proxy]
     documentation: https://docs.anthropic.com/en/api/messages
     license: Apache-2.0
-  visibility: listed
+  visibility: unlisted
   owner:
     team: ai-platform
     contact: ai-platform@openova.io
--- a/platform/librechat/blueprint.yaml
+++ b/platform/librechat/blueprint.yaml
@@ -18,7 +18,7 @@ spec:
     tags: [chat, llm, ui, openid, rag, application]
     documentation: https://www.librechat.ai/docs
     license: MIT
-  visibility: listed
+  visibility: unlisted
   owner:
     team: ai-platform
     contact: ai-platform@openova.io
--- a/platform/netbird/blueprint.yaml
+++ b/platform/netbird/blueprint.yaml
@@ -12,7 +12,7 @@ spec:
     summary: WireGuard-based zero-trust mesh + remote-access overlay. Operators / engineers / customer admins enroll devices via Keycloak SSO and reach Sovereign-internal services from any laptop. Includes management + signal + coturn (TURN/STUN). One NetBird per Sovereign per ADR-0001 §11.
     icon: netbird.svg
     category: platform
-  visibility: listed
+  visibility: unlisted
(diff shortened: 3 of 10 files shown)


emrahbaysal:
Status at `020e82607` — **fail=0, 58 pass, 6 pending**. Two gates that were red are now green, and one open question is worth naming before the browser suites finish.

## Cleared

- **`regenerate and prove no diff`** — was rejecting a hand-edit of the generated
  catalogs. Fixed by changing `spec.visibility` in each `blueprint.yaml` and
  running `build-catalog.mjs`, so committed outputs equal generator outputs.
- **`Reject Closes/Fixes/Resolves in commit messages`** — a real defect in my own
  commit body, not a false positive. I had written *"the defect that closed
  #6472"*; GitHub acts on that string, so merging would have auto-closed #6472 as
  a side effect of describing it. Reworded across all five commits via
  non-interactive `filter-branch`; zero banned verbs remain.

## Open question for whoever picks this up

`Playwright UI smoke` and `sovereign-admin console e2e` were failing on the prior
head and are still running here. They may be failing **because of this PR**: it
removes six cards from the customer catalog (21 listed -> 15), and a UI test that
asserts on a specific product or on a catalog COUNT would break on exactly that.

Two different fixes depending on which:

- asserts a **specific product** (`netbird`, `librechat`, …) is purchasable — the
  test encodes the oversell this PR removes, so the test is wrong and should be
  updated.
- asserts a **catalog count** — the fixture needs 21 -> 15.

Do not "fix" it by re-listing a product. Six of these have no Blueprint in the
catalog-seed, which is the whole defect (#6475): a customer could buy them and
the Sovereign had nothing to install.

## Content verification at this head

```
listed: 15   unlisted: 84   private: 0
six target products still listed: []      (sandbox, qa-app, anthropic-adapter,
                                           librechat, netbird, stalwart-sovereign)
CONTROL still listed: openclaw, agenity
check-listed-blueprints-are-seeded.py -> PASS (0 declared exceptions, 81 seed entries)
check-retired-products-not-listed.py  -> PASS (244 entries)
```

The `KNOWN_UNSEEDED` waiver register is now EMPTY — with every oversold product
unlisted at source there is nothing left to excuse, so the guard can no longer
pass on a waiver.

Issue #6513, opened 2026-08-20
bp-agenity 0.5.31 creds-resync sidecar lacks liveness/readiness probes — kyverno probes-present DENIES the StatefulSet, per-Org Agenity install never comes up (hw302)


emrahbaysal:
## Summary

The per-Org **Agenity** workspace (`bp-agenity` chart, version 0.5.31) fails to
install on a Sovereign. Its StatefulSet is **DENIED at admission** by the
kyverno `probes-present` ClusterPolicy (platform/kyverno-policies K1) because the
**`creds-resync` sidecar container** — introduced in 0.5.31 under
`anthropic.credentialResync.enabled` (default **true**, #6317) — declares
**neither a livenessProbe nor a readinessProbe**.

`probes-present` requires **every** container in a Deployment / StatefulSet /
DaemonSet to declare BOTH probes (JMESPath count equality:
`containers[?livenessProbe] | length(@)` must equal `containers | length(@)`,
same for readiness). The main `chepherd` container carries HTTP `/healthz`
liveness+readiness, but the new sidecar has none, so the whole StatefulSet is
rejected and the agenity Pod never comes up.

## Impact

Blocks the entire agentic-journey UAT cluster on hw302 (chepherd install → chat →
agent-driven `create_application`). Pillar 4 (per-Org Agenity + bp-openova-mcp).

Proven live on hw302. The `bp-openova-mcp` per-Org install uses the same image
but has NO such sidecar, so it installs fine — confirming the runtime path is
otherwise clear and the sidecar's missing probes are the sole blocker.

## Root cause

`products/agenity/chart/templates/statefulset.yaml` — the `creds-resync`
container (added under the `anthropic.credentialResync.enabled` gate) has
`command`, `volumeMounts`, `resources`, `securityContext` but no probes.

## Fix

Add an `exec` livenessProbe + readinessProbe to the `creds-resync` sidecar,
rendered under the same gate. The sidecar is a headless copy-forward loop (no
HTTP port), so exec probes are correct:

- readiness: the Secret mount it watches (`/creds`) is visible — always present
  (optional-secret volume mounts an empty dir even when absent), so it can never
  false-fail the pod NotReady and hide the dashboard.
- liveness: the container can still exec (`true`) — the resync is a single
  `sh -c` process, so a dead loop = container exit. We deliberately do NOT probe
  file freshness: a healthy sidecar leaves the credential untouched whenever no
[...]

Issue #6235, opened 2026-08-13
cutover-contract Case 40 is fail-open: pipefail turns a matched grep -q into 141, so a real resource-policy keep is never caught


emrahbaysal:
## Case 40 of the cutover contract test cannot detect the defect it exists to catch

`platform/self-sovereign-cutover/chart/tests/cutover-contract.sh` Case 40 keeps
`helm.sh/resource-policy: keep` off the cutover step ConfigMaps — a step CM carrying it is excluded
from the helm-upgrade UPDATE set, so a mid-cutover chart fix lands inert (#4982 Finding A, live
hw239). One line, two defects in opposite directions:

```sh
if awk "/^  name: ${cm}\$/,/^---\$/" render.yaml | grep -q 'helm.sh/resource-policy: keep'
```

### 1. Fail-open — the dominant behaviour

The script runs under `set -euo pipefail` (`cutover-contract.sh:37`). `grep -q` exits the instant it
matches, `awk` is killed by SIGPIPE (141), and **pipefail promotes 141 to the pipeline status** — so a
*found* match returns non-zero and the `if` reads it as "no match". Measured on the default render,
for the three ranges that actually contain the string:

```
cutover-step-06-helmrepository-patches: range=1535 lines, hits=1, pipeline rc=141
cutover-step-07-catalyst-api-env-patch: range=741  lines, hits=1, pipeline rc=141
cutover-step-09-gitea-token-mint:       range=262  lines, hits=1, pipeline rc=141
```

**Proof by mutant.** A genuine `helm.sh/resource-policy: keep` on step-05's `metadata.annotations`
renders cleanly (`helm template` rc=0; annotation confirmed present via `yq`):

| test version | result |
|---|---|
| `origin/main` | **PASS** — rc=0, 82 cases. The defect is invisible to it. |
| after the fix | **FAIL** at Case 40, naming `cutover-step-05-flux-gitrepository-patch` |

### 2. False positive when the race lands the other way

The range was the **whole document** including `data:`, and every step CM carries its shell script
there. #5216 (2026-07-18) added to `platform/self-sovereign-cutover/chart/templates/_helpers.tpl:222`:

```
# helm.sh/resource-policy: keep so nothing prunes the applied copy.
```

A comment *about* the annotation. When `awk` finishes before `grep` exits, the pipeline returns 0 and
Case 40 fires on that comment. Observed on the first run of the investigating session — rc=1, **42 of
[...]

PR #6514, opened 2026-08-20
fix(agenity): creds-resync sidecar liveness+readiness probes so kyverno probes-present admits the StatefulSet (#6513)

emrahbaysal:
## Problem

The per-Org **Agenity** workspace (`bp-agenity` chart, 0.5.31) fails to install
on a Sovereign. Its StatefulSet is **DENIED at admission** by the kyverno
`probes-present` ClusterPolicy (platform/kyverno-policies K1) because the
**`creds-resync` sidecar** — introduced in 0.5.31 under
`anthropic.credentialResync.enabled` (default **true**, #6317) — declares
**neither a livenessProbe nor a readinessProbe**. The policy requires every
container in a Deployment/StatefulSet/DaemonSet to declare both (a JMESPath
count-equality: `containers[?livenessProbe] | length(@)` must equal
`containers | length(@)`, same for readiness). The main `chepherd` container has
HTTP `/healthz` probes; the new sidecar had none, so the whole StatefulSet was
rejected and the workspace never came up.

Proven live on hw302, blocking the entire Pillar-4 agentic journey (chepherd
install → chat → agent-driven `create_application`). The `bp-openova-mcp` per-Org
install uses the same image but has no such sidecar and installs fine —
isolating the sidecar's missing probes as the sole cause.

## Remediation

`products/agenity/chart/templates/statefulset.yaml` — add an `exec`
livenessProbe + readinessProbe to the `creds-resync` sidecar, rendered under the
**same** `{{ if and .Values.anthropic.credentialsKey (.Values.anthropic.credentialResync).enabled }}`
gate as the container (so the probes exist iff the sidecar does). The sidecar is
a headless copy-forward loop with no HTTP port, so exec is the correct handler,
and each probe is chosen so it can never regress a running workspace:

- **readiness** `test -d /creds` — the Secret mount it watches is visible.
  `/creds` is always present (the `anthropic-creds` volume is an OPTIONAL secret
  that still mounts an empty dir when absent), so this can never false-fail the
  pod NotReady and hide the dashboard the main container must keep serving.
- **liveness** `true` — the container can still exec. The resync is a single
  `sh -c` process, so a dead loop = container exit; container-alive == loop-alive.

It deliberately does **not** probe the freshness of
`/claudehome/.claude/.credentials.json`: a healthy sidecar correctly leaves that
file untouched whenever the Secret carries no newer `expiresAt` (the NEWER-ONLY
predicate), so a freshness probe would kill a working sidecar. The sidecar's
copy-forward function and the `credentialResync` default are unchanged.

## Test

New `products/agenity/chart/tests/creds-resync-probes.sh` renders the chart and
asserts:
1. the sidecar renders when `credentialResync.enabled=true`;
2. it carries both probes, both using the `exec` handler (an httpGet regression
   would probe a nonexistent port);
3. the kyverno count-equality invariant holds across **every** container;
4. the sidecar-disabled path still satisfies the policy (probes render iff the
   sidecar does);
5. **vacuity** — stripping the sidecar's probes back out makes the invariant
   FAIL, proving the check is real.

Runs at PR time via `chart-tests-pr.yaml` under the release-pinned helm + yq; it
renders a single document (`--show-only templates/statefulset.yaml`) so it never
hits the multi-doc yq-version skew (#6235).

## Verified locally

- `helm template` with `credentialResync.enabled=true` → creds-resync now
  carries both exec probes; `containers=2 with-liveness=2 with-readiness=2` →
  kyverno `probes-present` would ADMIT the StatefulSet.
- New test PASS (incl. vacuity control); all existing agenity chart tests PASS
  (credential-resync-6317, credential-verdict, oidc-gate-companion-render).
- `check-chart-version-bumped.sh`: agenity 0.5.31→0.5.32, catalyst 1.4.1552→1.4.1553.
- `check-catalog-seed-lockstep.sh`: PASS (68 checked, 0 fail).
- catalog regen (`build-catalog.mjs`) deterministic, no drift.
- `check-bootstrap-kit-pin-sync.sh`: bp-catalyst-platform chart=1.4.1553 pin=1.4.1553.
- `helm lint`: 0 failed.

## Lockstep (bp-agenity 0.5.31 → 0.5.32)

- `products/agenity/chart/Chart.yaml` (+ statefulset template fix + new test)
- `products/agenity/blueprint.yaml`
- catalog-seed card+source: `products/catalyst/chart/templates/catalog-seed/blueprints.yaml`
- umbrella `products/catalyst/chart/Chart.yaml` 1.4.1552 → 1.4.1553
- bootstrap-kit slot 13 pin `clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml`
- regenerated catalog: `blueprints.json` + `catalog.generated.ts`

bp-agenity is OCI-delivered to Sovereigns, so this reaches hw302 by image/chart
pull once published — not gated by the GitRepository clone issue.

Refs #6513 #3988

### DIFF
--- a/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
+++ b/clusters/_template/bootstrap-kit/13-bp-catalyst-platform.yaml
@@ -1473,7 +1473,12 @@ spec:
       #   is honored); pool-domain logins STILL 400'd (proven live hw302). The
       #   organization-controller now registers each Org's CONCRETE console
       #   callback https://console.<slug>.<pool-tld>/* at reconcile time.
-      version: 1.4.1552
+      # 1.4.1553 — #6513: catalog-seed advances bp-agenity 0.5.31 -> 0.5.32 —
+      #   the creds-resync sidecar gains liveness + readiness probes so the
+      #   per-Org agenity StatefulSet passes kyverno probes-present (0.5.31
+      #   shipped it probe-less → DENIED at admission, install never came up,
+      #   hw302; blocked the Pillar-4 agentic walk).
+      version: 1.4.1553
       sourceRef:
         kind: HelmRepository
         name: bp-catalyst-platform
--- a/core/services/provisioning/gitops/helmrelease_apps.go
+++ b/core/services/provisioning/gitops/helmrelease_apps.go
@@ -723,7 +723,7 @@ spec:
 // the catalog seed to 0.5.31 (server-side Anthropic OAuth credential renewal +
 // resync sidecar); this funnel pin moves in lockstep so a purchase installs the
 // renewing chart, not the one whose credential dies every ~5h.
-const DefaultHRAppChartVersions = "openclaw=0.2.19,stalwart-mail=0.1.16,newapi=1.4.153,agenity=0.5.31"
+const DefaultHRAppChartVersions = "openclaw=0.2.19,stalwart-mail=0.1.16,newapi=1.4.153,agenity=0.5.32"
 
 // ParseHRAppVersions parses the CATALYST_HR_APP_CHART_VERSIONS wire format
 // ("slug=version,slug=version") into the HelmReleaseAppVersions map (#4706).
--- a/products/agenity/blueprint.yaml
+++ b/products/agenity/blueprint.yaml
@@ -18,7 +18,11 @@ spec:
   # Keycloak backchannel + upstream) — the per-Org namespace is default-deny
   # and the gate's ingress-only CNP left the OAuth callback 500ing on a DNS
   # timeout. Lockstep with products/agenity/chart + the catalog-seed pin.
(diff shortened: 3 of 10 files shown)


emrahbaysal:
Pushed 5a60f9708: bumped the Go funnel pin `DefaultHRAppChartVersions`
agenity 0.5.31→0.5.32 — the last lockstep site this PR had missed.
`TestDefaultHRAppPins_MatchCatalogSeed/agenity` now passes locally
(`go test ./core/services/provisioning/gitops` → ok). Also reworded the PR
body to clear the banned-words gate. Remaining red is `ghcr-pin-existence`
(bp-agenity@0.5.32 not yet published to ghcr) — a non-required check that
resolves when the chart publishes on merge.

Issue #6703, opened 2026-08-30
epic: unified Jobs/Resources surface (chips + list + dependency-DAG + CronJob schedule)


emrahbaysal:
Founder-signed-off design (2026-08-30). Deliver P1→P4 end-to-end on live hw305, no wipe.

**Model:** Resources=assets · Jobs=processes. HelmRelease is dual-natured (asset in Resources + install-attempt process in Jobs). Real engine names throughout. CronJob = recurring category.

**P1** treemap leaf-name fix (component-first) + /jobs chips + List/Graph toggle
**P2** Graph view = dependency DAG (reuse JobDependenciesGraph widget), grouped by orchestration unit
**P3** CronJob Schedule timeline (read-only): 'what fires at 12pm', collisions, scheduled-vs-actual
**P4** step contract linear-order → dependsOn DAG; + New schedule (ad-hoc)

Each phase ships as code → CI builds catalyst-ui/api → set-image roll onto live hw305 → browser-validate.

emrahbaysal:
**P1a shipped + validated live on hw305** (catalyst-ui:c4b92a1, PR #6704). Treemap tiles now lead with the humanized component (Cilium/Keycloak/Gitea…) instead of "Install…", so the ~65 HelmReleases are distinguishable in the Kind view. Verified via browser: buckets read HelmRelease/Kustomization/Job(task/step)/CronJob/OpenTofu, zero "install" tile text, component names render at width. Full UI suite green (2271) incl. a new leaf-name test.

P1b (/jobs chips + List/Graph) in progress (agent). Next: review+merge P1b → P2 grouped dependency-DAG.

emrahbaysal:
### Delivery progress — autonomous execution (2026-08-30)

Working the signed-off P1–P4 design end-to-end on the **existing live hw305** (no wipe/re-provision).

| Item | State | PR |
|---|---|---|
| **P1a** treemap leaf-name (component-first, real engine buckets) | ✅ merged + rolled + **validated live** (hw305 catalyst-ui:c4b92a1; screenshot sent) | #6704 |
| **P1b** /jobs per-kind chips + List⇄Graph toggle (mirror /cloud) | 🟡 PR open, CI running | #6706 |
| **P3-backend** watch batch/v1 CronJob + Job (schedule-timeline data) | 🟡 PR open, CI running (`test` green) | #6705 |
| **P2** graph = grouped dependency DAG | ⏭ next, builds on P1b's JobsGraphView | — |
[...]

PR #6707, opened 2026-08-30
feat(sovereign-ui): /jobs graph view = grouped dependency DAG + chips-in-graph (P2)

emrahbaysal:
## What

The /jobs **Graph** view (added in P1b #6706) now renders a **grouped dependency DAG** and keeps a **chip strip in graph view**. Epic #6703, P2. catalyst-ui only.

Founder-signed: "the grap view same as cloud grap, it should still contain the chips and we should ahve soem default gahtered grelavant groups".

## Changes (frontend only)

- `JobsGraphView.tsx` — rewritten: partitions jobs into **sections** — one per orchestration-unit **group row** (`type:'group'`, e.g. Bootstrap / Provision Hetzner / Apps / Reconcilers / Handover / Cutover) that has ≥1 child, plus a **Standalone tasks** section for parentless leaves. Each section renders its own `JobDependenciesGraph` over its members, so `dependsOn` edges connect only that unit's steps. Section titles come from the group row's `displayName` (never hardcoded); ordering via a `SECTION_ORDER` rank map, standalone last; empty groups render no section. Node-click nav preserved from P1b.
- `JobsPage.tsx` — the chip strip now renders in **graph** view too (was a spacer). Graph chips **highlight** the selected kind and dim the rest (opacity only — removing nodes would sever mid-chain `dependsOn` edges). Uses a graph-local highlight state (default none) distinct from the list's `?kind=` filter, since list=filter and graph=highlight are different semantics; clicking the active chip clears it.
- `JobDependenciesGraph.tsx` — new **optional** `dimNodeIds?: ReadonlySet<string>` prop (opacity-only dim of non-highlighted nodes/edges; absent ⇒ no attribute emitted, existing render/Playwright contract unchanged). Also **namespaced the arrow `<marker>` id via `useId()`** — P2 renders many graph instances per page, and the previously-hardcoded id would repeat (invalid duplicate DOM ids; `url(#id)` resolves only to the first).

## Verification (post-rebase onto main)

- `tsc -b` clean.
- Full ui vitest: **257 files / 2302 pass + 1 pre-existing expected-fail** (+15 tests vs P1b baseline: grouped sections, standalone, chips-in-graph highlight toggle, cross-kind dim, per-instance marker-id uniqueness). Zero regressions.
- No new eslint errors.

## Reviewer notes (design decisions)

- Chip-highlight uses a local state + an empty-string `activeKind` sentinel so `JobKindChips` is reused unmodified for "no highlight". Highlight persists across list⇄graph toggles within a session (un-highlighted on fresh entry, per spec).
- One `JobDependenciesGraph` per section (stacked) rather than one clustered mega-graph — reuses the widget as-is, lowest-risk.

## Rollout

catalyst-ui only → safe rolling roll to hw305 after merge + CI; validated live from the browser (screenshot on the epic).

Refs #6703

### DIFF
--- a/products/catalyst/bootstrap/ui/src/pages/sovereign/JobsGraphView.test.tsx
+++ b/products/catalyst/bootstrap/ui/src/pages/sovereign/JobsGraphView.test.tsx
@@ -1,0 +1,225 @@
+/**
+ * JobsGraphView.test.tsx — P2 grouped-DAG lock-in (Refs #6703).
+ *
+ * Asserts the Graph view:
+ *   • renders ONE JobDependenciesGraph per present group row (section),
+ *     titled by the group's displayName, with the member count;
+ *   • renders a "Standalone tasks" section for parentless leaves;
+ *   • orders sections by the fixed preference (provisioner → apps →
+ *     standalone), tolerating whichever groups are present;
+ *   • keeps the `data-testid="jobs-graph-view"` wrapper (existing contract);
+ *   • HIGHLIGHTS a kind by dimming every NON-matching node/edge (opacity),
+ *     never removing nodes — and dims nothing when highlightKind is null;
+ *   • renders a graceful empty state for zero jobs;
+ *   • preserves node-click navigation.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest'
+import { render, screen, cleanup, fireEvent } from '@testing-library/react'
+import {
+  RouterProvider,
+  createRouter,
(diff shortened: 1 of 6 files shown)


emrahbaysal:
Superseded by #6708 — the grouped-SVG graph here was the ad-hoc reinvention the founder rejected. #6708 reworks /jobs to reuse the canonical FlowCanvasOrganic DAG + one shared curatable chip strip (no clone).

